PR #542, opened 2024-09-19
remove ArgumentsAttribute constructor that always throws

SimonCropp:


thomhurst:
Thanks @SimonCropp - looks like there was an analyzer for this, and now a failing test so that might need to go too

SimonCropp:
@thomhurst done

SimonCropp:
well i remove the test. do i remove the analyser as well?

thomhurst:
> well i remove the test. do i remove the analyser as well?

Tbh I wrote that analyser very early on. If it's basically redundant (because the attribute constructor is redundant) then yeah delete it.

I basically want it to throw if no arguments were provided, and it still will do that right?

PR #607, opened 2024-09-23
docs: update extensibility.md

eltociear:
minor fix

### DIFF
--- a/docs/docs/tutorial-assertions/extensibility.md
+++ b/docs/docs/tutorial-assertions/extensibility.md
@@ -17,7 +17,7 @@ Delegates are bits of code that haven't executed yet - Instead they are passed i
 So to create a custom assertion:
 
 1. Create a class that inherits from `AssertCondition<TActual, TExpected>`
-   `TActual` will the type of object that is being asserted. For example if I started with `Assert.That("Some text")` then `TActual` would be a `string` because that's what we're asserting on.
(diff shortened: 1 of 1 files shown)


docs/docs/tutorial-assertions/extensibility.md
```diff
@@ -17,7 +17,7 @@ Delegates are bits of code that haven't executed yet - Instead they are passed i
 So to create a custom assertion:
 
 1. Create a class that inherits from `AssertCondition<TActual, TExpected>`
-   `TActual` will the the type of object that is being asserted. For example if I started with `Assert.That("Some text")` then `TActual` would be a `string` because that's what we're asserting on.
+   `TActual` will the type of object that is being asserted. For example if I started with `Assert.That("Some text")` then `TActual` would be a `string` because that's what we're asserting on.
```
thomhurst:
```suggestion
   `TActual` will be the type of object that is being asserted. For example if I started with `Assert.That("Some text")` then `TActual` would be a `string` because that's what we're asserting on.
```

dmiky-xamk, line 20:
```suggestion
   `TActual` will be the type of object that is being asserted. For example if I started with `Assert.That("Some text")` then `TActual` would be a `string` because that's what we're asserting on.
```

Just a simple grammar fix.

eltociear: thanks!

thomhurst:
Thanks @eltociear

PR #661, opened 2024-09-28
Use CultureInfo.InvariantCulture when calling ToString on objects

### LINKED ISSUE #660: NumberArgumentTests fail when compiled under locale which does not use . as decimal point (e.g. de-DE)
https://github.com/thomhurst/TUnit/blob/659b0801f806bf74f03d7c8e60fee8ec4d2708ef/TUnit.TestProject/NumberArgumentTests.cs#L12-L17

produces invalid code when compiled under locale which does not use . as decimal point (e.g. de-DE). Example:

```c#
global::System.Double methodArg0 = 1,1D;
```

Because the numeric ToStrings are culture aware:

https://github.com/thomhurst/TUnit/blob/659b0801f806bf74f03d7c8e60fee8ec4d2708ef/TUnit.Engine.SourceGenerator/CodeGenerators/Helpers/TypedConstantParser.cs#L29-L62

campersau:
Fixes #660


thomhurst (approved):

thomhurst:
Thanks @campersau ! This change is in v0.1.857

PR #663, opened 2024-09-28
Refactor DataDrivenArgumentsRetriever and MatrixRetriever to directly embed attribute value expressions

campersau:
Instead of extracting the values from `MatrixAttribute` / `ArgumentsAttribute` and then `ToString` them again, we could just use the argument expressions and embed them directly into the generated source code.
This avoids culture related issues like #660, adds support for more string expression like `""" """` and avoids updating the source generator in the future to support new ones.


thomhurst:
Nice one @campersau , thanks! Removing that logic from the `Argument` class makes me very happy. This is indeed a lot cleaner, thanks very much for the contribution! Will push out a new version 😄

PR #669, opened 2024-09-28
Fix tests in discovery and session hooks

### LINKED ISSUE #668: Context.AllTests empty for AfterEvery(TestDiscovery) and BeforeEvery(TestSession) Hooks
Hi there, ✌️

i just found out that there are no tests given in the contexts when hooking into `AfterEvery(TestDiscovery)` and `BeforeEvery(TestSession)`. 

So here a quick example:

Some dummy tests:
```
public class DummyTest
{
    [Test]
    public async Task DummyTestAsync()
    {
        var a = 1;
        var b = 2;
        await Assert.That(a + b).IsEqualTo(3);
    }

    [Test]
    public async Task DummyTest2Async()
    {
        var a = 2;
        var b = 2;
        await Assert.That(a + b).IsEqualTo(4);
    }
}
```

GlobalHooks.cs
```
public static class GlobalHooks
{
    [AfterEvery(TestDiscovery)]
    public static void AfterDiscovery(TestDiscoveryContext context)
    {
        var testCount = context.AllTests.Count();
        // always zero.
        Console.WriteLine($"Test count after discovery: {testCount}");
    }

    [BeforeEvery(TestSession)]
    public static void BeforeSession(TestSessionContext context)
    {
        var testCount = context.AllTests.Count();
        // always zero.
        Console.WriteLine($"Test count before session: {testCount}");
    }

    [Before(Assembly)]
    public static void BootstrapAsync(AssemblyHookContext context)
    {
        var testCount = context.AllTests.Count();
        // 2. But only contains the tests of the specifiy assembly.
        Console.WriteLine($"Test count before assembly: {testCount}");
    }
}
```

I would expect that the TestDiscoveryContext provides all tests that where discovered and the TestSessionContexts provides all tests that are meant to be run in the session regadless of in what assembly they are. It would be really nice to have these informations in these hooks.

I tested this with the latest vs preview version and also just using a plain `dotnet test`. Same result though.

Also one side question:
Why do we have to use `BeforeEvery` and `AfterEvery` for the TestSession and TestDiscovery Hooks? I tested using just `Before` and `After` but both of these did not work and the methods never were called. 

Thanks in advance. :)

tntwist:
fixes #668

### DIFF
--- a/TUnit.TestProject/AfterTests/TestSessionAfterTests.cs
+++ b/TUnit.TestProject/AfterTests/TestSessionAfterTests.cs
@@ -1,7 +1,11 @@
-﻿namespace TUnit.TestProject.AfterTests;
+﻿using TUnit.Assertions;
+using TUnit.Assertions.Extensions;
 
-public class TestSessionAfterTests
+namespace TUnit.TestProject.AfterTests; 
+
+public class TestSessionAfterHooksTests
 {
+    // TODO: The "After(TestSession)" hook is currently not being called/source generated
     [After(TestSession)]
     public static async Task AfterTestSession(TestSessionContext context)
     {
@@ -11,6 +15,17 @@ public class TestSessionAfterTests
     [AfterEvery(TestSession)]
     public static async Task AfterEveryTestSession(TestSessionContext context)
     {
-        await Task.CompletedTask;
+        var firstTest = context.AllTests.First();
+        await Assert.That(firstTest.ObjectBag["AfterEveryTestSessionHit"]).IsEqualTo(true);
+    }
+}
+
+public class TestSessionAfterTests
+{
+    [Test]
+    public async Task PepareForAfterSession()
+    {
+        TestContext.Current?.ObjectBag.Add("AfterEveryTestSessionHit", true);
+        await Assert.That(TestContext.Current?.ObjectBag["AfterEveryTestSessionHit"]).IsEqualTo(true);
     }
 }
--- a/TUnit.TestProject/AfterTests/TestSessionAfterTests.cs
+++ b/TUnit.TestProject/AfterTests/TestSessionAfterTests.cs
@@ -15,8 +15,16 @@ public class TestSessionAfterHooksTests
     [AfterEvery(TestSession)]
     public static async Task AfterEveryTestSession(TestSessionContext context)
     {
-        var firstTest = context.AllTests.First();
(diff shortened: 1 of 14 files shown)


thomhurst:
Sorry forgot to mention something before.

Could you make tests where each context access the first test and just print something? Thisll proove that the collection is populated as otherwise it'll throw exceptions.

Could you then go into the pipeline project and test modules folder. Copy one of the other classes and make it invoke your test class. This makes sure that the pipeline is executing these on each build. The pipeline test modules hopefully are self explanatory on how they filter to specific tests, but if not give me a shout and I'll explain/help.
You may not actually need to make changes to the test project as it'll just be picked up naturally since it's a discovery / session hook to the entire project.

tntwist:
Sure. will do this tomorrow.

tntwist:
@thomhurst 
I have added tests for the contexts. Please have a look.

thomhurst (approved):

TUnit.TestProject/AfterTests/TestSessionAfterTests.cs
```diff
@@ -0,0 +1,31 @@
...
+    public static async Task AfterTestSession(TestSessionContext context)
+    {
+        await Task.CompletedTask;
+    }
+
+    [AfterEvery(TestSession)]
+    public static async Task AfterEveryTestSession(TestSessionContext context)
+    {
+        var firstTest = context.AllTests.First();
+        await Assert.That(firstTest.ObjectBag["AfterEveryTestSessionHit"]).IsEqualTo(true);
```
thomhurst:
```suggestion
        var test = context.AllTests.FirstOrDefault(x =>
            x.TestDetails.TestName == nameof(TestSessionAfterTests.PepareForAfterSession));

        if (test == null)
        {
            // This test might not have been executed due to filters, so the below exception would cause problems.
            return;
        }

        await Assert.That(test.ObjectBag["AfterEveryTestSessionHit"]).IsEqualTo(true);
```

thomhurst:
Just pushed a change. You were asserting some files existed in the pipeline testing modules, but they were never being written.

I've just made the hooks actually write these files. And then the pipeline can assert they exist.

It was my original way of validating the hooks were run outside of the test process itself.

thomhurst:
Thanks @tntwist ! Should be fixed in version 0.1.872.

The Before/After TestSession/Discovery hooks should also be fixed! 😄

tntwist:
Awesome @thomhurst! Works like a charm  now ✌🏼

PR #693, opened 2024-10-02
Feature: add IsDefault/IsNotDefault assertions

c-schembri:
IsDefault/IsNotDefault assertions are similar to IsNull/IsNotNull in that they're a short-hand assertion for an equality check, except instead of comparing against `null` they compare against `default(T)`. This means that writing assertions against primarily value types becomes shorter.

E.g., 
```
await Assert.That(myVariable).IsNotEqualTo(default);
```
Becomes
```
await Assert.That(myVariable).IsNotDefault();
```


thomhurst (approved):

thomhurst:
Thanks @c-schembri ! Should be in v0.1.891

PR #733, opened 2024-10-04
Use _maximumParallelTests instead of calling GetParallelTestsLimit() again

campersau:
`_maximumParallelTests` gets initialized in the constructor.

---

Unrelated question:
Since both of these groups are running concurrently does it mean that the real concurrent test count is actually `_maximumParallelTests * 2` when using `NotInParallelAttribute([keys])` and "normal" tests together?

https://github.com/thomhurst/TUnit/blob/ba9dc4901873ac0919244541794a65aaf73eaee4/TUnit.Engine/Services/TestsExecutor.cs#L40-L44

### DIFF
--- a/TUnit.Engine/Services/TestsExecutor.cs
+++ b/TUnit.Engine/Services/TestsExecutor.cs
@@ -108,7 +108,7 @@ internal class TestsExecutor
     {
         await Parallel.ForEachAsync(queue, new ParallelOptions
         {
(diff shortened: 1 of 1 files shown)


thomhurst:
Nice spot! Yes you're right, this is an oversight on my part. Could you raise an issue please?

thomhurst:
Also thanks for the contribution

PR #803, opened 2024-10-07
Update BaseAssertCondition Char Format

campersau:
Looks like the `'` suffix is missing for formatting char.


thomhurst:
Thanks! 😄

PR #815, opened 2024-10-08
Added unit tests to cover assertion methods with both required and op…

nies14:
Added unit tests for assertion methods, covering scenarios with and without optional parameters.


thomhurst:
Thanks for the contribution!

PR #754, opened 2024-10-05
Add a because reason to assertions

vbreuss:
As described in #747:
Add a `Because` method to `InvokableValueAssertionBuilder<TActual>` that allows specifying a reason for the assertion.

Issue #294, opened 2024-08-08
Tests for all assertion methods


thomhurst:
With and without optional parameters

- This helps capture where generics cannot be inferred properly
- It helps capture any methods that are ambiguous and conflict with each other
- We need to just confirm the assertion works as expected

nies14:
PR Link: https://github.com/thomhurst/TUnit/pull/815

PR #838, opened 2024-10-10
Improve readability of assertion failure messages

### LINKED ISSUE #764: Improve readability of assertion failure messages
I am having trouble interpreting the error messages, as the assertions is always at the beginning and can be quite long, e.g.
image

I think it would improve readability, if the assertion messages are concise and to the point, similar to [fluentassertions](https://github.com/fluentassertions/fluentassertions).

vbreuss:
Fixes #764 as also discussed in [this comment](https://github.com/thomhurst/TUnit/pull/754#issuecomment-2395216981) by
- Moving the assertion expression at the end
- Replacing the `GetFailureMessage` with `GetExpectation` which returns only the expectation from the `BaseAssertCondition`
- Replacing the `Passes` method with `GetResult` which returns not only a boolean value, but an `AssertionResult` which also contains the failure message.

This allows the assertion message to be combined to read like an english sentence, e.g.
instead of
```
Assert.That(myClass).HasMember(x => x.Nested.Nested.Nested.Number).EqualTo(1)
MyClass.Number:
    Expected: 1
    Received: 123
```
it would read
```
Expected myClass MyClass.Number to be equal to 1, but received 123.
At Assert.That(myClass).HasMember(x => x.Nested.Nested.Nested.Number).EqualTo(1)
```

Also added some tests for correctly throwing exceptions (see #294) where the new output reads e.g.
```
Expected action to have Message equal to "Fails_For_Some_Other_Reason", but it differs at index 10:
              ↓
   "Fails_For_Exceptions_With_Different_Message"
   "Fails_For_Some_Other_Reason"
              ↑.
At Assert.That(action).ThrowsException.With.Message.EqualTo("Fails_For_Some_Other_Reason", StringCompar...
```


vbreuss:
@thomhurst : I don't understand why the pipeline fails.
Locally all unit tests compile and succeed, but on the pipeline the [TimeoutTests5](https://github.com/thomhurst/TUnit/actions/runs/11271136276/job/31344841403?pr=838#step:5:4610) fail 🙈

thomhurst:
> @thomhurst : I don't understand why the pipeline fails. Locally all unit tests compile and succeed, but on the pipeline the [TimeoutTests5](https://github.com/thomhurst/TUnit/actions/runs/11271136276/job/31344841403?pr=838#step:5:4610) fail 🙈

It looks like one of the assertions has broken. If you go to the pipeline output, then scroll above the table you can expand each step to see it's full output.

I'm seeing this:

```
failed MatrixTest(2) (46ms)
    Expected value to be equal to 1
     or
    to be equal to 2
     or
    to be equal to 3, but the received value 2 is different and the received value
  2 is different.
    At Assert.That(value).IsEqualTo(1).Or.IsEqualTo(2).Or.IsEqualTo(3)
```

TUnit.Assertions/AssertConditions/AssertionResult.cs
```diff
@@ -0,0 +1,67 @@
...
+
+	public AssertionResult Or(AssertionResult other)
+	{
+		if (!IsPassed && !other.IsPassed)
+		{
+			return Fail(this.Message + " and " + other.Message);
+		}
+
+		return this;
+	}
```
thomhurst: I think this is causing your test failures:

```suggestion
	public AssertionResult Or(AssertionResult other)
	{
		if (!IsPassed && !other.IsPassed)
		{
			return Fail(this.Message + " and " + other.Message);
		}

		return Passed;
	}
```

TUnit.Assertions/AssertConditions/Connectors/OrAssertCondition.cs
```diff
@@ -12,18 +12,23 @@ public OrAssertCondition(BaseAssertCondition<TActual> condition1, BaseAssertCond
         
         _condition1 = condition1;
         _condition2 = condition2;
-    }
+	}
 
-    protected internal override string GetFailureMessage() => $"{_condition1.OverriddenMessage ?? _condition1.GetFullFailureMessage()}{Environment.NewLine} or{Environment.NewLine}{_condition2.OverriddenMessage ?? _condition2.GetFullFailureMessage()}";
+    protected override string GetExpectation()
```
thomhurst: Does this not get called? If not I'd prefer to return an empty string than throw not supported.

TUnit.Assertions/AssertConditions/AssertionResult.cs
```diff
@@ -0,0 +1,67 @@
...
+    public bool IsPassed { get; }
+    public string? Message { get; }
+
+    private AssertionResult(bool isPassed, string? message)
+    {
+        IsPassed = isPassed;
+        Message = message;
+	}
+
+	public static AssertionResult FailIf(Func<bool> isFailed, string message)
```
thomhurst: Does this need to take a func? Looks like it invokes it immediately. Couldn't you just pass in the boolean result directly?

TUnit.Assertions/Extensions/StringExtensions.cs
```diff
@@ -23,7 +23,17 @@ public static string TruncateWithEllipsis(this string value, int maxLength)
         {
             return value;
         }
-        
-        return $"{value[..maxLength]}...";
+
+		return $"{value[..maxLength]}...";
```
thomhurst: Any reason for the tab increase?

TUnit.Assertions/AssertConditions/Connectors/OrAssertCondition.cs
vbreuss: No, it does not get called, as I overwrite the `GetExpectationWithReason` and combine the message from `_condition1` and `_condition2`.
I will change it to return an empty string and add a comment.

TUnit.Assertions/AssertConditions/AssertionResult.cs
vbreuss: I could probably get rid of it here, but not in the `OrFailIf` method, as it would then throw, e.g. in 
```csharp
		=> AssertionResult
			.FailIf(
				() => actualValue is null,
				$"{ActualExpression ?? typeof(TActual).Name} is null")
			.OrFailIf(
				() => actualValue.Contains(inner, equalityComparer),
				"it was found in the collection");
```
because both conditions are always evaluated and not just when all previous conditions didn't fail.
However having different method signatures for `FailIf` and `OrFailIf` looks strange to me...

TUnit.Assertions/Extensions/StringExtensions.cs
vbreuss: No, it seems I got somehow tabs mixed within. Will have a general look at my files to use consistently spaces, as in the other files...

thomhurst:
Thanks @vbreuss !

PR #844, opened 2024-10-10
Update extensibility docs for Assertions

thomhurst:
cc @vbreuss anything you think needs adding?


vbreuss:
@thomhurst Apart from the issue, I fixed with #846, the changes look good 👍

PR #870, opened 2024-10-12
Improve `Throw` assertions

vbreuss:
I have a couple of issues with the current implementation of the `Throw` assertions:
- AutoComplete suggests two generic overloads, which I must not use
  image
  First I thought, that I could specify the type of the Exception, but I have to use `.ThrowsException().OfType<ArgumentException>(()` instead of `ThrowsException<ArgumentException>()`
- When awaiting `Assert.That(action).ThrowsException()` I don't get the exception back to make further assertions on it
- I cannot combine multiple assertions on an exception, as the `.With.Message` does not keep the reference to the original exception.

My suggestion in this PR is to refactor the delegate assertions as follows:
- The main entry point is `.Throws()` which removes the generic overload of `TActual` from subsequent calls
- Then we have 
  - `.Nothing()` (no exception is thrown)
  - `.Exception()` (any exception is thrown)
  - `.OfType<TException>()` (an exception of type TException or any subtype is thrown)
  - `.Exactly<TException>()` (an exception of type TException is thrown)
- For all thrown exceptions we have
  - `.WithMessage(string)` (checks that the message is equal)
  - `.WithMessageMatching(string)` (checks that the message matches the (wildcard) pattern)
  - `WithInnerException()` which allows checks on the inner exception


thomhurst:
Sorry I've just merged mine in. 

> The main entry point is .Throws() which removes the generic overload of TActual from subsequent calls

You've probably noticed then that the extension methods have the issue where if you want to supply an extra generic argument, e.g. `TExpected`, the user still has to pass in the `TActual` value too which makes the assertion difficult to write and harder to understand.

I want to have this syntax:

`ArgumentException exception = await Assert.That(() => SomeMethod()).Throws<ArgumentException>();`

While the extensions will cause problems with the generic typing, this can be mitigated by defining these methods on the `IDelegateSource` interface. I've actually just done that with the `TypeOf<>` methods for `IValueSource` here: https://github.dev/thomhurst/TUnit/tree/main/TUnit.Assertions

Issue #872, opened 2024-10-12
Generic `Which` assertion


vbreuss:
I think it would be a great improvement to have generic `Which` assertions with the following signature:
```csharp
InvokableValueAssertionBuilder<TActual> Which<TActual, TExpected>(this IValueSource<TActual> valueSource, Func<TActual, TExpected> mapper, Action<IValueSource<TExpected>> assertions);
InvokableValueAssertionBuilder<TActual> WhichAsync<TActual, TExpected>(this IValueSource<TActual> valueSource, Func<TActual, Task<TExpected>> asyncMapper, Action<IValueSource<TExpected>> assertions)
```
This would allow combining multiple assertions on properties of an object, e.g.
```csharp
    HttpResponseMessage sut = new();
    await Assert.That(sut)
        .Which(x => x.StatusCode, s => s.IsEquatableOrEqualTo(System.Net.HttpStatusCode.OK)).And
        .WhichAsync(x => x.Content.ReadAsStringAsync(), s => s.Equals("foo"));
```
For commonly used types, it would be easy to implement special extensions, e.g.
```csharp
    await Assert.That(sut)
        .HasSuccessStatusCode().And
        .HasContentEqualTo("foo");

    public static InvokableValueAssertionBuilder<HttpResponseMessage> HasSuccessStatusCode(this IValueSource<HttpResponseMessage> valueSource)
        => valueSource.Which(x => x.StatusCode, s => s.IsEquatableOrEqualTo(System.Net.HttpStatusCode.OK));

    public static InvokableValueAssertionBuilder<HttpResponseMessage> HasContentEqualTo(this IValueSource<HttpResponseMessage> valueSource, string content)
        => valueSource.WhichAsync(x => x.Content.ReadAsStringAsync(), s => s.Equals(content));
```

PR #877, opened 2024-10-12
Satisfies extensions methods for selecting strongly typed properties/…

thomhurst:
Implements #872


thomhurst:
@vbreuss I went with the verb "Satisfies" as I think it reads better. What do you think?

vbreuss:
> @vbreuss I went with the verb "Satisfies" as I think it reads better. What do you think?

Yes, that is a good idea.
I had the following use case in mind:
```csharp
await Assert.That(sut).IsOfType<SomeType>().Which(x => x.Property1, p => p.IsEqualTo("foo"));
```
but without some prior assertion, this read strange and `Satisfies` is clearer!

PR #881, opened 2024-10-12
Remove `Throw` methods from `Assert`

vbreuss:
As we improved the `Throw` methods with #870, we should use them consistently. Therefore I would suggest to remove the `Throw` methods from the static `Assert` class, as they are redundant.

### DIFF
--- a/TUnit.Assertions.Tests/AssertConditions/BecauseTests.cs
+++ b/TUnit.Assertions.Tests/AssertConditions/BecauseTests.cs
@@ -13,7 +13,7 @@ public class BecauseTests
             await Assert.That(variable).IsFalse().Because(because);
         };
 
-        var exception = await Assert.ThrowsAsync<AssertionException>(action);
+        var exception = await Assert.That(action).Throws<AssertionException>();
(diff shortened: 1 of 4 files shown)


thomhurst:
I think I'm happy to leave them to be honest. Some people will prefer the shorthand versions.

vbreuss:
For me, it would be a benefit to have a consistent assertion API, where I can search for `Assert.That` and find all assertions in a code base. When there are shortcuts/alternatives for this, it gets harder to distinguish the normal pattern, especially as here the order of the exception type and the subject under test is different.
That was the main difficulty I had, when migrating a previous project from NUnit to xunit, that the order of the arguments was sometimes different.

And it is also not that much shorter 😄 
```csharp
await Assert.Throws<CustomException>(action);
// vs.
await Assert.That(action).Throws<CustomException>();
```

*That's just my thoughts. In the end it is up to you, and I can close the PR...*

thomhurst:
Yeah let's close it for now. The API you desire is still there. And it might annoy some people if we remove this overload. If you don't wanna use it then you don't have to 😄

PR #882, opened 2024-10-12
Add `OnlyIf` method for throw assertions

vbreuss:
I often have the issue that I want to assert in a test, that an exception was only thrown when a certain condition applies.
Currently I have to introduce an if/else clause, e.g.
```csharp
if (expectMatch)
{
    await Assert.That(sut).ThrowsNothing();
}
else
{
    await Assert.That(sut).ThrowsException().WithMessageMatching(expectedExpression);
}
```

This PR adds an `OnlyIf` method to the `ThrowsException` class, so that this can be simplified to
```csharp
await Assert.That(sut).ThrowsException().WithMessageMatching(expectedExpression).OnlyIf(!expectMatch);
```

The `OnlyIf` will check the predicate, and if it is `true` it will ensure, that the exception was thrown as specified. If it is `false`, it will instead ensure, that no exception was thrown.

### DIFF
--- a/TUnit.Assertions.Tests/Assertions/Delegates/Throws.OnlyIfTests.cs
+++ b/TUnit.Assertions.Tests/Assertions/Delegates/Throws.OnlyIfTests.cs
@@ -1,0 +1,64 @@
+﻿namespace TUnit.Assertions.Tests.Assertions.Delegates;
+
+public partial class Throws
+{
+    public class OnlyIfTests
+    {
+        [Test]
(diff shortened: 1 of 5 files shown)


thomhurst:
Wondering if this could/should be available for any condition

vbreuss:
> Wondering if this could/should be available for any condition

I am not sure, I understand your question.
Currently this method would be available for all assertions, that an exception was thrown. Is this too much?

PR #883, opened 2024-10-12
Switched to the more performant System.Threading.Lock on .NET 9.0+

MarkCiliaVincenti:
### Please check the following before creating a Pull Request

- If this is a new feature or piece of functionality, have you started a discussion and gotten agreement on it?
- If it fixes a bug or problem, is there an issue to track it? If not, create one first and link it please so there's clear visibility.
- Did you write tests to ensure you code works properly?

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -5,7 +5,7 @@
 
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
-    <LangVersion>latest</LangVersion>
+    <LangVersion>preview</LangVersion>
     <TrimmerSingleWarn>false</TrimmerSingleWarn>
     <EnableTrimAnalyzer>true</EnableTrimAnalyzer>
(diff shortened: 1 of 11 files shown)


thomhurst:
Looks good! Only thing is I'd prefer to keep the dependencies minimal. Any chance we can do a compiler if statement?

If net 8 > object
If net 9 > Lock

MarkCiliaVincenti:
It's a very micro polyfill dependency and will allow you to use the full feature set of System.Threading.Lock not just using the lock keyword.

MarkCiliaVincenti:
In case it's not clear, I am also the author of the library dependency.

MarkCiliaVincenti:
@thomhurst a new version of Backport.System.Threading.Lock has come out that acts as a source generator and basically dropping the DLL as a dependency. If this new development makes you reconsider, I will amend this PR accordingly.

thomhurst:
If it doesn't bring in anything new to the package then sure!

MarkCiliaVincenti:
Applied in #1234

PR #885, opened 2024-10-12
Fix location of `StringDifference` when actual starts with expected

vbreuss:
When the actual value starts with the expected value (but is not identical), the `StringDifference` class incorrectly returned -1 as index of first mismatch.


thomhurst:
Haha I noticed this too earlier!

Went a slightly different approach though. But yours seems simpler

[TUnit.Assertions/Helpers/StringDifference.cs](https://github.com/thomhurst/TUnit/pull/884/files#diff-25b850fbb69dade8eb573e02d5f21551034b7d0bb580b532ac5b6722048fcfcd)

PR #887, opened 2024-10-13
Fix the displayed expression for string assertion `IsEqualTo`

vbreuss:
The displayed expression for the string assertion `IsEqualTo` incorrectly also adds the default `StringComparison.Ordinal`, although it is not provided.


TUnit.Assertions/Extensions/Is/Strings/StringIsExtensions.cs
```diff
@@ -13,7 +13,10 @@ public static class StringIsExtensions
 {
     public static StringEqualToAssertionBuilderWrapper IsEqualTo(this IValueSource<string> valueSource, string expected, [CallerArgumentExpression("expected")] string doNotPopulateThisValue1 = "")
     {
-        return IsEqualTo(valueSource, expected, StringComparison.Ordinal, doNotPopulateThisValue1);
+        var assertionBuilder = valueSource.RegisterAssertion(new StringEqualsExpectedValueAssertCondition(expected, StringComparison.Ordinal)
```
thomhurst: Could we still call the overload but pass in an empty expression?

thomhurst:
Got a merge conflict

vbreuss:
> Got a merge conflict 

I will have a look at it later...

TUnit.Assertions/Extensions/Is/Strings/StringIsExtensions.cs
vbreuss: I tried it, but both, with null or an empty string, the result contained it in the generated expression string, so I had to revert this commit.

thomhurst: Can you try again with the latest main merged in? I believe I've sorted that

vbreuss: I merged in `main` and it seems to work when I provide `null`.

PR #889, opened 2024-10-13
Enable nullability in assertion classes

vbreuss:
Nullability was arbitrarily disabled in some assertion classes, even though the missing annotations were trivial.
Especially as they are "outside-facing" methods, having properly applied nullable annotations, adds great value to users.


thomhurst:
(I'm on mobile ATM so can't check) but I did this initially because if for example you were asserting on a nullable string, then all the specific string extensions would you a warning, because they were specifying a non-null string.

The same would happen if they took in a nullable string and the item being asserted was a non null string.

I don't want this assertion library to be flagging lots of compiler warnings. Have you managed to sort that then?

vbreuss:
I see your point. I would try to have a deeper look in the future...

PR #909, opened 2024-10-15
Reduce string Allocations on the happy path

vbreuss:
Currently on the happy path all messages in the AssertionResult are generated and passed directly in the method. Especially for more complex messages like the string difference can be quite expensive and should only be calculated, when the assertion fails.

This pull request changes the `FailWith` overload to accept a `Func<string>` instead of directly a string and only lazy evaluates it, when needed.


thomhurst:
Out of curiousity how much does a new Func<> allocate?

viceroypenguin:
> Out of curiousity how much does a new Func<> allocate?

24b per instance, IIRC

thomhurst:
> > Out of curiousity how much does a new Func<> allocate?
> 
> 24b per instance, IIRC

Would that make this not really worthwhile then?

vbreuss:
> > Out of curiousity how much does a new Func<> allocate?
> 
> 24b per instance, IIRC

Where do you get this number? As far as I am aware, a `Func` is a delegate and would only allocate, if it needs to capture variables from an outside scope.

viceroypenguin:
> > > Out of curiousity how much does a new Func<> allocate?
> > 
> > 
> > 24b per instance, IIRC
> 
> Where do you get this number? As far as I am aware, a `Func` is a delegate and would only allocate, if it needs to capture variables from an outside scope.

24b is the minimum size of any CLR instance on the heap, and `Func<>` is a CLR instance. It may be a special instance of a `delegate`, but it is an instance nonetheless. The minimum 24b includes: pointer to the type information; pointer to the method table; and minimum of 8b of data.

In further consideration, I think a `Func<>` is actually 32b, since it must contain space for the `this` pointer that the method is based on, as well as the method pointer itself. This is true whether the delegate points to a static method or an instance method, since the consumer of the `Func<>` (or any delegate) cannot know a-priori whether the delegate refers to an instance or not. 

You referenced closures. In the case of a closure, there will actually be _two_ allocations; one for the delegate itself and one for the closure class. These are separate allocations, and the delegate will contain the closure class as the `this` reference in it's data.

Now all of this may be elided automatically in T1 JIT, based on information available, but this is not a guarantee, and should not be relied upon (except in circumstances such as hot-paths where the effort may be justified to prove that a particular arrangement will consistently be JIT'd in a certain way).

viceroypenguin:
Quickly reading the PR, I would recommend instead using interpolated strings. You can review [ISLE](https://github.com/fedarovich/isle) for examples of how to set up interpolated string handlers that will not allocate on one path, but will on another path. 

You can set up the method to receive an interpolated string, decide on the happy path to ignore the handler, and no allocations will be done.

However, I would question whether this effort is actually necessary. Assertions are generally only used in unit tests, and not used in deployed applications, where allocation concerns are necessary. It is generally of no concern (at least to me) how much a unit test allocates.

thomhurst:
Yeah I think keeping the code simpler outweighs any allocation benefits. If the garbage collector needs to kick in it just means your tests might take a few seconds longer which I don't think is a massive deal

vbreuss:
The motivation stems mainly from expectation message like the following:
```csharp
$"found message {Format(actualValue.Message).TruncateWithEllipsis(100)} which {new StringDifference(actualValue.Message, expectedValue)}");
```
Creating this string is not trivial and doing this for every (successful) test in every test run, will slow down large test suites. I think it is important to consider test execution speed...

vbreuss:
I created a benchmark to compare `TUnit`, `fluentassertions` and [my own project](https://github.com/Testably/Testably.Expectations) for the simple assertion, that "foo" is equal to "foo":
```csharp
	private readonly string _stringExpectation = "foo";
	private readonly string _stringSubject = "foo";

	[Benchmark]
	public AndConstraint<StringAssertions> String_FluentAssertions()
		=> _stringSubject.Should().Be(_stringExpectation);

	[Benchmark]
	public async Task<string?> String_TestablyExpectations()
		=> await Expect.That(_stringSubject).Should().Be(_stringExpectation);

	[Benchmark]
	public async Task<string?> String_TUnit()
		=> await Assert.That(_stringSubject).IsEqualTo(_stringExpectation);
```

Here is [the result](https://github.com/Testably/Testably.Expectations/actions/runs/11766595490/job/32774435047):
| Method                      | Mean       | Error   | StdDev  | Gen0   | Allocated |
|---------------------------- |-----------:|--------:|--------:|-------:|----------:|
| String_FluentAssertions     |   389.6 ns | 2.07 ns | 1.94 ns | 0.0215 |    1832 B |
| String_TestablyExpectations |   435.3 ns | 0.86 ns | 0.76 ns | 0.0134 |    1144 B |
| String_TUnit                | 1,418.9 ns | 8.89 ns | 7.88 ns | 0.0381 |    3248 B |


I *think* a lot of it stems from the allocation of the string difference in the happy path...

thomhurst:
@vbreuss Can we see the benchmark with your change also?

vbreuss:
I could only run it locally:

This branch:
| Method                      | Mean     | Error    | StdDev   | Gen0   | Gen1   | Allocated |
|---------------------------- |---------:|---------:|---------:|-------:|-------:|----------:|
| String_FluentAssertions     | 337.9 ns | 33.48 ns | 98.71 ns | 0.2189 |      - |   1.79 KB |
| String_TestablyExpectations | 382.9 ns | 15.81 ns | 46.61 ns | 0.1364 |      - |   1.12 KB |
| String_TUnit                | 507.9 ns |  9.48 ns |  9.74 ns | 0.2975 | 0.0010 |   2.44 KB |

Current `main` branch:
| Method                      | Mean       | Error    | StdDev   | Gen0   | Allocated |
|---------------------------- |-----------:|---------:|---------:|-------:|----------:|
| String_FluentAssertions     |   343.5 ns | 12.03 ns | 33.14 ns | 0.2189 |   1.79 KB |
| String_TestablyExpectations |   294.7 ns |  5.95 ns | 12.28 ns | 0.1364 |   1.12 KB |
| String_TUnit                | 1,035.3 ns | 15.10 ns | 12.61 ns | 0.3891 |   3.19 KB |

thomhurst:
Halving the time is pretty decent actually. Do you wanna rebase this?

viceroypenguin:
I will reiterate:
> Quickly reading the PR, I would recommend instead using interpolated strings. You can review [ISLE](https://github.com/fedarovich/isle) for examples of how to set up interpolated string handlers that will not allocate on one path, but will on another path. 

I'd much prefer to see this (with even fewer allocations) than the `Func<>` version

TUnit.Assertions/AssertConditions/AssertionResult.cs
```diff
@@ -71,9 +84,10 @@ public AssertionResult OrFailIf(Func<bool> isFailed, string message)
         return new AssertionResult(false, message);
     }
 
-    public static AssertionResult Fail(string message) => new(false, message);
+    public static AssertionResult Fail(Func<string> message)
+        => new(false, message);
 
-    public static AssertionResult Passed { get; } = new(true, null);
+    public static AssertionResult Passed { get; } = new(true, () => "");
```
thomhurst:
```suggestion
    public static AssertionResult Passed { get; } = new(true, () => string.Empty);
```

thomhurst:
> I will reiterate:
> 
> > Quickly reading the PR, I would recommend instead using interpolated strings. You can review [ISLE](https://github.com/fedarovich/isle) for examples of how to set up interpolated string handlers that will not allocate on one path, but will on another path.
> 
> I'd much prefer to see this (with even fewer allocations) than the `Func<>` version

Sorry I missed that. I'll have a read through it. I take it it's providing constants and arguments like ILogger? Which then will do a string.format or whatever when needed?

Edit: Hmm it doesn't. It somehow does this magically? I need to dig deeper but I have no idea how that wouldn't allocate initially and how it automatically destructures it!

viceroypenguin:
it's based on this: https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/tutorials/interpolated-string-handler

Basically, you have a chance to not call all of the `.AppendXxx()` and `.ToString()` unless you actually need to.

vbreuss:
> it's based on this: https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/tutorials/interpolated-string-handler
> 
> Basically, you have a chance to not call all of the `.AppendXxx()` and `.ToString()` unless you actually need to.

I am not sure, how this would work in this case. The example you provided is about logging, where I provide existing variables and want to avoid creating a formatted string.
Here, we don't even want the variables to be created. In my example above:
```csharp
$"found message {Format(actualValue.Message).TruncateWithEllipsis(100)} which {new StringDifference(actualValue.Message, expectedValue)}");
```
I would not want to create a `StringDifference` class or call `Format(actualValue.Message).TruncateWithEllipsis(100)`, as both are "expensive" operations. As far as I understood the example in the link you provided, using an interpolated string handler would only avoid creating the combined string with the provided variables...

viceroypenguin:
> > it's based on this: https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/tutorials/interpolated-string-handler
> > Basically, you have a chance to not call all of the `.AppendXxx()` and `.ToString()` unless you actually need to.
> 
> I am not sure, how this would work in this case. The example you provided is about logging, where I provide existing variables and want to avoid creating a formatted string. Here, we don't even want the variables to be created. In my example above:
> 
> ```cs
> $"found message {Format(actualValue.Message).TruncateWithEllipsis(100)} which {new StringDifference(actualValue.Message, expectedValue)}");
> ```
> 
> I would not want to create a `StringDifference` class or call `Format(actualValue.Message).TruncateWithEllipsis(100)`, as both are "expensive" operations. As far as I understood the example in the link you provided, using an interpolated string handler would only avoid creating the combined string with the provided variables...

From the doc I linked:
> You can make one final update to the handler's constructor that improves efficiency. The handler can add a final out bool parameter. Setting that parameter to false indicates that the handler shouldn't be called at all to process the interpolated string expression:

From there, you can see the following changes to the compiled IL.

```cs
var logger = new NullLogger<int>();

logger.LogInformation($"Test{GetStringValue()}");

string GetStringValue()
{
	Thread.Sleep(1_000);
	return "Test";
}
```

turns into:
```cs
  NullLogger<int> logger = new NullLogger<int> ();
   ILogger logger2 = logger;
   ILogger logger3 = logger2;
   bool isEnabled;
   InformationLogInterpolatedStringHandler handler = new InformationLogInterpolatedStringHandler (4, 1, logger2, out isEnabled);
   if (isEnabled)
   {
        handler.AppendLiteral ("Test");
        handler.AppendFormatted (<Main>g__GetStringValue|4_0 (), 0, null, "GetStringValue()");
   }
   Isle.Extensions.Logging.LoggerExtensions.LogInformation (logger3, ref handler);
```

In the case of TUnit, you would similarly return an `out isEnabled` that determines whether to build the string. Basically the assertion comes first, and returns whether passed or failed via the `out`, which is used to determine whether to attempt to build the string, which is done in a separate method entirely.

@thomhurst Read this comment. Has nothing to do with `Append<T>`

thomhurst:
> I would not want to create a StringDifference class or call Format(actualValue.Message).TruncateWithEllipsis(100), as both are "expensive" operations.

Looking at the MS article, you add an `Append<T>` to your interpolation handler. I guess the solution to avoid allocating would be to make your types structs so they're not allocating. Make sure the constructors only set fields and don't do any of the expensive work. And then the expensive work can be done in the .ToString().

Then since that never gets called in a happy path, nothing is allocated?

@viceroypenguin you seem more clued up on this than me, but is that sensible?

thomhurst:
Ah I see! So it all compiles down to be inside an if statement which won't be entered on happy paths. This is clever stuff, thanks for explaining.

Yeah this definitely sounds like the right approach if we want to speed up our assertions and reduce allocations.

@vbreuss sound like something you wanna take on? If not happy to give it a go myself.

vbreuss:
> Quote reply

@thomhurst : I am really intrigued by this and would like to give it a go next weekend, please leave this with me.
I created #1186 for this.

Could we still close this PR, as it is quite an improvement?
Then I could more easily create a benchmark to compare this version with the version with interpolated string handlers, to verify, if and by how much this would speedup the process.

PR #911, opened 2024-10-15
Throw an exception when mixing AND and OR assertions

### LINKED ISSUE #891: Unclear precedence of AND and OR
I couldn't find anything in the [documentation](https://thomhurst.github.io/TUnit/docs/category/tutorial---assertions) with regards to the precedence between [AND](https://thomhurst.github.io/TUnit/docs/tutorial-assertions/and-conditions) and [OR](https://thomhurst.github.io/TUnit/docs/tutorial-assertions/or-conditions).

Consider the following test:
```csharp
[Test]
public async Task And_Or_Precedence_Test()
{
    char[] sut = "CD".ToCharArray();

    await Assert.That(sut)
        .Contains('A').And
        .Contains('B').Or
        .Contains('C').And
        .Contains('D');
}
```

I would have expected the test to fail, as I would expect the condition to be interpreted as `(A && B) || (C && D)`, to be in line with the [operator precedence](https://thomhurst.github.io/TUnit/docs/category/tutorial---assertions) in C#.

However, this test succeeds, because `A && B || C && D` seems to be interpreted as `(((A && B) || C) && D)`.

vbreuss:
Fixes #891 by throwing an exception when AND and OR assertions are mixed.

### DIFF
--- a/TUnit.Assertions/Exceptions/MixedAndOrAssertionsException.cs
+++ b/TUnit.Assertions/Exceptions/MixedAndOrAssertionsException.cs
@@ -1,0 +1,4 @@
+﻿namespace TUnit.Assertions.Exceptions;
+
+public class MixedAndOrAssertionsException()
+    : Exception("Don't mix 'Or' & 'And' operators in assertions.");
--- a/TUnit.Assertions/Exceptions/MixedAndOrAssertionsException.cs
+++ b/TUnit.Assertions/Exceptions/MixedAndOrAssertionsException.cs
@@ -1,4 +1,4 @@
 ﻿namespace TUnit.Assertions.Exceptions;
 
 public class MixedAndOrAssertionsException()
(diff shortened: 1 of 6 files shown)


TUnit.Assertions/Exceptions/MixedAndOrAssertionsException.cs
```diff
@@ -0,0 +1,4 @@
+﻿namespace TUnit.Assertions.Exceptions;
+
+public class MixedAndOrAssertionsException()
+    : Exception("Don't mix 'Or' & 'And' operators in assertions.");
```
thomhurst: Can this inherit from AssertionException?

thomhurst: Could you also mention groups as an alternative?

https://thomhurst.github.io/TUnit/docs/tutorial-assertions/assertion-groups

thomhurst (approved):

PR #923, opened 2024-10-17
using System.Threading.Lock in .NET 9+

BarionLP:
this is an alternative to #883 that uses preprocessors to use Lock in .NET 9+ and object in .NET 8


BarionLP:
The problem reported by Codacy is a false flag

thomhurst:
Thanks @BarionLP !

PR #955, opened 2024-10-19
Allow passing inputs into `MethodDataSourceAttribute`

thomhurst:


viceroypenguin:
@thomhurst capabilities shown here look great. Thank you!

PR #956, opened 2024-10-19
Correct generation of code involving delegates

### LINKED ISSUE #954: [BUG]: `MethodDataSource` referencing an `Action` type
The following code has a syntax error in the generated code:
```cs
public static IEnumerable<Action> ActionData() => [() => { }];
[Test]
[MethodDataSource(nameof(ActionData))]
public void ActionTest(Action action) { }
```

Specifically, it attempts to use the following illegal code:
```cs
var methodInfo = testClassType.GetMethod("ActionTest", 0, [typeof(void global::System.Action())]);
```

NB: This also affects `Action<int>`, and `Func<int, int>`.

viceroypenguin:
Fixes #954

### DIFF
--- a/TUnit.Engine.SourceGenerator.Tests/MethodDataSourceDrivenTests.cs
+++ b/TUnit.Engine.SourceGenerator.Tests/MethodDataSourceDrivenTests.cs
@@ -11,12 +11,15 @@ internal class MethodDataSourceDrivenTests : TestsBase<TestsGenerator>
             "MethodDataSourceDrivenTests.cs"),
         async generatedFiles =>
         {
-            await Assert.That(generatedFiles.Length).IsEqualTo(2);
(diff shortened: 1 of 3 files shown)


viceroypenguin:
nit: When running on my machine, I get a lot of test failures. For example, `ConcreteClass1.AssertClassName()` fails on `ConcreteClass1` being not equal to `Concrete1`.

thomhurst:
Thanks!

> nit: When running on my machine, I get a lot of test failures. For example, `ConcreteClass1.AssertClassName()` fails on `ConcreteClass1` being not equal to `Concrete1`.

Yeah the "TUnit.TestProject" isn't meant to all pass. It's a mixture of passing and failing tests, which are then invoked from the pipeline, and then the pipeline asserts that the test results have the expected number of passed/failed tests.

PR #1005, opened 2024-10-26
Re-register tests dynamically with new data

thomhurst:


TUnit.Core/AsyncConvert.cs
```diff
@@ -21,18 +21,16 @@ public static async Task Convert(Func<ValueTask> action)
...
         if (invoke is Task task)
         {
-            return task;
+            await task;
         }
 
         if (invoke is ValueTask valueTask)
         {
-            return valueTask.AsTask();
+            await valueTask.AsTask();
```
campersau: Looks like `AsTask` isn`t needed anymore?

```suggestion
            await valueTask;
```

Issue #1018, opened 2024-10-26
Test running spins indefinitely when exception thrown from resilience pipeline


Vijay-Braidable:
In VS I'm in debug mode, when I *debug* the test it seems to execute just fine, 100% of the time, when I *run* the test and the resiliancepipeline throws a couple of exceptions (for the multiple failures) it just spins forever (I left a 1 second test for many minutes to check this) unwinding a stacktrace. This did not happen when I was using the non-preview version of VS and XUnit, only when I ported to Preview with TUnit.

Running via dotnet test shows the stacktrace being written to the console for over 5 minutes.

I've speculatively included information I think may be helpful, let me know if I can provide something else.

**TUnit Version**: 0.2.26
**Operating System**: Windows 10 Pro
**Visual Studio Version**: 2022 (640bit) - Preview (17.12.0 Preview 4.0)
**.NET Version**: net8.0
**Extensions**: The only installed extension is ReSharper, I disabled this and the test still failed in the same way.

**The Test**:
```
    [Test]
    [MethodDataSource(typeof(TestBase), nameof(TestData))]
    public async Task TestSomething(SomeType someType)
    {
...
```

**TestData in base class** 
```
    public static IEnumerable<SomeType > TestData()
    {
        yield return ManyTypes.SomeType;
    }
```

**Generated Test Hooks**: 
```
// <auto-generated/>
#pragma warning disable
using global::System.Linq;
using global::System.Reflection;
using global::System.Runtime.CompilerServices;
using global::TUnit.Core;
using global::TUnit.Core.Interfaces;

namespace TUnit.SourceGenerated;

[global::System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage]
file partial class Hooks_REDACTED_Tests : TUnit.Core.Interfaces.SourceGenerator.ITestHookSource
{
    [global::System.Runtime.CompilerServices.ModuleInitializer]
    public static void Initialise()
    {
        var instance = new Hooks_REDACTED_Tests();
        SourceRegistrar.RegisterTestHookSource(instance);
    }
    public global::System.Collections.Generic.IReadOnlyList<StaticHookMethod<TestContext>> CollectBeforeEveryTestHooks()
    {
        return
        [
        ];
    }
    public global::System.Collections.Generic.IReadOnlyList<StaticHookMethod<TestContext>> CollectAfterEveryTestHooks()
    {
        return
        [
        ];
    }
[...]

PR #1020, opened 2024-10-26
Added ReferenceHandler.Preserve to avoid cyclic exceptions outputting…

Vijay-Braidable:
Some Exceptions are cyclic, when this occurs and the test fails, the test handler appears to dump the exceptions endlessly (in a loop) to the console.

* https://github.com/thomhurst/TUnit/issues/1018

It didn't seem appropriate to create tests given there are no tests currently specifically for serialization, it is only tested by proxy in the RPC tests.


Vijay-Highlo:
Closing due to this not fixing the core issue it was raised for.

PR #1088, opened 2024-10-29
Remove unused HasTimeoutAttribute from HooksDataModel

campersau:


TUnit.UnitTests/TestExtensionsTests.cs
```diff
@@ -20,7 +20,6 @@ public void TopLevelClass()
             .Without(x => x.ClassAttributes)
             .Without(x => x.DataAttributes)
             .Without(x => x.TestAttributes)
-            .Without(x => x.ParallelLimit)
```
campersau, line 23: Test fails because it is now internal an cannot be set.
https://github.com/thomhurst/TUnit/commit/2038a56ae4fb3c6195f76b0e6ef99d06e9a6b54d#diff-1a62fba055d60bf3bfe524a4823e438acc34253b991a6aeef02b949ab007c763R70

thomhurst: Yep my bad I pushed to the wrong branch 🤦 Should be fixed on the latest

PR #1159, opened 2024-11-05
Parallel Group

thomhurst:


TUnit.Engine/Services/TestsExecutor.cs
```diff
@@ -92,6 +95,15 @@ await testsToProcess
             .ProcessInParallel(_maximumParallelTests);
     }
 
+    private async Task ProcessParallelGroups(IDictionary<string, List<DiscoveredTest>> groups,
+        ITestExecutionFilter? filter, ExecuteRequestContext context)
+    {
+        foreach (var (_, value) in groups)
+        {
+            await ProcessParallelTests(value.ToQueue(), filter, context);
```
campersau, line 103: `.ToQueue()` isn't needed since `ProcessParallelTests` now accepts an `IEnumerable`.

PR #1162, opened 2024-11-05
Remove unnecessary ToQueue call

campersau:
See https://github.com/thomhurst/TUnit/pull/1159/files#r1829940136


thomhurst:
Eagle eyed! Thanks @campersau

PR #1219, opened 2024-11-16
Fix Equivalent comparer if using different object types

thomhurst:


TUnit.Assertions/Compare.cs
```diff
@@ -106,25 +117,26 @@ public static IEnumerable<ComparisonFailure> CheckEquivalent<T>(T actual, T expe
...
                 yield return comparisonFailure;
             }
         }
         
-        foreach (var propertyInfo in actual.GetType().GetProperties().Concat(expected!.GetType().GetProperties())
+        foreach (var propertyName in actual.GetType().GetProperties().Concat(expected.GetType().GetProperties())
                      .Distinct()
-                     .Where(p => p.GetIndexParameters().Length == 0))
+                     .Where(p => p.GetIndexParameters().Length == 0)
+                     .Select(x => x.Name))
```
campersau, line 129: Should this `Distinct` also be moved to the end here?

thomhurst: #1220

PR #1224, opened 2024-11-16
VS preview is no longer required

campersau:
VS 17.12 got released this week which supports the preview feature.
Updated text similar to Rider.


thomhurst:
Cheers! I think there's a markdown file in the docs folder too they might need tweaking

PR #1231, opened 2024-11-17
Only run After hooks if the Before hooks ran.

thomhurst:
@campersau Running only if we know the setups were run.


campersau:
Looks like this addresses point 3 but what about point 2 from https://github.com/thomhurst/TUnit/commit/3286de308d91246e6d999dc016b883d213547116#commitcomment-149172613 ?
If you have two tests one with `SkipAttribute` and another one manually throws a `SkipTestException` then the `TearDown` also isn't called because the `SkipAttribute` one does not get decreased in the instance tracker.

> Because `ExecuteStaticAfterHooks` is only called in `ExecuteTest`
https://github.com/thomhurst/TUnit/blob/4fb354f2b1da47537982e6c93ce82202ade94087/TUnit.Engine/Services/SingleTestExecutor.cs#L225
Which means `TearDown` isn't called.

thomhurst:
@campersau Hopefully all better in 0.3.20

PR #1233, opened 2024-11-17
ITestSkippedEventReceiver.cs

thomhurst:


campersau:
What about these skips?

https://github.com/thomhurst/TUnit/blob/21073887b0dbdacd9976e10276a7da314a87bfef/TUnit.Engine/Services/SingleTestExecutor.cs#L76-L86

Or is it just meant for manually thrown `SkipTestException`s inside of a test?

thomhurst:
@campersau Yep I still want those to produce events too. I've just re-jigged some of the logic so it should all hit the same catch/finally logic block now. This should hopefully fix the other issue with teardowns as well I think.

campersau:
Yes, looks good when using both PRs!

PR #1260, opened 2024-11-20
docs: update method-data-source.md example to not use `func<..>`

jacob7395:
Update the docs for `docs/docs/tutorial-basics/method-data-source.md` to build/work by removing the `func<..>` syntax.

Maybe it's overkill but could have these example as part of the test suit, then use a build action to copy them into files referenced in the docs?

Not needed for this but could help prevent the issue in future.

### DIFF
--- a/docs/docs/tutorial-basics/method-data-source.md
+++ b/docs/docs/tutorial-basics/method-data-source.md
@@ -12,7 +12,7 @@ If we want test data represented in the form of objects, or just to use somethin
 - If you pass in one argument, this is the method name containing your data. TUnit will assume this is in the current test class.
 - If you pass in two arguments, the first should be the `Type` of the class containing your test source data method, and the second should be the name of the method.
 
-If methods are returning reference types, they should return a `Func<T>` rather than just a `T` - This ensures each test has its own instance of that object and tests aren't sharing objects which could lead to unintended side effects.
+If methods are returning reference types, they should return a `T` rather than just a `T` - This ensures each test has its own instance of that object and tests aren't sharing objects which could lead to unintended side effects.
 
 Here's an example returning a simple object:
 
@@ -27,9 +27,9 @@ public record AdditionTestData(int Value1, int Value2, int ExpectedResult);
 
 public static class MyTestDataSources
 {
-    public static Func<AdditionTestData> AdditionTestData()
+    public static AdditionTestData AdditionTestData()
     {
-        return () => new AdditionTestData(1, 2, 3);
+        return new AdditionTestData(1, 2, 3);
     }
 }
 
@@ -62,9 +62,9 @@ namespace MyTestProject;
 
 public static class MyTestDataSources
 {
-    public static Func<(int, int, int)> AdditionTestData()
+    public static (int, int, int) AdditionTestData()
     {
-        return () => (1, 2, 3);
+        return (1, 2, 3);
     }
 }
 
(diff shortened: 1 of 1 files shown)


jacob7395:
I probably should have read the docs first dose this no longer apply;

```
If methods are returning reference types, they should return a `Func<T>` rather than just a `T` - This ensures each test has its own instance of that object and tests aren't sharing objects which could lead to unintended side effects.
```

I tried the example bellow and everything is fine;

```c#
public class AdditionTestData
{
    public int Value1 { get; }
    public int Value2 { get; }
    public int ExpectedResult { get; }

    public AdditionTestData(int value1, int value2, int expectedResult)
    {
        Value1 = value1;
        Value2 = value2;
        ExpectedResult = expectedResult;
    }
}

public static class MyTestDataSources
{
    public static AdditionTestData AdditionTestData()
    {
        return new AdditionTestData(1, 2, 3);
    }
}

public class MyTestClass
{
    [Test]
    [MethodDataSource(typeof(MyTestDataSources), nameof(MyTestDataSources.AdditionTestData))]
    public async Task MyTest(AdditionTestData additionTestData)
    {
        var result = Add(additionTestData.Value1, additionTestData.Value2);

        await Assert.That(result).IsEqualTo(additionTestData.ExpectedResult);
    }

    private int Add(int x, int y)
    {
        return x + y;
    }
}
```

Are there some types this is still required for? and in those cases should your method take Func<T> like bellow;

```c#
public class AdditionTestData
{
    public int Value1 { get; }
    public int Value2 { get; }
    public int ExpectedResult { get; }

    public AdditionTestData(int value1, int value2, int expectedResult)
    {
        Value1 = value1;
        Value2 = value2;
        ExpectedResult = expectedResult;
    }
}

public static class MyTestDataSources
{
    public static IEnumerable<Func<AdditionTestData>> AdditionTestData()
    {
        yield return () => new AdditionTestData(1, 2, 3);
    }
}

public class MyTestClass
{
    [Test]
    [MethodDataSource(typeof(MyTestDataSources), nameof(MyTestDataSources.AdditionTestData))]
    public async Task MyTest(Func<AdditionTestData> testDataBuilder)
    {
        var additionTestData = testDataBuilder.Invoke();
        
        var result = Add(additionTestData.Value1, additionTestData.Value2);

        await Assert.That(result).IsEqualTo(additionTestData.ExpectedResult);
    }

    private int Add(int x, int y)
    {
        return x + y;
    }
}
```.

jacob7395:
Reading it again I guess this is to trying to prevent tests modifying a shared reference (just do immutability bro), in this case is the `MethodDataSource` no longer supporting `func<..>` a bug, if you try one of the example in the current docs it give's a build error.

thomhurst:
> in this case is the MethodDataSource no longer supporting func<..> a bug, if you try one of the example in the current docs it give's a build error.

Which example?

jacob7395:
I am missing something, maybe I did encounter an issue since there are tests for this in `TUnit.TestProject` and although I get errors before building it builds fine.

jacob7395:
> > in this case is the MethodDataSource no longer supporting func<..> a bug, if you try one of the example in the current docs it give's a build error.
> 
> Which example?

When I copy any of the example into Rider I get the bellow error in my IDE, but they run and build fine?

image

jacob7395:
Okay, I think the issue I was getting was due to my main project being out of date.

I was trying to replicate the issue in an example project when I get the above error that threw me off. Updating my package solved the build error on my main project.

thomhurst:
I've found sometimes updating packed with analyzers requires restarting the ide to reload the analyzers with the new logic

jacob7395:
Will close this as this wasn't an issue, the IDE error shown above goes away once the test is run. Maybe something to look into but I am guessing just something to do with how the tests are generated.

PR #1271, opened 2024-11-22
Make default running examples without code coverage and trx reports

### LINKED ISSUE #1270: Unknown option --report-trx and --coverage
If you follow the documentation, it doesn't work.

To reproduce:
```shell
dotnet new console --name TUnitDocs
cd .\tunitdocs\
rm .\Program.cs
```
Add a new file, as per docs with
```csharp
using TUnit.Core;

namespace TUnitTest;

public class MyTestClass 
{
    [Test]
    public async Task MyTest()
    {
        var result = Add(1, 2);

        await Assert.That(result).IsEqualTo(3);
    }

    private int Add(int x, int y)
    {
        return x + y;
    }
}
```
Then try to run, as per https://thomhurst.github.io/TUnit/docs/tutorial-basics/running-your-tests, `dotnet run -c Release --report-trx --coverage`

This gives the error:

```shell
TUnit v0.3.34.0 | 64-bit | Microsoft Windows 10.0.22631 | win-x64 | .NET 9.0.0 | Microsoft Testing Platform v1.4.3

Unknown option '--report-trx'
Unknown option '--coverage'
```

The fix would be to run:
```shell
dotnet add package Microsoft.Testing.Extensions.TrxReport
dotnet add package Microsoft.Testing.Extensions.CodeCoverage
```

Maybe it'd be better to have this "running-your-tests" page without the codecov and report, but instead add better instructions to https://thomhurst.github.io/TUnit/docs/extensions/? I can make a PR if you agree.

evolvedlight:
Fixes #1270
It doesn't make sense to remove the options completely, as that might be useful, but at least they aren't the default, given that following the "5 minute tutorial" doesn't include adding these

### DIFF
--- a/docs/docs/extensions/extensions.md
+++ b/docs/docs/extensions/extensions.md
@@ -10,10 +10,28 @@ Here will list a few of them.
 ## Code Coverage
 Code coverage is provided via the `Microsoft.Testing.Extensions.CodeCoverage` NuGet package.
 
+Install:
+```
+dotnet add package Microsoft.Testing.Extensions.CodeCoverage
+```
+Then you can run your tests with the `--coverage` flag.
+```
+dotnet run --configuration Release --coverage
+```
+
 [More information can be found here](https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-platform-extensions-code-coverage)
 
 ## TRX Test Reports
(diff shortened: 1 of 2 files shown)


thomhurst:
Thanks!

PR #1339, opened 2024-12-02
Capture and Surface Exceptions thrown within AssertionScope

thomhurst:


TUnit.Assertions/AssertionScope.cs
```diff
@@ -7,14 +7,27 @@ internal class AssertionScope : IDisposable
 {
     private static readonly AsyncLocal<AssertionScope?> CurrentScope = new();
     private readonly AssertionScope? _parent;
-    private readonly List<AssertionException> _exceptions = [];
+    private readonly List<Exception> _exceptions = [];
 
+    static AssertionScope()
+    {
+        AppDomain.CurrentDomain.FirstChanceException += InterceptException;
```
afscrome, line 14: The first chance exception event will pick up both handled and unhandled exceptions, so I imagine it'll count the Exception thrown below, even though it's handled.

```cs
        using (Assert.Multiple())
        {
            try
            {
                throw new Exception();
            }
            catch
            {
                // Handle event without re=throwing
            }
            await Assert.That(1).IsEqualTo(1)
        }
```

thomhurst: Doesn't really matter - It'll just add it to the assertion scope, but we've already exited that assertion scope so it'll just be a no-op

thomhurst: Oh wait I misunderstood. Let me check this.

thomhurst: Yeah you are right. There is no way for me to determine if it was caught which is annoying. I've included exceptions thrown in scopes in the message anyway, with a caveat saying this may or may not have been caught. Seems it's the best I can do 😢

PR #1419, opened 2024-12-16
Logo

thomhurst:


viceroypenguin:
Nice logo!

agray:
Where did you get the green tick icon from?

Surprisingly, Haven't been able to find an exact match

image

thomhurst:
https://www.iconpacks.net/free-icon/green-check-mark-verified-circle-16223.html

agray:
I suggest your logo should be just this:

image

Looks better imo. More like NUnit and XUnit:

NUnit
image

My Logo is 

image

https://github.com/Sauceforge/Saucery

PR #1436, opened 2024-12-20
Add .NET Standard 2.0 target

AArnott:
As discussed in https://github.com/thomhurst/TUnit/discussions/1415, here is the change to adds .NET Standard 2.0 targeting.


AArnott:
@thomhurst I got the 3 projects we discussed building. I haven't done anything with the tests to ensure they test the functionality on .NET Framework, which I expect is important. There is also some odd build break that is related to the source generator.

Can you take it from here?

thomhurst:
Thanks @AArnott ! I'll take a look when I get some time

thomhurst:
@AArnott I'm getting errors around attributes and generic types?

Are you sure they're supported in net framework world?

> System.NotSupportedException: Generic types are not valid.
          at System.RuntimeTypeHandle.CreateCaInstance(RuntimeType type, IRuntimeMethodInfo ctor)
          at System.Reflection.CustomAttribute.GetCustomAttributes(RuntimeModule decoratedModule, Int32 decoratedMetadataToken, Int32 pcaCount, RuntimeType attributeFilterType, Boolean mustBeInheritable, IList derivedAttributes, Boolean isDecoratedTargetSecurityTransparent)

AArnott:
Well, crap. I was about to send you [this](https://github.com/AArnott/Nerdbank.MessagePack/blob/13f027fe71cc5b046191d0317f29cea64169326d/test/Nerdbank.MessagePack.Tests/MessagePackSerializerTests.cs#L500) as proof that generic attributes work on .NET Framework because that code does indeed compile against net472 (when using the latest compiler).
However when I actually run code that reflects over that attribute, it fails on .NET Framework like it does for you.

I guess I never hit this before because these generic attributes are not in fact used at runtime, but rather at compile time by a source generator. So if you need them by a source generator only, you're good. But if you need to reflect over them, your SOL.

I'm so sorry to have misled you.

AArnott:
FWIW, you _could_ workaround this by using System.Reflection.Metadata to read the generic attributes instead of System.Reflection. Given how much work you've already put into adding .NET Framework support (thank you!) I suspect it would be worthwhile to help you get the 'last mile' of the work done.
But S.R.M. isn't good for any kind of _execution_ of code inside the attribute. It is strictly for looking at the attribute and the values of the arguments and properties set on it.

I've used S.R.M. quite a bit, so if you need help, I can.

thomhurst:
> But S.R.M. isn't good for any kind of execution of code inside the attribute

This sounds like a blocker. A lot of the attributes are used to drive execution logic

AArnott:
I wonder if we could get creative. .net reflection has some ways to executing methods and instantiating types that would be independent of the fact that they had been used as attributes. With a little bit of system reflection metadata Kung Fu and some reflection, I think we might be able to get execution to work for you. I'd be willing to spend some time experimenting if you'd like.

thomhurst:
Sure, whenever you get some time

AArnott:
Early work on this is promising. On .NET Framework I'm able to use System.Reflection.Metadata to find the attribute and get a System.Type object that represents it (with its type argument).
Now I just have to decode the arguments and named property assignments in order to instantiate the attribute, at which point what I have is a .NET Framework compatible attribute reflection API that supports generic attributes.

thomhurst:
I've just been thinking we could probably actually side step this. The problem is I'm calling reflection APIs "GetCustomAttribute". However, since attributes require constant values, I have all the information available to me in the source generation context. I can just extract the type and the arguments, and "new" it up myself. The framework doesn't even need to 'know' it's an attribute at all

AArnott:
Well good.

FWIW, I have my prototype fully working. It could be exposed as an extension method alongside the other `GetCustomAttribute` methods and return attributes just like the rest, except it would work with generic attributes too. So let me know if you want to try it. Otherwise I'll just archive it somewhere.

AArnott:
Here it is, FYI: https://gist.github.com/AArnott/acaca32ab812b9f97fd2f5fc7301b1e9

It certainly hasn't been battle-hardened. e.g. if the attribute is declared in a different assembly from the one that the attribute is applied inside, it would probably break. I expect that all such issues would be readily fixable though, given the need.

thomhurst:
@AArnott looks like I've finally got it working, with the caveat of users need to bring in a poly fill library or add a definition for the ModuleInitializerAttribute to allow the source generated code to register itself

AArnott:
That is awesome. Your source generator could look to see if the module initializer attribute was already defined in the compilation and if not it could define it itself.

AArnott:
That is awesome. Your source generator could look to see if the module initializer attribute was already defined in the compilation and if not it could define it itself.

thomhurst:
> That is awesome. Your source generator could look to see if the module initializer attribute was already defined in the compilation and if not it could define it itself.

I was thinking that but if other polyfill libraries are doing the same it'd cause type conflicts. And source generators can't see other source generated code so I wouldn't be able to counter that scenario. So might just have to be a caveat that users define or polyfill it themselves.

AArnott:
Good point.

thomhurst:
@AArnott this should now be working in 0.6.0

AArnott:
I'll try it out on Nerdbank.MessagePack right away. Thanks.

AArnott:
FYI I was able to get my analyzers test project (which is the smaller of my two test projects) running on TUnit. It required [several workarounds](https://github.com/aarnott/Nerdbank.MessagePack/compare/tunit#diff-edbae5a493b5970907d0156eda51bf4e6744cd25664b123f2f91d84d1181af8b) however.

PR #1451, opened 2024-12-24
manually construct attributes

thomhurst:


thomhurst:
@AArnott this should help migrate to netstandard, and also since we're not using reflection we may actually get speed improvements

Issue #1460, opened 2024-12-26
.NET Standard 2.0 target should avoid a System.Text.Json 9.0.0 dependency


AArnott:
The package dependency only shows up in the netstandard2.0 target framework anyway, and by referencing 9.0.0, even .NET 8 test projects were required to reference STJ 9 rather than the one that comes with the runtime.
Given tests should be testing the library in their natural environment, we should avoid forcing a test to update dependencies beyond what the library under test would naturally require.

Referencing STJ 9 is fine once .NET 9 is the oldest supported runtime by Microsoft (in about a year).

viceroypenguin:
Nit: net9 will never be the oldest supported framework. Net8 support will continue until net11 is released.

thomhurst:
Are these libraries actually tightly coupled to the .NET version? If it's only in .net standard anyway it won't be affecting net core runtimes.

viceroypenguin:
> Are these libraries actually tightly coupled to the .NET version? If it's only in .net standard anyway it won't be affecting net core runtimes.

Correct. STJ is not considered tightly coupled to the net sdk, and developers are encouraged to use newer STJ versions with older sdk.

AArnott:
> If it's only in .net standard anyway it won't be affecting net core runtimes.

It does, actually.
Consider a .NET Standard 2.0 library that consumes this one. It now requires STJ 9.
Now a .NET 8 application project comes along and references the .NET Standard 2.0 library. Now that .NET 8 app must ship STJ 9, instead of relying on the one that .NET 8 ships. This makes the app larger. It also increases the servicing onus on the app author because now the (relatively regular) servicing updates to STJ have to ship in new app servicing releases instead of just when the app user keeps their .NET 8 runtime current.

viceroypenguin:
> It also increases the servicing onus on the app author because now the (relatively regular) servicing updates to STJ have to ship in new app servicing releases instead of just when the app user keeps their .NET 8 runtime current.

I'm not sure what you mean here, since both net8 and net9 have monthly servicing releases, and STJ follows the SDK release schedule.

AArnott:
@viceroypenguin A .NET 8 app that does _not_ ship System.Text.Json.dll (because it uses v8 which is included in the runtime) has no servicing responsibility for that dll, because the user that installed the app is responsible to keep their runtime updated.

[...]

PR #1462, opened 2024-12-26
Utilize dependabot to help keep dependencies current

AArnott:
I include a special bit for System.Text.Json to prevent major version upgrades (while allowing minor updates) to be consistent with .NET Standard 2.0 target should avoid a System.Text.Json 9.0.0 dependency #1460, though I haven't included #1461 directly in this one so you can make the decisions regarding the PRs independently.


thomhurst:
I actually use renovate to keep dependencies up to date: https://github.com/thomhurst/RenovateConfig/

AArnott:
Oh! I've never heard of Renovate before. Thanks for exposing me to something new!

PR #1548, opened 2025-01-13
chore: Simplified InternalsVisibleTo

samtrion:
Minor tidying-up measure to reduce complexity.


thomhurst:
Nice didn't know about that tag

PR #1549, opened 2025-01-13
chore: Simplified usage of NuGet Package Microsoft.SourceLink.GitHub

samtrion:
Minor tidying-up measure to reduce complexity and reducing this to only one file.

A [GlobalPackageReference](https://learn.microsoft.com/en-us/nuget/consume-packages/Central-Package-Management#global-package-references) is automatically referenced by all projects within the solution including `PrivateAssets=‘All’`.


thomhurst:
Thanks!

PR #1604, opened 2025-01-18
Add Assertions on Enumerables taking a Func matcher

Genmutant:
I added a first basic implementation to use Funcs to find matches in IEnumerables. Currently missing some more of those (like NotContains) and nicer error messages.


thomhurst:
If I merge this would you be happy to continue and add the NotContains etc?

Genmutant:
Sure. But might have to wait until next week ;-)

thomhurst:
Sure thing 😄

PR #1613, opened 2025-01-18
Interpolated String Handler for Assertions to reduce allocations

thomhurst:


thomhurst:
@viceroypenguin does this look right to you?

TUnit.Assertions/AssertConditions/AssertionResult.cs
```diff
@@ -1,32 +1,38 @@
...
-        get
-        {
-            return _message.Value;
-        }
-    }
-    private readonly Lazy<string> _message;
+    public string Message { get; }
 
-    private AssertionResult(bool isPassed, Func<string> messageGenerator)
+    private AssertionResult(bool isPassed, string messageGenerator)
```
viceroypenguin: Recommend correcting this parameter name back to the original `message` name

viceroypenguin (approved):
Yes, this looks correct!

Decompiling `ExceptionMessageContainingExpectedValueAssertCondition`1.GetResult()` looks like:
```cs
	protected override AssertionResult GetResult(TException? actualValue, string? expectedValue)
	{
		if (actualValue?.Message == null)
		{
			return AssertionResult.FailIf(expectedValue != null, "the exception message was null");
		}
		AssertionResult assertionResult = AssertionResult.FailIf(expectedValue == null, "expected value was null");
		bool flag = !actualValue.Message.Contains(expectedValue, <stringComparison>P);
		bool isFailed = flag;
		bool enabled;
		AssertionResult.InterpolatedStringHandler message = new AssertionResult.InterpolatedStringHandler(14, 1, flag, out enabled);
		if (enabled)
		{
			message.AppendLiteral("found message ");
			message.AppendFormatted(Formatter.Format(actualValue.Message).TruncateWithEllipsis(100));
		}
		return assertionResult.OrFailIf(isFailed, message);
	}
```

which, as you may see, does not actually add things to the message unless `flag` is `true`.

Since the handler is passed as is to the `OrFailIf()`, the `GetFormattedText()` is also only called if the `flag` is `true`, which means the final string is only generated on failure conditions.

thomhurst:
> Yes, this looks correct!
> 
> Decompiling `ExceptionMessageContainingExpectedValueAssertCondition<T>.GetResult()` looks like:
> 
> ```cs
> 	protected override AssertionResult GetResult(TException? actualValue, string? expectedValue)
> 	{
> 		if (actualValue?.Message == null)
> 		{
> 			return AssertionResult.FailIf(expectedValue != null, "the exception message was null");
> 		}
> 		AssertionResult assertionResult = AssertionResult.FailIf(expectedValue == null, "expected value was null");
> 		bool flag = !actualValue.Message.Contains(expectedValue, <stringComparison>P);
> 		bool isFailed = flag;
> 		bool enabled;
> 		AssertionResult.InterpolatedStringHandler message = new AssertionResult.InterpolatedStringHandler(14, 1, flag, out enabled);
> 		if (enabled)
> 		{
> 			message.AppendLiteral("found message ");
> 			message.AppendFormatted(Formatter.Format(actualValue.Message).TruncateWithEllipsis(100));
> 		}
> 		return assertionResult.OrFailIf(isFailed, message);
> 	}
> ```
> 
> which, as you may see, does not actually add things to the message unless `flag` is `true`.
> 
> Since the handler is passed as is to the `OrFailIf()`, the `GetFormattedText()` is also only called if the `flag` is `true`, which means the final string is only generated on failure conditions.

Amazing! Thanks for all your help on this 😄

TUnit.Assertions/AssertConditions/AssertionResult.cs
thomhurst: Done! 😄

vbreuss:
Wow, I really like how this reads much nicer. I learned a lot from looking over it. Thanks! 👍

Issue #1629, opened 2025-01-19
Null Reference Exception in Throws


MrKWatkins:
The following test:

```
[Test]
public async Task Test()
{
    await Assert.That(() => { }).Throws<InvalidOperationException>();
}
```

Throws a NullReferenceException rather than the expected test failure. I have a PR to fix it if you want; you'll need to grant me permission to add PRs.

Alternatively it's a quick two character fix if you want to do it yourself. Just change [ThrowsOfTypeAssertCondition line 13](https://github.com/thomhurst/TUnit/blob/dca591fdd74ead8dd38987afdfc2fa1b3849e638/TUnit.Assertions/Assertions/Throws/ThrowsOfTypeAssertCondition.cs#L13) to `exception?.GetType()`, i.e. ? instead of !. The same issue affects ThrowExactly, so change the same on [ThrowsExactlyAssertCondition line 15](https://github.com/thomhurst/TUnit/blob/dca591fdd74ead8dd38987afdfc2fa1b3849e638/TUnit.Assertions/Assertions/Throws/ThrowsExactlyAssertCondition.cs#L15).

The weird thing is you already have tests for this, OfTypeTests.Fails_For_Code_Without_Exceptions and ExactlyTests.Fails_For_Code_Without_Exceptions. When I checked out the solution and ran them locally they failed for me. You might want to double check they're running as part of the build, given that it passes.

thomhurst:
If you've got a fix raise it as a pr and I'll look at the tests separately :)

MrKWatkins:
Done, https://github.com/thomhurst/TUnit/pull/1630.

PR #1630, opened 2025-01-19
Fix NullReferenceException in Throws and ThrowsExactly.

MrKWatkins:
Fixes https://github.com/thomhurst/TUnit/issues/1629.


vbreuss:
When looking where [`OrFailIf`](https://github.com/search?q=repo%3Athomhurst%2FTUnit%20OrFailIf&type=code) occurs, this seems like a general problem that was probably introduced with #1613.

PR #1686, opened 2025-01-26
Enumerable satisfy

Genmutant:
Adds the corresponding AllSatisfy to the Satisfies Assertions, fixes #1575.
I tried to add a nice assertion message, which makes the implementation a bit less straightforward than I initially thought.
Would be great to be merged, pending any changes you want (me) to make.


TUnit.Assertions/Assertions/Generics/GenericSatisfiesExtensions.cs
```diff
@@ -26,4 +26,39 @@ public static InvokableValueAssertionBuilder<TActual> Satisfies<TActual, TExpect
         return valueSource.RegisterAssertion(new SatisfiesAssertCondition<TActual, TExpected>(asyncMapper, assert, mapperExpression, assertionBuilderExpression),
             [mapperExpression, assertionBuilderExpression]);
     }
+    
+    public static InvokableValueAssertionBuilder<TActual> AllSatisfy<TActual, TInner, TExpected>(
```
thomhurst, line 30: I think this might read better:

```suggestion
    public static InvokableValueAssertionBuilder<TActual> SatisfiesAll<TActual, TInner, TExpected>(
```

TUnit.Assertions/AssertConditions/EnumerableSatisfiesAssertCondition.cs
```diff
@@ -0,0 +1,62 @@
...
+            return AssertionResult.Fail("is null");
+        }
+        var mergedAsserts = AssertionResult.Passed;
+        var i = 0;
+        foreach (var itemValue in actualValue)
+        {
+            var currentIndex = i++;
+            var assertionResult = await GetResult(itemValue, exception);
+            if (assertionResult.IsPassed) continue;
+            var failMessage = mergedAsserts.IsPassed ? "items not satisfying the condition where found:" : mergedAsserts.Message;
```
thomhurst:
```suggestion
            var failMessage = mergedAsserts.IsPassed ? "items not satisfying the condition were found:" : mergedAsserts.Message;
```

Genmutant: whoops

TUnit.Assertions/Assertions/Generics/GenericSatisfiesExtensions.cs
Genmutant: I thought that might imply that a single item might satisfy multiple conditions, while the method checks all items if the satisfy a single assertion chain?

thomhurst: Ah yeah, that makes sense.

TUnit.Assertions.Tests/SatisfiesTests.cs
```diff
@@ -207,6 +209,69 @@ at Assert.That(myModel).Satisfies(model => model.Nested, assert =>
...
+
+    [Test]
+    public async Task All_Satisfy_DirectValue_Good()
+    {
+        var myModel = new MyModel { Value = "Hello" };
+        var myModel2 = new MyModel { Value = "World" };
+        var myModel3 = new MyModel { Value = "!" };
+        List<MyModel> models = [myModel, myModel2, myModel3];
+
+        await Assert.That(models).AllSatisfy((IValueSource<MyModel?> assert) => assert.IsNotNull());
```
thomhurst, line 231: I know the type inference has been an issue for generic inner types. Don't suppose there's a way we can avoid having to specify `IValueSource` here is there?

thomhurst: Thinking about it I don't think there is. But they can use the above overload.

thomhurst:
Thanks for the contribution!

TUnit.Assertions.Tests/SatisfiesTests.cs
Genmutant: I tried a view ways to get it to recognize it, but couldn't get it to work. I think there is currently no way with the way the assertions are set up. If something changes in the future and it can be done nicer, that shouldn't be a breaking change I think?

vbreuss: @thomhurst, @Genmutant  :
Actually, there is. If you would use covariance on the [`IValueSource`](https://github.com/thomhurst/TUnit/blob/a346c54efa02a2795b59f735cddc6a05bb2c213e/TUnit.Assertions/AssertConditions/Interfaces/IValueSource.cs#L5) interface, you can add different overloads, including one that encapsulates the type, so you could write:
```csharp
public static Elements<TItem> All<TItem>(this IValueSource<IEnumerable<TItem>> subject)
{
//...
}
// with an Elements class:
public class Elements<TItem>
{
    public InvokableValueAssertionBuilder<TItem> Satisfy(
        Func<IValueSource<TItem?>, InvokableAssertionBuilder<TItem?>> assert,
        [CallerArgumentExpression(nameof(assert))] string assertionBuilderExpression = "")
    {
       // ...
    }
}
```

It then reads a bit differently:
```csharp
IEnumerable<MyClass> subject = [];

await Assert.That(subject).All().Satisfy(x => x.IsNotNull());
```

I used a similar mechanism in [aweXpect](https://awexpect.com/docs/expectations/collections#all-satisfy). You can have a look there 😃

thomhurst: @vbreuss thanks for the tip! I've just done some work to make the generic covariant. Ended up being a lot of files changed haha. But can see the type inference is much better now! Can create extensions on `IEnumerable<T>` instead of `T : where IEnumerable` which makes life a lot easier!

vbreuss: I am glad it worked 👍

PR #1792, opened 2025-02-08
Update docs for class-data-source with updated `SharedType` enums

MarcelMichau:
Hi there!

First of all, thank you for making this library - I have immensely enjoyed writing tests using it. :)

Reading the docs for [Injectable Class Data Source](https://thomhurst.github.io/TUnit/docs/tutorial-basics/class-data-source), I noticed that the `SharedType` enums have not yet been updated with the enums in the current version of the package. This PR updates the docs to match [SharedType.cs](https://github.com/thomhurst/TUnit/blob/09048285c05a3e43f2bb4ff211afe1448bc28402/TUnit.Core/Attributes/TestData/SharedType.cs#L3)

The descriptions of the enums might not be entirely correct, please let me know if there is anything which can be changed/improved.

### DIFF
--- a/docs/docs/tutorial-basics/class-data-source.md
+++ b/docs/docs/tutorial-basics/class-data-source.md
@@ -18,11 +18,14 @@ Options are:
 ### Shared = SharedType.None
 The instance is not shared ever. A new one will be created for you.
 
-### Shared = SharedType.Globally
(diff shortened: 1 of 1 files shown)


thomhurst:
Thanks :smile:

PR #1797, opened 2025-02-08
`CompletesWithin(TimeSpan)` and `ThrowsWithin(TimeSpan)` assertions

thomhurst:
cc @AlexandreBossard

### DIFF
--- a/TUnit.Assertions.Tests/ExecutionTimeTests.cs
+++ b/TUnit.Assertions.Tests/ExecutionTimeTests.cs
@@ -1,0 +1,46 @@
+﻿using TUnit.Assertions.AssertConditions;
+using TUnit.Assertions.AssertConditions.Throws;
+using TUnit.Assertions.Assertions.Delegates;
+
+namespace TUnit.Assertions.Tests;
+
+public class ExecutionTimeTests
+{
+    [Test]
+    public async Task Completes_Within_Happy()
+    {
+        var action = () => Thread.Sleep(100);
+
+        await Assert.That(action).CompletesWithin(TimeSpan.FromMilliseconds(250));
+    }
+    
+    [Test]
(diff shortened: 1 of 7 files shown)


AlexandreBossard:
Thank you ! 
Looking at your code, it's not clear to me that the Task will be aborted if it exceed the TimeSpan ?

thomhurst:
> Thank you ! Looking at your code, it's not clear to me that the Task will be aborted if it exceed the TimeSpan ?

You're right, it wouldn't. The assertions library isn't currently set up to abort tasks. I'd need to pass a cancellation token to the delegate used in the Assert.That and you'd need to use that in your task. Is that how FA did it?

AlexandreBossard:
> The assertions library isn't currently set up to abort tasks

Yeah that's what I figured when I tried it myself.

FA don't bother with cancellation token. It creates a new Task with a Timer and use `Task.WhenAny(target, timeoutTask)` for timeouts and manually cancel the operation. The code is [here](https://github.com/fluentassertions/fluentassertions/blob/main/Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs#L297).

thomhurst:
> > The assertions library isn't currently set up to abort tasks
> 
> Yeah that's what I figured when I tried it myself.
> 
> FA don't bother with cancellation token. It creates a new Task with a Timer and use `Task.WhenAny(target, timeoutTask)` for timeouts and manually cancel the operation. The code is [here](https://github.com/fluentassertions/fluentassertions/blob/main/Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs#L297).

From what I can tell, that won't cancel that task, it'll just leave it in the background. But it does solve the problem of not waiting for it which TUnit currently would. I'll have to have a think. Might be a chunky refactor.

PR #1814, opened 2025-02-11
Don't wait for delegates in assertions if they exceed the `CompletesWithin(TimeSpan)` value 

thomhurst:
cc @AlexandreBossard

### DIFF
--- a/TUnit.Assertions/AssertionBuilders/AssertionBuilder.cs
+++ b/TUnit.Assertions/AssertionBuilders/AssertionBuilder.cs
@@ -6,6 +6,7 @@ using TUnit.Assertions.AssertConditions;
 using TUnit.Assertions.AssertConditions.Connectors;
 using TUnit.Assertions.AssertConditions.Interfaces;
 using TUnit.Assertions.Exceptions;
+using TUnit.Assertions.Helpers;
 
 namespace TUnit.Assertions.AssertionBuilders;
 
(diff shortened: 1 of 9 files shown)


TUnit.Assertions/AssertionBuilders/AssertionBuilder.cs
```diff
@@ -175,7 +176,30 @@ internal async Task<AssertionData> ProcessAssertionsAsync()
...
+    private async Task<AssertionData> GetAssertionData()
+    {
+        var minimumWait = _assertions.Select(x => x.WaitFor).Min();
+        
+        if(minimumWait is null)
+        {
+            return await _assertionDataTask;
+        }
+        
+        using var cts = new CancellationTokenSource();
```
campersau, line 189: What is the cts for? Looks like it is not really used. Or should this get cancelled when _assertionDataTask.AsTask() finishes first?

When the cts gets disposed it does not automatically cancel the token:

```c#
var cts = new CancellationTokenSource();
var token = cts.Token;
token.Register(() =>
{
    Console.WriteLine("CANCELLED NOT CALLED");
});
cts.Dispose();

Console.WriteLine(token.IsCancellationRequested); // false
```

```c#
var cts = new CancellationTokenSource();
var token = cts.Token;
token.Register(() =>
{
    Console.WriteLine("CANCELLED");
});
cts.Cancel();
cts.Dispose();

Console.WriteLine(token.IsCancellationRequested); // true
```

thomhurst: Good spot - Thanks!

Addressed here: https://github.com/thomhurst/TUnit/pull/1822

PR #1837, opened 2025-02-13
Resolve Dependencies earlier and run them bottom-up (most nested dependencies run first)

thomhurst:
cc. @Artmos

By executing the most nested dependencies first, we might be able to resolve the hanging by not overloading the framework with lots of suspended tasks.

This may or may not work. Would you be able to try it out when it's released?

You'd need to switch back to trying `[DependsOn]` again as that's what this affects.

### DIFF
--- a/TUnit.Core/DiscoveredTest.cs
+++ b/TUnit.Core/DiscoveredTest.cs
@@ -41,6 +41,8 @@ internal abstract record DiscoveredTest : IComparable<DiscoveredTest>, IComparab
     public abstract IClassConstructor? ClassConstructor { get; }
     
     public IHookExecutor? HookExecutor { get; internal set; }
+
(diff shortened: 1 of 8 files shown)


Artmos:
> This may or may not work. Would you be able to try it out when it's released?

Yeah, I'd be happy to try it out.

thomhurst:
> > This may or may not work. Would you be able to try it out when it's released?
> 
> Yeah, I'd be happy to try it out.

It's in [0.12.13](https://www.nuget.org/packages/TUnit/0.12.13) :)

Artmos:
> It's in [0.12.13](https://www.nuget.org/packages/TUnit/0.12.13) :)

Great, I'll try it out tomorrow :)

Issue #1841, opened 2025-02-13
Docs on Assertion extensions are inconsistent with what is available


AnnaSasDev:
With the one of the latest updates to TUnit, the return value of a BaseAssertion.GetResult(...) was changed to be a Task, so we can use asynchrnous calls, see : discussion.

The Docs were partially updated to reflect this: 
- docs: https://thomhurst.github.io/TUnit/docs/tutorial-assertions/extensibility 
- source: https://github.com/thomhurst/TUnit/blob/main/docs/docs/tutorial-assertions/extensibility.md

The main issue I see is with the code block:
```csharp
public class StringEqualsExpectedValueAssertCondition(string expected, StringComparison stringComparison)
    : ExpectedValueAssertCondition<string, string>(expected)
{
    protected override string GetExpectation()
        => $"to be equal to {Format(expected).TruncateWithEllipsis(100)}";

    protected override AssertionResult GetResult(string? actualValue, string? expectedValue)
    {
        if (actualValue is null)
        {
            return AssertionResult
                .FailIf(
                    () => expectedValue is not null,
                    "it was null");
        }

        return AssertionResult
            .FailIf(
                () => !string.Equals(actualValue, expectedValue, stringComparison),
                $"found {Format(actualValue).TruncateWithEllipsis(100)}");
    }
}
```

This is still based on the old, pre `Task` code. So I tried to fork the codebase and try to create an updated version, but I git another roadblock. 

In the example the `Format(...)` is used, but this isn't easily resolved by my IDE. After some digging I believe this to be `TUnit.Assertions.Helpers.Formatter.Format(...)`, but this is an internal abstract class. This means it can't be used by externally written code, and thus would result in issues when people try to actually copy-paste this example into their own solution.
After some more digging, the `.TruncateithEllipsis(100)` comes from `TUnit.Assertions.Extensions.StringExtensions` which is also an internal class.

My proposed edit would be, if Format and StringExtensions aren't to be made public :
```csharp
public class StringEqualsExpectedValueAssertCondition(string expected, StringComparison stringComparison)
    : ExpectedValueAssertCondition<string, string>(expected)
{
    protected override string GetExpectation()
        => $"to be equal to \"{expected}\"";

    protected override async Task<AssertionResult> GetResult(string? actualValue, string? expectedValue)
    {
        if (actualValue is null)
        {
            return AssertionResult
                .FailIf(
                    expectedValue is not null,
                    "it was null");
        }

        return AssertionResult
[...]

PR #1842, opened 2025-02-13
Make classes for formatting objects in assertion messages public

AnnaSasDev:
see #1841

### DIFF
--- a/TUnit.Assertions/Extensions/StringExtensions.cs
+++ b/TUnit.Assertions/Extensions/StringExtensions.cs
@@ -1,6 +1,6 @@
 ﻿namespace TUnit.Assertions.Extensions;
 
-internal static class StringExtensions
+public static class StringExtensions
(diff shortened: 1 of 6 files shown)


thomhurst:
Thanks @AnnaSasDev . You can make that formatter class public if you like.

Scarlet-Bytes:
> Thanks @AnnaSasDev . You can make that formatter class public if you like.

I've done two things, I've made the classes in `TUnit.Assertions.Helpers` public, and I've made the `TUnit.Assertions.Extensions.StringExtentions` public. This would allow for my updated example to also inlude the original `Format().TruncateWithEllipsis()` you included in the original example. Do I update my revised example to work with the now public Formatter and string Extentions?

thomhurst:
Nah the example is just to show how to set up an assertion. No need to complicate it :smile:

PR #1854, opened 2025-02-14
Add ExcludeFromCodeCoverage attribute to templates

### LINKED ISSUE #1853: [Feature Request] Set [ExcludeFromCodeCoverage] in templates
This is a feature request / suggestion that is operating under the principle that "defaults matter".

By default, code coverage tools calculate code coverage for all loaded assemblies, including the test assemblies themselves. This tends to artificially inflate the code coverage %, as the test code is included in the calculation. This behavior is often surprising to devs.

There's a mild consensus that test code shouldn't be included in % calculations (see https://about.codecov.io/blog/should-i-include-test-files-in-code-coverage-calculations and https://softwareengineering.stackexchange.com/questions/309015/when-does-it-make-sense-to-include-test-code-in-coverage as examples), and among devs that are calculating %, the prevailing method to do that appears to be setting `[ExcludeFromCodeCoverage]` on the test assembly (see https://github.com/search?q=%22assembly%3A+ExcludeFromCodeCoverage%22+language%3AC%23&type=code for some hits).

The specific request is to add `[assembly: ExcludeFromCodeCoverage]` to the templates. In my opinion, this strikes a good balance of leading devs towards the [Pit of Success](https://ricomariani.medium.com/the-pit-of-success-cfefc6cb64c8), while letting them remove it if they disagree. I'm making the same argument in https://github.com/dotnet/sdk/pull/46866, so if that change is accepted / rejected those arguments would apply here as well.

MattKotsenas:
Fixes #1853 

This PR is operating under the principle that "defaults matter".

By default, code coverage tools calculate code coverage for all loaded assemblies, including the test assemblies themselves. This tends to artificially inflate the code coverage %, as the test code is included in the calculation. This behavior is often surprising to devs.

There's a mild consensus that test code shouldn't be included in % calculations (see https://about.codecov.io/blog/should-i-include-test-files-in-code-coverage-calculations and https://softwareengineering.stackexchange.com/questions/309015/when-does-it-make-sense-to-include-test-code-in-coverage as examples), and among devs that are calculating %, the prevailing method to do that appears to be setting `[ExcludeFromCodeCoverage]` on the test assembly (see https://github.com/search?q=%22assembly%3A+ExcludeFromCodeCoverage%22+language%3AC%23&type=code for some hits).

For this change I added the assembly attribute to the templates. This strikes a good balance of leading devs towards the [Pit of Success](https://ricomariani.medium.com/the-pit-of-success-cfefc6cb64c8), while letting them remove it if they disagree.


MattKotsenas:
Per the PR template, there don't appear to be any tests for the templates. If you want / need them let me know; the template tool provides https://github.com/dotnet/templating/blob/main/docs/authoring-tools/Templates-Testing-Tooling.md.

TUnit.Templates/content/TUnit.AspNet/TestProject/Tests.cs
```diff
@@ -1,5 +1,7 @@
 ﻿namespace TestProject;
 
+[assembly: System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage]
```
thomhurst: Could we split this out into a GlobalSetup file again just for separation of concerns?

thomhurst:
> Per the PR template, there don't appear to be any tests for the templates. If you want / need them let me know; the template tool provides https://github.com/dotnet/templating/blob/main/docs/authoring-tools/Templates-Testing-Tooling.md.

If you've used that before and it's easy to implement I won't say no for tests to be added :smile:

thomhurst:
There's also a Playwright template

MattKotsenas:
> > Per the PR template, there don't appear to be any tests for the templates. If you want / need them let me know; the template tool provides https://github.com/dotnet/templating/blob/main/docs/authoring-tools/Templates-Testing-Tooling.md.
> 
> If you've used that before and it's easy to implement I won't say no for tests to be added 😄

#1863 

Once that looks good to you and is merged I'll rebase this PR on top of that one

TUnit.Templates/content/TUnit.AspNet/TestProject/Tests.cs
MattKotsenas: Fixed!

MattKotsenas:
> There's also a Playwright template

I didn't add it initially because Playwright is more an integration test tool, so I would be surprised if code coverage was useful there at all?

That said, I think it's reasonable to keep the templates in sync with each other, so added to Playwright as well :)

PR #1859, opened 2025-02-15
Use ValueTask instead of Task in Assertion logic to reduce allocations

AnnaSasDev:
A priliminary search of which `Task` methods can be replaced by `ValueTask`
#1850


TUnit.Assertions/AssertConditions/AssertionResult.cs
Scarlet-Bytes, line 1: I'm not really happy with this, but there is an "issue" that NetStandard 2.0 doesn't have ValueTask.FromResult().
I believe this to be the best solution, although a bit ugly

thomhurst: I'd just simplify to new valuetask to avoid the compilation if statement. I think from result just calls that anyway

Scarlet-Bytes: Reverted to just `new(result)`

thomhurst:
All looks good! Could you update the doc examples too? 😇

thomhurst:
Thanks @AnnaSasDev !

PR #1863, opened 2025-02-15
Add tests for dotnet new templates

MattKotsenas:
Add tests for the `dotnet new` templates. The tests use `VerificationEngine` as provided by https://github.com/dotnet/templating/blob/main/docs/authoring-tools/Templates-Testing-Tooling.md.

Each test:
1. Installs the template to a temporary location
2. Instantiates the templates
3. Compares the output of the template with "verified" files (a.k.a. snapshot testing or golden master testing)
4. Cleans up

Each test therefore verifies:
- The template can be installed by users
- No missing / extra files
- Variable and name substitution occurs as expected
- Any template parameters (we don't have any) are respected

In the process of making the tests I found some trailing whitespace in the templates, so I fixed that as well.


thomhurst:
Nice one!

PR #1865, opened 2025-02-15
Add new TUnit.Template.Tests to the build pipeline

thomhurst:


thomhurst:
@MattKotsenas this repo automatically bumps the TUnit version after each release, meaning the templates always have up to date versions.
But the snapshots for the csproj files are expecting a specific version.

Is it possible to regex the version or anything?

MattKotsenas:
Yeah, you can do a "scrubber" like defined here: https://github.com/dotnet/dotnet/blob/refs/heads/main/src/templating/docs/authoring-tools/Templates-Testing-Tooling.md#configuration

To replace the version with some constant so the snapshots don't change every time.

I probably can't push a commit to your branch, but let me know if you want some kind of snippet or help.

MattKotsenas:
Here's an example of kind of the simplest thing that works: https://github.com/mattkotsenas/TUnit/commit/c8cac868456ad47f0e1f149fab67bf04a97b148b

```csharp
protected TemplateVerifierOptions Options => new TemplateVerifierOptions(TemplateShortName)
{
    TemplatePath = Path.Combine(TestContext.CurrentContext.TestDirectory, "content", TemplateShortName),
}.WithCustomScrubbers(ScrubbersDefinition.Empty.AddScrubber(sb =>
{
    string original = sb.ToString();
    Match match = Regex.Match(original, """<PackageReference Include="TUnit[^"]*" Version="([^"]*)" />""");

    if (match.Success)
    {
        string line = match.Groups[0].Value.Replace(match.Groups[1].Value, "{VERSION}");
        sb.Replace(match.Value, line);
    }

}, "csproj"));
```

If we need to make it more robust or faster, let me know.

thomhurst:
> Here's an example of kind of the simplest thing that works: [MattKotsenas@c8cac86](https://github.com/MattKotsenas/TUnit/commit/c8cac868456ad47f0e1f149fab67bf04a97b148b)
> 
> ```cs
> protected TemplateVerifierOptions Options => new TemplateVerifierOptions(TemplateShortName)
> {
>     TemplatePath = Path.Combine(TestContext.CurrentContext.TestDirectory, "content", TemplateShortName),
> }.WithCustomScrubbers(ScrubbersDefinition.Empty.AddScrubber(sb =>
> {
>     string original = sb.ToString();
>     Match match = Regex.Match(original, """<PackageReference Include="TUnit[^"]*" Version="([^"]*)" />""");
> 
>     if (match.Success)
>     {
>         string line = match.Groups[0].Value.Replace(match.Groups[1].Value, "{VERSION}");
>         sb.Replace(match.Value, line);
>     }
> 
> }, "csproj"));
> ```
> 
> If we need to make it more robust or faster, let me know.

Nice, thanks!

PR #1980, opened 2025-03-02
Document how to create and use global test identifiers

dystopiandev:
Hi @thomhurst 

I was able to put together a guide for my use case: Instrumenting TUnit to assign unique IDs to tests run in parallel.

Please indicate if I need to make any adjustments. Thanks

### DIFF
--- a/docs/docs/examples/instrumenting_global_test_ids.md
+++ b/docs/docs/examples/instrumenting_global_test_ids.md
@@ -1,0 +1,91 @@
+---
+sidebar_position: 3
+---
+
+# Instrumenting: Global Test IDs
+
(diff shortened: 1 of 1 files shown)


thomhurst:
Do you need an integer id? There's a unique string identifier out of the box. `testContext.TestDetails.TestId`

dystopiandev:
> Do you need an integer id? There's a unique string identifier out of the box. `testContext.TestDetails.TestId`

I do need an integer ID to work with Redis' database *numbers*, and that's the basis of the contrived example, hence the use case.

I should've indicated that this is an alternative to the inbuilt string ID?

thomhurst:
Thanks for the contribution 😄 https://thomhurst.github.io/TUnit/docs/examples/instrumenting_global_test_ids

PR #1998, opened 2025-03-04
Short circuit 'Or' asserts on the first success

### LINKED ISSUE #1997: Bug: Chaining assertions with `.Or` doesn't short-circuit.
Example:
```csharp
internal class FooTests
{
    [Test]
    public async Task Basic()
    {
        var foo = await Assert.That(ThrowException).ThrowsException();
        await Assert.That(foo)
                    .IsNotAssignableTo<ArgumentOutOfRangeException>()
                    .Or
                    .Satisfies(x => (ArgumentOutOfRangeException)x,
                               x => x.HasMember(y => y.ActualValue).EqualTo("foo"));
    }
    
    private void ThrowException() => throw new InvalidOperationException("foo");
}
```
This throws an `InvalidCastException` even though the assertion should be satisfied - it should never reach the cast to `ArgumentOutOfRangeException`.

andrewimcclement:
This resolves #1997 , though the implementation and test are extremely basic, merely enough to demonstrate the issue and a potential fix. Please let me know where you would like me to go with this PR.

Note the breaking change in removing `AssertionResult.Or` in favour of `OrAsync`.

### DIFF
--- a/TUnit.Assertions/AssertConditions/AssertionResult.cs
+++ b/TUnit.Assertions/AssertConditions/AssertionResult.cs
@@ -56,19 +56,6 @@ public class AssertionResult
             : Fail(Message + " and " + other.Message);
     }
 
-    // TODO: Should this be removed/obsoleted?
-    public AssertionResult Or(AssertionResult other)
-    {
-        if (!IsPassed && !other.IsPassed)
-        {
-            return Message == other.Message 
-                ? Fail(Message) 
-                : Fail(Message + " and " + other.Message);
-        }
-
-        return Passed;
-    }
-
     public async ValueTask<AssertionResult> OrAsync(Func<ValueTask<AssertionResult>> otherResult)
     {
         if (IsPassed)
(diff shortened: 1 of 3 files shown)


TUnit.Assertions/AssertConditions/AssertionResult.cs
```diff
@@ -56,6 +56,7 @@ public AssertionResult And(AssertionResult other)
             : Fail(Message + " and " + other.Message);
     }
 
+    // TODO: Should this be removed/obsoleted?
     public AssertionResult Or(AssertionResult other)
     {
```
thomhurst: If this isn't used by the core code anymore, remove it. We can add a breaking label to the pr which will appear in the release notes

andrewimcclement: Removed it.

thomhurst:
Thanks for the fix @andrewimcclement ! 😄

Issue #1972, opened 2025-03-01
Equivalency with missing members in expectation


mrnustik:
Hi, I like to use partial equivalency for assertions in my tests (as seen in the sample below). Unfortunately TUnit `IsEquivalentTo` assertion requires `actual` to have the same shape as `expected`. I know there is `IgnoringMembers(nameof(Member))`, but that becomes tedious with large objects. 

Would you be open to add some sort option, that would ignore all members that are present in `actual`, but missing from `expected`? 

I have already inspected the `TUnit` codebase and would be willing to implement it myself, I just want to check whether this is something you would like to be added.

```csharp
public class Test
{
    [Test]
    public async Task SampleTest()
    {
       var something = new Something("Test", 42);

        // Fails on missing IntValue
        await Assert.That(something)
              .IsEquivalentTo(new
              {
                  StringValue = "Test",
              });
    }
    
    public record Something(
        string StringValue,
        int IntValue);
}
```

thomhurst:
Could we add an overload with an enum specifically stating it's a partial equality? I just want to make the code clear on what it's doing.

If so, happy for you to take this on!

PR #2021, opened 2025-03-07
feat: add support for partial equivalency

mrnustik:
### Please check the following before creating a Pull Request

- If this is a new feature or piece of functionality, have you started a discussion and gotten agreement on it?
    - Discussed in #1972 
- If it fixes a bug or problem, is there an issue to track it? If not, create one first and link it please so there's clear visibility.
- Did you write tests to ensure you code works properly?
    - Yes 

In case you have any issues with the public API feel free to comment. I also thought about adding it as a method to `EquivalentToAssertionBuilderWrapper` similarly to how `IgnoringMember` is used and am open to adding to it there. I added it as an overload because I did not have any strong opinion. 🙂

### DIFF
--- a/TUnit.Assertions.UnitTests/EquivalentAssertionTests.cs
+++ b/TUnit.Assertions.UnitTests/EquivalentAssertionTests.cs
@@ -376,6 +376,162 @@ public class EquivalentAssertionTests
         await TUnitAssert.That(object1).IsEquivalentTo(object2);
     }
     
+    [Test]
+    public void Objects_With_Partial_Properties_Match_With_Full_Equivalency_Are_Not_Equivalent()
+    {
+        var object1 = new MyClass
+        {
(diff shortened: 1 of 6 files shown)


thomhurst:
Actually adding it as a method on the builder would be easier to maintain going forwards I think. Otherwise there's going to be a lot of combination of overloads to manage

mrnustik:
Replaced it with single method on the builder. What do you think?

Also can you please give me a hint how to suppress the codacy warnings on the classes used for fields equivalency testing?

thomhurst:
Perfect!

Don't worry about codacy, it won't block merging.

mrnustik:
FYI: Just pushed the update to the non-nullable fields in test to prevent build warnings.

thomhurst:
Thanks for the contribution!

PR #2088, opened 2025-03-13
Support automatic generation of values for`[MatrixDataSource]` when using bool? and TEnum?

andrewimcclement:
This brings TUnit inline with NUnit `[Values]`/`[Theory]` value generation.


thomhurst:
Thanks @andrewimcclement !

andrewimcclement:
Hmm. Looks like I broke something :(

andrewimcclement:
Thanks for fixing it up - didn't have time to keep working on it.

samtrion:
@thomhurst @andrewimcclement - This PR seems to break the creation of an `enum` (not nullable) based matrix. Only the value for the enum `default` is created.

After the update, instead of >2800 tests, I only get 50 tests that are created.

thomhurst:
@samtrion oh really? Can you give a repro and I'll take a look?

samtrion:
> @samtrion oh really? Can you give a repro and I'll take a look?

I will try to provide you with one for reproduction in the course of the evening.

PR #2156, opened 2025-03-25
STAThreadExecutorAttribute, CultureAttribute and abstract DedicatedThreadExecutor

thomhurst:


TUnit.Core/Executors/DedicatedThreadExecutor.cs
```diff
@@ -0,0 +1,52 @@
+﻿namespace TUnit.Core;
+
+public class DedicatedThreadExecutor : GenericAbstractExecutor
+{
+    protected sealed override async ValueTask ExecuteAsync(Func<ValueTask> action)
+    {
+        var tcs = new TaskCompletionSource<object?>();
+
+        var thread = new Thread(async void () =>
```
hartmair: I think you should stay with non-async `valueTask.AsTask.GetResult()` see https://stackoverflow.com/a/78936077/1220972

thomhurst: Thanks! Good shout

PR #2164, opened 2025-03-27
`await Assert.That(array).Contains(predicate)` returns the found item

### LINKED ISSUE #2142: Assertions Contains should return the item
Would be handy to have `Assert.That(collection).Contains(predicate)` return the found item, i.e.

```cs
public class A
{
    public int Id { get; set; }
    public required string Name { get; set; }
}

[Test]
public async Task Test1()
{
    // Arrange
    A[] collection = [new A() { Id = 1, Name = "First" }];

    // Act
    collection[0].Name = "1st";

    // Assert
    A actual = await Assert.That(collection).Contains(x => x.Id == 1);
    await Assert.That(actual.Name).IsEqualTo("1st");
}
```

Maybe with a separate fluent method, which would allow for chaining too:

```cs
Assert.That(collection).Contains(x => x.Id == 1).Which().HasMember(x => x.Name).EqualTo("1st")
```

Alternatively, one could add another method similar to `HasSingleItem` and leave `Contains` as it is.

```cs
A actual = await Assert.That(collection).HasItem(x => x.Id == 1);
```

thomhurst:
Fixes #2142


TUnit.Assertions/Assertions/Generics/DoesExtensions_Generic.cs
```diff
@@ -16,10 +16,10 @@ public static InvokableValueAssertionBuilder<TActual> Contains<TActual, TInner>(
             , [doNotPopulateThisValue]);
     }
     
-    public static InvokableValueAssertionBuilder<IEnumerable<TInner>> Contains<TInner>(this IValueSource<IEnumerable<TInner>> valueSource, Func<TInner, bool> matcher, [CallerArgumentExpression(nameof(matcher))] string doNotPopulateThisValue = null)
+    public static MappableResultAssertionBuilder<IEnumerable<TInner>, TInner> Contains<TInner>(this IValueSource<IEnumerable<TInner>> valueSource, Func<TInner, bool> matcher, [CallerArgumentExpression(nameof(matcher))] string doNotPopulateThisValue = null)
     {
-        return valueSource.RegisterAssertion(new EnumerableContainsExpectedFuncAssertCondition<IEnumerable<TInner>, TInner>(matcher, doNotPopulateThisValue)
-            , [doNotPopulateThisValue]);
+        return new MappableResultAssertionBuilder<IEnumerable<TInner>, TInner>(valueSource.RegisterAssertion(new EnumerableContainsExpectedFuncAssertCondition<IEnumerable<TInner>, TInner>(matcher, doNotPopulateThisValue)
```
hartmair, line 21: This will enumerate the `IEnumerable` twice and also evaluate `matcher` for every item twice, right?

thomhurst: If awaited directly yeah. Didn't think about that. Maybe I can opt to pull it from the assert condition instead.

PR #2403, opened 2025-05-19
New Advanced Tutorial F# Interactive

licon4812:
I created an F# Interactive tutorial for the documentation.

This will showcase how to write and execute TUnit tests from the built-in REPL for F# 

Go [ here ](https://learn.microsoft.com/en-us/dotnet/fsharp/tools/fsharp-interactive/)to read what F# interactive is.

## Local build of docusaurus
image


thomhurst:
Thanks @licon4812 !

PR #2429, opened 2025-05-21
F# ASP.NET templates

licon4812:
As F# is an officially supported Language in the ASP.NET Core ecosystem. I have added a template to support this.

Thankfully, it was a lot easier to do than the Basic template made in a prior commit


thomhurst:
Nice one :)

licon4812:
@thomhurst  I timed that one final commit poorly 🤣

PR #2427, opened 2025-05-21
Added more f# tests

licon4812:
This PR will add more Tests to the Example project. This is to ensure all Assertions and Attributes work correctly in F#

PR #2431, opened 2025-05-21
Updated TUnit.Assertions.FSharp.Operations check function to support assertions with Throw Type

licon4812:
Added support for the Throw assertion type in the check function. 

To keep syntax and utilisation simple. I had to use pattern matching inside the check.  Good old fashion Functional paradimes.

Fortunately we can add to it later on. I would have liked to have made reusable code, but I think it is best we resolve the types in this way to ensure of type safety, and avoid surprise F# type conversions later on


licon4812:
Tests for this are in the draft PR #2427

I tested them out, and they behaved as intended

PR #2437, opened 2025-05-22
Don't overwrite LangVersion if set and check raise diagnostic error if level not high enough

### LINKED ISSUE #2426: TUnit.Engine sets LangVersion=latest, overridding my repo policy
I set LangVersion to 13 in my repo for all my projects. I prefer explicit versions set rather than `latest` so that the compilation output is the same regardless of SDK version or VS version installed on the machine.

tunit.engine.props overrides this property with `latest`, and doesn't give me an opportunity to disable this.

Perhaps you can add a property I can set to indicate I want project properties like this to be left alone.
Or maybe don't set it if it's already set.
Or maybe just produce a build error if it isn't set to a high enough value, so that users understand and can manage this property within their project file, as they are used to.

thomhurst:
Fixes #2426

### DIFF
--- a/TUnit.Core.SourceGenerator/CodeGenerators/LanguageVersionCheckGenerator.cs
+++ b/TUnit.Core.SourceGenerator/CodeGenerators/LanguageVersionCheckGenerator.cs
@@ -1,0 +1,42 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+
+namespace TUnit.Core.SourceGenerator;
+
+[Generator]
+public class LanguageVersionCheckGenerator : IIncrementalGenerator
+{
+    public void Initialize(IncrementalGeneratorInitializationContext context)
+    {
+        var settings = context.CompilationProvider
+            .Select((c, _)  => 
+            {
+                LanguageVersion? csharpVersion = c is CSharpCompilation comp
+                    ? comp.LanguageVersion
+                    : null;
+                
+                return csharpVersion;
+            });
+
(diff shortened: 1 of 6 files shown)


TUnit.Core.SourceGenerator/CodeGenerators/LanguageVersionCheckGenerator.cs
```diff
@@ -0,0 +1,42 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+
+namespace TUnit.Core.SourceGenerator;
+
+[Generator]
+public class LanguageVersionCheckGenerator : IIncrementalGenerator
+{
+    public void Initialize(IncrementalGeneratorInitializationContext context)
```
Youssef1313, line 9: Looks to me like this shouldn't be a source generator at all. It should instead be a diagnostic analyzer.

TUnit/TUnit.props
```diff
@@ -7,7 +7,6 @@
         <TestingPlatformDotnetTestSupport>true</TestingPlatformDotnetTestSupport>
         <TestingPlatformShowTestsFailure>true</TestingPlatformShowTestsFailure>
         <TestingPlatformCaptureOutput>false</TestingPlatformCaptureOutput>
-        <LangVersion>latest</LangVersion>
         <OutputType>Exe</OutputType>
```
Youssef1313, line 10: Unrelated to your PR. But setting OutputType in NuGet props/targets is very bad and should NOT be done. It should be on the user to set this.

PR #2600, opened 2025-06-23
A bit of a rewrite

thomhurst:
Reworked Source Generators
Async Data Sources
Nested Data Source Injection
More unified approach building tests via source generation or reflection
Reworked Object Lifetime management


agray:
The "Big Beautiful Rewrite"?

thomhurst:
@copilot Can you list all the breaking changes users might experience from the Public API?

PR #2821, opened 2025-08-07
IAsyncEnumerable streaming throughout test lifecycle - Discovery > Build > Execute

thomhurst:


TUnit.Core/DataSources/DataSourceProcessor.cs
```diff
@@ -269,7 +276,10 @@ public static class DataSourceProcessor
             
             if (items.Count > 0)
             {
-                yield return items.ToArray();
+                // Optimize: Use pre-sized array instead of ToArray()
```
viceroypenguin, line 279: fyi, `items.ToArray()` will pre-size automatically, since `items` is an `ICollection<>`

[source](https://source.dot.net/#System.Linq/System/Linq/ToCollection.cs,45)

PR #2827, opened 2025-08-07
Add support for ExpandValueTuple in SymbolDisplayFormat for Roslyn 4.7+

thomhurst:
Fixes #2799


thomhurst:
That looks like it worked a treat. Thanks @Brokolis !

```
        if (value is global::System.ValueTuple<int, int> typedValue)
        {
            return (global::TUnit.TestProject.Bugs._2798.Foo)typedValue;
        }
```

PR #2843, opened 2025-08-08
Use MSBuild version comparison functions instead of numerical comparisons

thomhurst:
@Brokolis fingers crossed!


Brokolis:
I can confirm that everything works in 0.55.0. Thanks for the fixes!

PR #2932, opened 2025-08-17
Fix test discovery intermittent hanging issues in reflection mode

thomhurst:
This commit addresses multiple critical deadlock scenarios that could cause test discovery to hang intermittently:

1. Missing ConfigureAwait(false) calls (40+ locations fixed)
   - Added to all await calls in TestDiscoveryService, HookOrchestrator, and TestBuilderPipeline
   - Prevents SynchronizationContext capture that causes deadlocks in UI/ASP.NET environments

2. Thread pool exhaustion from unbounded parallelism
   - Added SemaphoreSlim throttling to limit concurrent assembly discovery to ProcessorCount * 2
   - Prevents thread pool starvation when discovering tests in solutions with many assemblies

3. Static lock bottleneck causing serialization
   - Replaced HashSet<Assembly> with ConcurrentDictionary<Assembly, bool>
   - Replaced List<TestMetadata> with ConcurrentBag<TestMetadata>
   - Eliminated global lock that was serializing all parallel discovery operations

### DIFF
--- a/TUnit.Engine/Building/TestBuilderPipeline.cs
+++ b/TUnit.Engine/Building/TestBuilderPipeline.cs
@@ -30,9 +30,9 @@ internal sealed class TestBuilderPipeline
     public async Task<IEnumerable<AbstractExecutableTest>> BuildTestsAsync(string testSessionId, HashSet<Type>? filterTypes)
     {
         var dataCollector = _dataCollectorFactory(filterTypes);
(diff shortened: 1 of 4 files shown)


thomhurst:
@claude Fix the failing build

thomhurst:
@claude Fix the failing tests

thomhurst:
@claude update the conflicting depends on tests so that the assertion works with either test being at the beginning of the chain, as parallelism makes it non deterministic which will run first.

PR #3023, opened 2025-08-31
Add a simple retry to the GithubReporter to prevent crashing on file locks

### LINKED ISSUE #3022: Github Actions summary reporter doesn't handle locked files gracefully
Hi! Thanks for this awesome testing framework.

I was running some TUnit tests in Github Actions and noticed that sometimes the logger would fail due to the [step summary path being locked](https://github.com/baronfel/tuf-dotnet/actions/runs/17337965908/job/49227307751) presumably because another test process was already writing to it in my multi-project, multi-TFM setup). I think a simple delay-retry-with-jitter could help patch over this exception, because other than reporting the tests pass!

baronfel:
Fixes #3022 with a simple delay-based retry loop.

Sorry if this is a bit presumptuous - it was a quick thing to add and so I figured I'd send it up.

### DIFF
--- a/TUnit.Engine/Reporters/GitHubReporter.cs
+++ b/TUnit.Engine/Reporters/GitHubReporter.cs
@@ -49,10 +49,16 @@ public class GitHubReporter(IExtension extension) : IDataConsumer, ITestHostAppl
     public string Description => extension.Description;
 
     private readonly ConcurrentDictionary<string, List<TestNodeUpdateMessage>> _updates = [];
+    private static readonly Random _jitterRandom =
+#if NET
+                Random.Shared;
+#else
+                new();
+#endif
 
     public Task ConsumeAsync(IDataProducer dataProducer, IData value, CancellationToken cancellationToken)
     {
-        var testNodeUpdateMessage = (TestNodeUpdateMessage) value;
+        var testNodeUpdateMessage = (TestNodeUpdateMessage)value;
 
         _updates.GetOrAdd(testNodeUpdateMessage.TestNode.Uid.Value, []).Add(testNodeUpdateMessage);
 
@@ -173,7 +179,7 @@ public class GitHubReporter(IExtension extension) : IDataConsumer, ITestHostAppl
         return WriteFile(stringBuilder.ToString());
     }
 
-    private Task WriteFile(string contents)
+    private async Task WriteFile(string contents)
     {
         var fileInfo = new FileInfo(_outputSummaryFilePath);
         var currentFileSize = fileInfo.Exists ? fileInfo.Length : 0;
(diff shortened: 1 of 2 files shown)


TUnit.Engine/Reporters/GitHubReporter.cs
```diff
@@ -183,14 +189,49 @@ private Task WriteFile(string contents)
...
-        return File.AppendAllTextAsync(_outputSummaryFilePath, contents, Encoding.UTF8);
+        var maxRetries = int.TryParse(EnvironmentVariableCache.Get("TUNIT_GITHUB_ACTIONS_REPORTER_RETRY_COUNT"), out var retryCount) ? retryCount : 5;
+        var attempts = 0;
+
+        while (attempts < maxRetries)
+        {
+            attempts++;
+            try
+            {
+                await Append(_outputSummaryFilePath, contents);
```
thomhurst, line 203: Do we not want a return statement after it's successful to avoid it appending it 5 times?

TUnit.Engine/Reporters/GitHubReporter.cs
```diff
@@ -183,14 +189,49 @@ private Task WriteFile(string contents)
...
+        }
+    }
+
+    private static Task Sleep(int delayMs)
+    {
+#if NET 
+        return Task.Delay(delayMs);
 #else
-        File.AppendAllText(_outputSummaryFilePath, contents, Encoding.UTF8);
+        System.Threading.Thread.Sleep(delayMs);
```
thomhurst, line 224: Why not Task Delay for NET Framework too?

TUnit.Engine/Reporters/GitHubReporter.cs
```diff
@@ -183,14 +189,49 @@ private Task WriteFile(string contents)
...
+        return Task.CompletedTask;
+#endif
+    }
 
+    private static Task Append(string filePath, string contents)
+    {
+#if NET
+        return File.AppendAllTextAsync(filePath, contents, Encoding.UTF8);
```
thomhurst, line 232: You should be able to use `FilePolyfill` - Which will work for both .NET and .NET Framework

TUnit.Engine/Reporters/GitHubReporter.cs
baronfel: yeah we do, just going too fast in a github codespace :-/

TUnit.Engine/Reporters/GitHubReporter.cs
baronfel: mostly because copilot generated most of my PR

TUnit.Engine/Reporters/GitHubReporter.cs
baronfel: TUnit.Engine doesn't have a dependency on the `polyfill` library yet - it's in TUnit.Core already, so I could bring it in if you're ok with that.

thomhurst: Ah don't worry for now then. I think it might cause conflicts

PR #3081, opened 2025-09-09
Detect injected properties in base classes

### LINKED ISSUE #3072: Regression: Nested ClassDataSource property not injected in derived data source class after upgrade to 0.57.24
### Summary
After upgrading from TUnit 0.25.21 to 0.57.24, I noticed that property injection doesn’t seem to work the same way anymore for data source properties defined in a base class.

In my setup, the base class has a property decorated with [ClassDataSource<T>], and a derived class (which inherits that property) is itself used as a data source for a test. In 0.25.21 this worked fine, but in 0.57.24 the inherited property isn’t injected—it stays null during initialization.

### Code to Reproduce
```csharp
public abstract class BaseClass
{
    [ClassDataSource<DataClass>(Shared = SharedType.PerTestSession)]
    public required DataClass TestData { get; init; }
}

public class TestFactory : BaseClass, IAsyncInitializer
{
    public Task InitializeAsync()
    {
        var test = TestData.Testroperty; // TestData is null here in 0.57.24
        return Task.CompletedTask;
    }
}

public class Tests : IAsyncInitializer
{
    [ClassDataSource<TestFactory>(Shared = SharedType.PerTestSession)]
    public required TestFactory TestDataFactory { get; init; }
}
```

### Expected Behavior
- On TUnit 0.25.21, this pattern works: `TestData` is properly injected and accessible in `TestFactory.InitializeAsync()`.

### Actual Behavior
- On TUnit 0.57.24, `TestData` is null in `TestFactory.InitializeAsync()` and test execution fails with:
  > Failed to resolve data source for property 'TestDataFactory': Object reference not set to an instance of an object.

**Questions:**
- Is this a regression or expected behavior change?
- Is there a new pattern required for this scenario in TUnit 0.57.24?

thomhurst:
Fixes #3072

### DIFF
--- a/TUnit.Core.SourceGenerator.Tests/AbstractTests.Concrete1.verified.txt
+++ b/TUnit.Core.SourceGenerator.Tests/AbstractTests.Concrete1.verified.txt
@@ -34,7 +34,7 @@ internal sealed class ConcreteClass1_AssertClassName_TestSource_GUID : global::T
             },
             InheritanceDepth = 1,
             FilePath = @"",
-            LineNumber = 3,
+            LineNumber = 5,
             MethodMetadata = new global::TUnit.Core.MethodMetadata
             {
                 Type = typeof(global::TUnit.TestProject.AbstractTests.AbstractBaseClass),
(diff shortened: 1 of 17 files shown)


aradalvand:
@thomhurst I think this caused a bug: #3084
Probably [this part](https://github.com/thomhurst/TUnit/pull/3081/files#diff-b7e15e31b0851263209bc6284ecbc8eb494e600fa6b6569d78beb0e3f0970983), though I didn't inspect the PR too meticulously.

PR #3093, opened 2025-09-11
Restructure the test runner for better code coherency

### LINKED ISSUE #3044: Before(Class) and After(Class) appear to be parallel when in different ParallelGroups
I'm not sure if I'm not doing something correctly but let's see.
I have several test classes which overlap in database calls so I have different ParallelGroup's at the top of the classes. However, I have noticed that the After(Class) method's of classes will sometimes run before the Before(Class) method of other classes (in my case, this resulted in tables getting cleaned up after another class set them up). Is there a way to make sure that the Before(Class) of a class only runs after the After(Class) of another is complete when they're in different ParallelGroups? I'm not sure if DependsOn works since the Before and After methods aren't tests (although it would be annoying to try to keep track of which class goes after another class).

Edit: Sorry, I previously had Before(Test) when I meant Before(Class) and such.

thomhurst:
Fixes #3044

### DIFF
--- a/TUnit.Engine/Services/TestExecution/TestCoordinator.cs
+++ b/TUnit.Engine/Services/TestExecution/TestCoordinator.cs
@@ -1,5 +1,6 @@
 using System.Linq;
 using TUnit.Core;
+using TUnit.Core.Exceptions;
 using TUnit.Core.Logging;
 using TUnit.Engine.Interfaces;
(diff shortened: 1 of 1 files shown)


thomhurst:
@claude Fix this pr. There seems to be an issue with skipped tests

Issue #3137, opened 2025-09-20
Async lambdas passed to Assert.That(...) are evaluated too early


m-gasser:
Lambdas passed to Assert.That(...) are intended to be evaluated during AssertionBuilder.GetAssertionData(). However async lambdas are already started when AsAssertionData() is called, which happens already during builder execution.
The underlying reason is, that when a Func<Task> is invoked to get the Task, the Task may immediately be scheduled on the thread pool.

It might be possible to wrap a Func<Task> into a ValueTask in a way that guarantees lazy evaluation, which would fix the bug. However making the laziness requirement more explicit is probably better. This way it is not only less error-prone but it should also make the code more clear.

m-gasser:
My draft PR fixes the issue, but ConvertedDelegateSource and ConvertedValueSource are currently broken. Before I invest time in implementing those correctly, I would like to get some feedback on the approach I have chosen and the current changes.

PR #3138, opened 2025-09-20
Fix async lambdas passed to Assert.That(...) are evaluated too early

m-gasser:
Attempt at fixing https://github.com/thomhurst/TUnit/issues/3137

### DIFF
--- a/TUnit.PublicAPI/Tests.Assertions_Library_Has_No_API_Changes.Net4_7.verified.txt
+++ b/TUnit.PublicAPI/Tests.Assertions_Library_Has_No_API_Changes.Net4_7.verified.txt
@@ -2606,7 +2606,6 @@ namespace .Extensions
     {
         public static .<.> IsCanceled(this ..IValueSource<.> valueSource) { }
         public static .<.> IsCompleted(this ..IValueSource<.> valueSource) { }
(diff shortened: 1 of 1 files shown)


TestProject.props
```diff
@@ -9,7 +9,7 @@
         Condition="'$(MSBuildProjectExtension)' == '.fsproj'" />
 
     <PropertyGroup>
-        <TargetFrameworks>net472;net8.0;net9.0</TargetFrameworks>
+        <TargetFrameworks>net8.0;net9.0</TargetFrameworks>
```
m-gasser: Revert this (this was temporarily removed to avoid installing Mono runtime on linux)

thomhurst:
This looks good to me - I'm happy for it to be lazily evaluated 😄

TUnit.Assertions.Tests/Assertions/Delegates/Throws.ExactlyTests.cs
```diff
@@ -121,7 +121,9 @@ await Assert.That(action)
 
             await Assert.That(assertionException).HasMessageStartingWith("""
                                                                          Expected action to throw exactly an Exception
-                                                                         
+                                                                          and message to be equal to "Foo bar message"
```
m-gasser: The behavior here seems to have changed. Not sure if that is bad or an improvement.

PR #3164, opened 2025-09-21
feat: add parsing assertions for various types with format provider support

### LINKED ISSUE #2447: Assertions for IParsable<T>,  ISpanParsable<T> and IUtf8SpanParsable<T>
Hi!

Just as an FYI: Before I switched to TUnit I had already created an issue a while back at fluentassertions/fluentassertions#2632.

The idea is pretty much to verify, that a given string is in such a format, that an instance of type `TTarget` can be created out of it.

So yes, `TTarget` is expected to have a ready implementation of the Parsing Methods.

In my previous issue I hadn't payed attention, but there are also interfaces dedicated to `ReadOnlySpan<char>` and `ReadOnlySpan<byte>`, so I'd include those in my idea here.

Since there are always two methods `Parse` and `TryParse` there was a discussion on which one was to be used, with the former being settled on.

In my mind TUnit would then end up having new APIs like so:

```csharp

Assert.That(...)
    .IsParsableInto<TTarget>()
    .WithFormatProvider(IFormatProvider? provider);
Assert.That(...)
    .IsNotParsableInto<TTarget>()
    .WithFormatProvider(IFormatProvider? provider);

```

thomhurst:
Fixes #2447

### DIFF
--- a/TUnit.Assertions.Tests/ParseAssertionTests.cs
+++ b/TUnit.Assertions.Tests/ParseAssertionTests.cs
@@ -1,6 +1,7 @@
 using System.Globalization;
+using TUnit.Assertions.Assertions.Strings;
 
-namespace TUnit.TestProject;
+namespace TUnit.Assertions.Tests;
 
 public class ParseAssertionTests
 {


OwnageIsMagic:
Code for `IParsable` is not really needed cause you calling `GetMthod` on target type so it is resolved to the same method as non interface code path. Also `null` may be valid value for user defined `TryParse` method, so this shortcut is bogus.
You left empty block checking for 
`ISpanParsable`.
`ISpanParsable` inherits from `IParsable`, so there is no point checking if not using spans.

PR #3165, opened 2025-09-21
fix: add WithMessageNotContaining assertion for exception messages

### LINKED ISSUE #3104: add WithMessageNotContaining() to ThrowsException<>()
The `WithMessageContaining()` extension is helpful. But sometimes, you actually want to make sure the exception message does not contain a certain phrase. 

I can add a PR if desired.

thomhurst:
Fixes #3104

### DIFF
--- a/TUnit.Assertions/Assertions/Throws/ThrowsException.cs
+++ b/TUnit.Assertions/Assertions/Throws/ThrowsException.cs
@@ -50,6 +50,20 @@ public class ThrowsException<TActual, TException> where TException : Exception
         return this;
     }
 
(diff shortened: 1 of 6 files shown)


TUnit.Assertions/Assertions/Throws/ThrowsException.cs
```diff
@@ -50,6 +50,20 @@ public ThrowsException<TActual, TException> WithMessageContaining(string expecte
         return this;
     }
 
+    public ThrowsException<TActual, TException> WithMessageNotContaining(string expected, [CallerArgumentExpression(nameof(expected))] string? doNotPopulateThisValue = null)
```
vslee, line 53: nit: perhaps the parameter should be named ```notExpected``` instead of ```expected```?

PR #3188, opened 2025-09-23
fix float, double and decimal precision handling and culture handling in source generated code

### LINKED ISSUE #3184: Test throw with wrong arguments count when using float
This test
```cs
[Test]
[Arguments("helloworld", "HELLowORLD", 0.19999999f)]
public async Task GetLevenshteinScore(string? value1, string? value2, float score)
{
    await Assert.That(StringHelper.GetLevenshteinScore(value1, value2)).IsEqualTo(score);
}
```

fail with
```cs
Expected exactly 3 arguments, but got 4
```

There is an issue with arguments in source generated part because from VS Test Explorer the test show `0, 20000000` instead of `0.19999999`:
Image

thomhurst:
Fixes #3184


TUnit.Core.SourceGenerator/CodeGenerators/Helpers/TypedConstantParser.cs
```diff
@@ -68,15 +68,9 @@ public static string GetFullyQualifiedTypeNameFromTypedConstantValue(TypedConsta
...
-        // Use the formatter for consistent handling
         return _formatter.FormatForCode(typedConstant, targetType);
     }
 
-    private static string FormatPrimitive(TypedConstant typedConstant)
-    {
-        return FormatPrimitive(typedConstant.Value);
-    }
-
     public static string FormatPrimitive(object? value)
```
Dona278, line 74: Simple question: Why not shared with `TypedConstantFormatter.cs`?

PR #3274, opened 2025-10-03
Expose CurrentRetryAttempt on TestContext

thomhurst:
* Added a new property `CurrentRetryAttempt` to the `TestContext` class to record the current retry attempt for each test execution.
* Updated the `ExecuteWithRetry` method in `RetryHelper.cs` to set the `CurrentRetryAttempt` property on each retry loop iteration.

### DIFF
--- a/TUnit.Core/TestContext.cs
+++ b/TUnit.Core/TestContext.cs
@@ -320,6 +320,8 @@ public class TestContext : Context
 
     public DateTimeOffset? TestEnd { get; set; }
 
+    public int CurrentRetryAttempt { get; internal set; }
+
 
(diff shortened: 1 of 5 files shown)


thomhurst:
@Dona278 - A value of 0 would indicate its not retried yet (it's the first initial run), and counts higher will be how many times it has retried

Dona278:
@thomhurst awesome! Is what I'm searching for.. thank you! I'll try

thomhurst:
> @thomhurst awesome! Is what I'm searching for.. thank you! I'll try

Its not released yet but it'll be out in a couple of hours or so 😄

PR #3296, opened 2025-10-05
Add some documentation around [Culture] attribute

### LINKED ISSUE #2987: Culture Aware unit test?
Is it also possible with TUnit to define that a unit test should run a specific culture?

Like this project which I build for XUnit?
https://github.com/StefH/CultureAwareTesting

andrewimcclement:
This aims to resolve #2987.

`culture.md` cribs a lot from https://docs.nunit.org/articles/nunit/writing-tests/attributes/setculture.html. The example test is essentially taken from `CultureTests.cs`.


thomhurst:
Thanks!

PR #3308, opened 2025-10-06
Add .NET 10 TargetFramework

thomhurst:


TUnit.Engine.Tests/InvokableTestBase.cs
```diff
@@ -174,7 +174,7 @@ private async Task RunWithFailureLogging(Command command, string trxFilename, Li
                                  @$"Output: {commandResult?.StandardOutput}"
 
                                  Expression: {assertionExpression}
-                                 """, e);
```
KifoPL, line 177: Why was this change introduced here?

PR #3316, opened 2025-10-07
Fix Argument Formatters docs code

slang25:
Just spotted some code in need of an update when looking through the docs


thomhurst:
Thanks!

PR #3345, opened 2025-10-09
docs: Added extra information about multiple test projects and PerTestSession usage

stigrune:
I also updated documentation that had references to the removed `SharedType.Globally` enum value.


thomhurst:
Thanks!

PR #3418, opened 2025-10-16
Playwright | additional constructor for page testing

AdemCatamak:
This PR introduces changes that address [the question](https://github.com/thomhurst/TUnit/discussions/622#discussioncomment-13957522) under [discussion](https://github.com/thomhurst/TUnit/discussions/622).

With this update, users can now pass browser launch settings through the constructor when creating their own test classes by inheriting from the PageTest class.

For example, users can set Headless = false via the constructor to control the browser launch behavior according to their needs.

# Motivation

The goal of this change is to provide more flexibility and configurability during browser initialization, improving the overall developer experience when working with PageTest-based tests.


thomhurst:
You need to update the public API snapshots too please

AdemCatamak:
Thanks for the feedback! Could you please clarify how I should update the public API snapshots? Is there a specific command or step I should follow?

thomhurst:
> Thanks for the feedback! Could you please clarify how I should update the public API snapshots? Is there a specific command or step I should follow?

You just need to run the PublicAPI test project, and then you'll get failures due to snapshots needing updating.

Then you can install this: `dotnet tool install -g verify.tool` and then run `dotnet verify accept -y`. Then just commit the updated snapshots

AdemCatamak:
> > Thanks for the feedback! Could you please clarify how I should update the public API snapshots? Is there a specific command or step I should follow?
> 
> You just need to run the PublicAPI test project, and then you'll get failures due to snapshots needing updating.
> 
> Then you can install this: `dotnet tool install -g verify.tool` and then run `dotnet verify accept -y`. Then just commit the updated snapshots

Thank you for the information. The necessary action has been taken.

PR #3749, opened 2025-11-10
refactor(artifacts): change Artifacts to be read-only and update AddArtifact method

thomhurst:


thomhurst:
@claude Run the public API tests and accept and commit the updated snapshots

thomhurst:
@claude investigate the test failure

PR #3917, opened 2025-11-26
Update command for generating TRX report 

kbegiedza:
Previous command results in:

```
Option '--report-trx-filename' has invalid arguments:
file name argument must not contain path (e.g. --report-trx-filename myreport.trx)
```


thomhurst:
Thanks!

PR #3977, opened 2025-12-04
add taskAssert computation expression for simplified assertions in F#

### LINKED ISSUE #3971: Better support for Assert in F#
Hey, 

not sure if you want to add this to your library, but here is a CE that makes the assertions "awaitable" like normale tasks in F#. It's not uncommon to use custom CE's with Task in F# like taskResult, taskOption CE's from [FsErrorHandling](https://demystifyfp.gitbook.io/fstoolkit-errorhandling/fstoolkit.errorhandling)

The code would look like this:

```fsharp
    [<Test>]
    member _.``with check``() = task {
        do! check (Assert.That(1 + 1).IsEqualTo(2))        
    }
    
    
    [<Test>]
    member _.``with CE``() = taskAssert {
        do! Assert.That(1 + 1).IsEqualTo(2)        
    }

```

https://gist.github.com/marcpiechura/352de8eb001f0d1aac457606d211e980

marcpiechura:
As discussed, I've used the new taskAssert expression in all of my tests and it's working as expected, so I think it's good to go. 

closes #3971

### DIFF
--- a/TUnit.TestProject.FSharp/TaskAssertTests.fs
+++ b/TUnit.TestProject.FSharp/TaskAssertTests.fs
@@ -53,3 +53,10 @@ type TaskAssertTests() =
             do! Assert.That(class4.IsInitialized).IsTrue()
             do! Assert.That(class5.IsInitialized).IsTrue()
         }
+
+    [<Test>]
+    [<Category("Fail")>]
+    member _.FailingAssertion_ThrowsCorrectly() : Task =
+        taskAssert {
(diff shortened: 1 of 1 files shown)


thomhurst:
Thanks @marcpiechura ! Do you mind adding the negative assertion test as suggested by Claude? So we can confirm it correctly throws if an assertion fails.

marcpiechura:
Sure thing, I've added the test.

PR #3999, opened 2025-12-06
fix: JUnit reporter

### LINKED ISSUE #3985: CI/CD reporting for GitLab (junit)
Hello 👋 
Firstable, thanks for the amount of work invested in this new test framework!
I'm considering adopting it for my projects, but I'm on GitLab and would very much like to be able to the failing tests right away, like so :
Image
This is only possible if GitLab can parse the generated artifacts. 

[The doc](https://tunit.dev/docs/execution/ci-cd-reporting#gitlab-ci:~:text=GitLab%20can%20parse%20test%20results%20in%20various%20formats) claims that passing `--report-trx` will work with GitLab CI/CD, but the GitLab's doc claims that [only the junit format is accepted](https://docs.gitlab.com/ci/testing/unit_test_reports/#:~:text=Use%20JUnit%20XML%20format%20with%20.xml%20file%20extension.).

I hoped for the best and migrated my tests anyway, but even though the files are generated and detected, they are not parsed properly :/ 
(Prior to the migration i was using [JunitXml.TestLogger ](https://github.com/spekt/junit.testlogger)which supports VSTest platform and not the `Microsoft.Testing.Platform`)

Is there an option to change the output format to JUnit format? Or a known workaround?

rchoffardet:
This should fix junit reporter 
Fixes #3985

### DIFF
--- a/TUnit.Engine/Xml/JUnitXmlWriter.cs
+++ b/TUnit.Engine/Xml/JUnitXmlWriter.cs
@@ -62,7 +62,7 @@ internal static class JUnitXmlWriter
 
         xmlWriter.WriteEndElement(); // testsuites
         xmlWriter.WriteEndDocument();
-
+        xmlWriter.Flush();
         return stringWriter.ToString();
     }
 


thomhurst:
Thanks!

thomhurst:
@rchoffardet new version with these changes should be out

PR #4001, opened 2025-12-06
fix: JUnit reporter was encoding xml files in UTF-16 instead of UTF-8

rchoffardet:
Unfortunately the JUnit Reporter is encoding theX XML files in UTF-16 which is not supported by GitLab
image
Opening the file, i'm seeing that the header contains the wrong encoding: `<?xml version="1.0" encoding="utf-16"?>`

This fix will force the XML files to be encoded in UTF-8: `<?xml version="1.0" encoding="utf-8"?>` 

Fixes https://github.com/thomhurst/TUnit/issues/3985


thomhurst:
New version out again!

PR #4094, opened 2025-12-17
chore: Add ExcludeFromCodeCoverage to generated classes

samtrion:
Generated classes in AssemblyLoaderGenerator and DisableReflectionScannerGenerator now include the [ExcludeFromCodeCoverage] attribute. This ensures that code coverage tools ignore these auto-generated files, resulting in more accurate coverage metrics for user-written code.


thomhurst:
Thanks!

PR #4134, opened 2025-12-21
perf: remove unnecessary manual `string.Replace` code.

TimothyMakkison:
Remove redundant manual `string.Replace` code. `string.Replace` is almost certainly faster, doesn't allocate when there aren't any changes to the `string`, uses `string.Create`, uses `Vectorised` lookups to find targets and won't allocate an intermediary buffer when the input string is larger than 256 characters.

`Contains` call isn't needed because `string.Replace` does the same thing and early returns with the original `string`.


thomhurst:
Thanks!

PR #4172, opened 2025-12-26
perf: replace redundant allocations with `Linq`

TimothyMakkison:
Replace redundant repeat allocations with `Linq`. TUnit has a lot of code that replaces a 24 byte `Linq` allocation (sometimes Linq doesn't even allocate) with a `O(n)` temporary collection allocation.

There are way more of these for instance 0b1df1a7a7cf433c072a2b20088c5c3992b4d2eb, I don't think the AI understand `Linq`, as a rule of thumb, any comment starting with the word `Optimize` or mentions the word `Linq` is suspect

### DIFF
--- a/TUnit.Engine/Services/TestRegistry.cs
+++ b/TUnit.Engine/Services/TestRegistry.cs
@@ -261,7 +261,9 @@ internal sealed class TestRegistry : ITestRegistry
             TestName = testName,
             TestClassType = result.TestClassType,
             TestMethodName = methodInfo.Name,
-            Dependencies = GetDependenciesOptimized(result.Attributes),
+            Dependencies = result.Attributes.OfType<DependsOnAttribute>()
+                .Select(x => x.ToTestDependency())
+                .ToArray(),
             DataSources = [],
             ClassDataSources = [],
             PropertyDataSources = [],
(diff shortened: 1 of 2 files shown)


TUnit.Engine/Services/TestRegistry.cs
```diff
@@ -345,15 +345,9 @@ private sealed class PendingDynamicTest
...
-        var dependencies = new List<TestDependency>(attributes.Count);
-        foreach (var attr in attributes)
-        {
-            if (attr is DependsOnAttribute dependsOn)
-            {
-                dependencies.Add(dependsOn.ToTestDependency());
-            }
-        }
-        return dependencies.ToArray();
+        return attributes.Where(attr => attr is DependsOnAttribute)
```
thomhurst: Would it be better to use OfType<> here?

thomhurst:
Thanks!

TUnit.Engine/Services/TestRegistry.cs
TimothyMakkison: Mb, thinking about it, this method isn't needed. I'll inline it, unless you want to keep it for readability

PR #4164, opened 2025-12-24
perf: eliminate LINQ allocations in hot paths

thomhurst:
## Summary

- Replace LINQ patterns with manual loops and in-place sorting to reduce allocations during test discovery and execution
- Target hot paths in test scheduling, building, and hook collection
- Fixes #4160 (P0: LINQ Elimination in Hot Paths)

## Changes

| File | Change |
|------|--------|
| `ConstraintKeyScheduler.cs` | Replace `OrderBy().ToArray()` with `Array.Sort()`, add `AnyKeyLocked()` helper |
| `TestScheduler.cs` | Remove redundant `OrderBy()` on already-sorted `SortedDictionary` |
| `EventReceiverOrchestrator.cs` | Replace `OrderBy()` with `List.Sort()` |
| `ReflectionTestDataCollector.cs` | Replace `Where()` iterators with inline filtering |
| `TestBuilder.cs` | Add `HasInstanceDataAccessor()` helper, manual loop for error messages |
| `HookCollectionService.cs` | Replace `OrderBy().Select().ToList()` with `List.Sort()` + manual extraction |

## Test plan

- [x] Build succeeds with no errors
- [x] Source generator tests pass (376 tests)
- [x] Assertions tests pass (5,409 tests)
- [x] Engine tests pass (232 pass, 2 pre-existing failures unrelated to changes)
- [x] Benchmark project runs successfully (1,000 tests in ~294ms)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4176, opened 2025-12-27
perf: remove unnecessary intermediary collections

TimothyMakkison:
Remove temporary collections and `ToArray` calls

- I wonder if the `List` capacities are really needed, I don't know enough about the data to tell
  - I don't think I've seen a `List` instantiation without a capacity
- Lots of other weird stuff
- Why does `BuildTestMetadata` return a `Task`? I don't see it calling `async` or other methods. I might have missed something TBH

I've never used claude, but is it possible that the project context is making it allergic to `Linq`, overusing `List` capacity and creating unneeded temporary collections?

### DIFF
--- a/TUnit.Engine/Discovery/ReflectionTestDataCollector.cs
+++ b/TUnit.Engine/Discovery/ReflectionTestDataCollector.cs
@@ -92,16 +92,14 @@ internal sealed class ReflectionTestDataCollector : ITestDataCollector
                 assembliesList.Add(assembly);
             }
         }
-        var assemblies = assembliesList;
 
-
-        var maxConcurrency = Math.Min(assemblies.Count, Environment.ProcessorCount * 2);
+        var maxConcurrency = Math.Min(assembliesList.Count, Environment.ProcessorCount * 2);
         var semaphore = new SemaphoreSlim(maxConcurrency, maxConcurrency);
-        var tasks = new Task<List<TestMetadata>>[assemblies.Count];
+        var tasks = new Task<List<TestMetadata>>[assembliesList.Count];
 
-        for (var i = 0; i < assemblies.Count; i++)
+        for (var i = 0; i < assembliesList.Count; i++)
         {
-            var assembly = assemblies[i];
+            var assembly = assembliesList[i];
             var index = i;
 
             tasks[index] = ProcessAssemblyAsync(assembly, semaphore);
@@ -138,19 +136,15 @@ internal sealed class ReflectionTestDataCollector : ITestDataCollector
     {
         // Get assemblies to scan
         var allAssemblies = Assemblies;
-        var assemblies = new List<Assembly>(allAssemblies.Length);
-        foreach (var assembly in allAssemblies)
-        {
-            if (ShouldScanAssembly(assembly))
-            {
-                assemblies.Add(assembly);
-            }
-        }
-
 
         // Stream tests from each assembly
-        foreach (var assembly in assemblies)
+        foreach (var assembly in allAssemblies)
         {
+            if(!ShouldScanAssembly(assembly))
+            {
+                continue;
+            }
+
             cancellationToken.ThrowIfCancellationRequested();
 
             // Use lock-free ConcurrentDictionary for assembly tracking
@@ -359,7 +353,7 @@ internal sealed class ReflectionTestDataCollector : ITestDataCollector
                 continue;
             }
 
-            MethodInfo[] testMethods;
+            IEnumerable<MethodInfo> testMethods;
             try
             {
                 // Check if this class inherits tests from base classes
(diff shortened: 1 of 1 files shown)


TimothyMakkison:
Okay, I've removed all remaining `Linq` and switched back to array in the cache. 

I looked at the benchmarking scripts in #4164, (I think run-baseline is broken) and didn't get any valuable information from it. I'm probably reading the flamegraph wrong but I couldn't see any TUnit code being called, I'll try again tomorrow. Would be super helpful if I knew where the problem areas were 😅

thomhurst:
## Summary
Removes temporary collections and ToArray() calls to reduce allocations in test discovery hot paths.

## Critical Issues

### 1. Multiple Enumeration Risk in DiscoverTestsInAssembly (lines 362-372)

The change from `MethodInfo[]` to `IEnumerable<MethodInfo>` creates a **multiple enumeration anti-pattern**:

```csharp
// After PR - problematic
IEnumerable<MethodInfo> testMethods;
testMethods = allMethods.Where(method => method.IsDefined(typeof(TestAttribute), inherit: false) && !method.IsAbstract);

foreach (var method in testMethods)  // Enumerates the LINQ query
{
    discoveredTests.Add(await BuildTestMetadata(type, method).ConfigureAwait(false));
}
```

**Problem**: The LINQ `.Where()` query will be re-executed for **every iteration** of the foreach loop because it returns a lazy `IEnumerable`. This means:
- The `IsDefined` attribute check runs N² times instead of N times
- Significantly worse performance than the original code
- Defeats the entire purpose of this "performance" PR

**Evidence**: The original code explicitly avoided this with the comment on line 371: `// Optimize: Manual filtering instead of LINQ Where().ToArray()`

**Fix**: Either keep the manual loop + ToArray(), or use `.ToList()` to materialize once:
```csharp
testMethods = allMethods.Where(method => method.IsDefined(typeof(TestAttribute), inherit: false) && !method.IsAbstract).ToList();
```

This affects **4 locations**:
- ReflectionTestDataCollector.cs:365 (DiscoverTestsInAssembly - inheritsTests path)
- ReflectionTestDataCollector.cs:370 (DiscoverTestsInAssembly - declaredMethods path)  
- Similar pattern in DiscoverGenericTests and DiscoverGenericTestsStreamingAsync

### 2. Multiple Enumeration Risk in IsCovariantCompatible (line 1334)

```csharp
// After PR
IEnumerable<Type> argInterfaces = AssemblyReferenceCache.GetInterfaces(argType);
if (argType.IsInterface)
{
    argInterfaces = argInterfaces.Concat([argType]);  // Lazy concatenation
}

foreach (var iface in argInterfaces)  // Could re-enumerate GetInterfaces
```

**Problem**: If `GetInterfaces()` returns an `IEnumerable` (not array/list), the `.Concat()` creates a lazy chain that will call `GetInterfaces()` **for every iteration**. We need to verify what `AssemblyReferenceCache.GetInterfaces()` returns.

**Fix**: Add `.ToArray()` after the Concat to materialize once.

## Suggestions

### 1. Reconsider Removed Optimizations

The PR removes explicit "manual filtering" optimizations that were intentionally added. The comments indicated these were performance improvements. Before removing them, benchmarks should verify the LINQ versions are actually faster.

### 2. List Capacity Pre-allocation

The author questions if List capacities are needed. In hot paths (test discovery runs thousands of times), pre-allocating capacity **does** help reduce allocations. The original code's capacity hints were appropriate.

## Previous Review Status
Unable to retrieve previous comments due to GitHub API scopes.

## Verdict
⚠️ **REQUEST CHANGES** - Multiple enumeration bugs will make performance worse, not better.

thomhurst:
## Summary
Removes unnecessary intermediate collections and array allocations in test discovery code.

## Critical Issues
None found

## Suggestions

Performance validation: While these changes are solid optimizations, consider verifying that:
1. The switch from eager (.ToArray()) to lazy evaluation (.Where()) in DiscoverTestsInAssembly does not cause multiple enumeration issues. Currently safe since the result is only enumerated once in the foreach loop.
2. The removed .ToArray() call in IsCovariantCompatible (line 1334) is appropriate - looks good since the result is used in a single foreach loop.

Regarding the PR author questions:

The capacities in places like DiscoverTestsInAssembly (line 323: new List TestMetadata 100) and GetDataFromSourceAsync (line 778: new List object array 16) are performance optimizations to avoid List growth/reallocation. Based on TUnit Rule 4 Performance First, keeping these capacities is appropriate.

BuildTestMetadata returns a Task (line 830) for API consistency with async discovery workflows, even though this specific path uses Task.FromResult(). Changing this is outside the scope of this PR.

Code quality: The changes improve readability and reduce allocations. The early-continue pattern in CollectTestsStreamingAsync and DiscoverDynamicTests is cleaner than the previous nested approach.

## Verdict
APPROVE - No critical issues

This PR successfully removes unnecessary allocations in the test discovery hot path, aligning with TUnit performance-first principles. The changes are correct and improve code clarity.

PR #4195, opened 2025-12-30
perf: skip `List` creation in `GetDynamicTestAttributes`

TimothyMakkison:
Skip `List` creation if `TestClassType` is `null`. 
- Would be a good candidate for `ValueListBuilder` if this is a hot path

### DIFF
--- a/TUnit.Engine/Building/Collectors/AotTestDataCollector.cs
+++ b/TUnit.Engine/Building/Collectors/AotTestDataCollector.cs
@@ -155,15 +155,17 @@ internal sealed class AotTestDataCollector : ITestDataCollector
 
     private static Attribute[] GetDynamicTestAttributes(DynamicDiscoveryResult result)
     {
+        if (result.TestClassType == null)
+        {
+            return result.Attributes.ToArray();
+        }
+
         // Merge explicitly provided attributes with inherited class/assembly attributes
         // Order matches GetAllAttributes: method-level first (explicit), then class, then assembly
         var attributes = new List<Attribute>(result.Attributes);
 
-        if (result.TestClassType != null)
-        {
-            attributes.AddRange(result.TestClassType.GetCustomAttributes().OfType<Attribute>());
-            attributes.AddRange(result.TestClassType.Assembly.GetCustomAttributes().OfType<Attribute>());
-        }
+        attributes.AddRange(result.TestClassType.GetCustomAttributes());
+        attributes.AddRange(result.TestClassType.Assembly.GetCustomAttributes());
 
         return attributes.ToArray();
     }
--- a/TUnit.Engine/Discovery/ReflectionTestDataCollector.cs
+++ b/TUnit.Engine/Discovery/ReflectionTestDataCollector.cs
@@ -1932,15 +1932,17 @@ internal sealed class ReflectionTestDataCollector : ITestDataCollector
 
     private static Attribute[] GetDynamicTestAttributes(DynamicDiscoveryResult result)
     {
+        if (result.TestClassType == null)
+        {
+            return result.Attributes.ToArray();
+        }
+
         // Merge explicitly provided attributes with inherited class/assembly attributes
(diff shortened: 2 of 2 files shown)


TUnit.Engine/Building/Collectors/AotTestDataCollector.cs
```diff
@@ -155,15 +155,17 @@ private Task<TestMetadata> CreateMetadataFromDynamicDiscoveryResult(DynamicDisco
...
         // Order matches GetAllAttributes: method-level first (explicit), then class, then assembly
         var attributes = new List<Attribute>(result.Attributes);
 
-        if (result.TestClassType != null)
-        {
-            attributes.AddRange(result.TestClassType.GetCustomAttributes().OfType<Attribute>());
-            attributes.AddRange(result.TestClassType.Assembly.GetCustomAttributes().OfType<Attribute>());
-        }
+        attributes.AddRange(result.TestClassType.GetCustomAttributes().OfType<Attribute>());
+        attributes.AddRange(result.TestClassType.Assembly.GetCustomAttributes().OfType<Attribute>());
```
wallstop: If you wanted perf, would skipping LINQ here and just iterating + manual type checking + manual `Add` calls allocate less (and be more performant)? 🤔

TimothyMakkison: Good point, although it's possible it doesn't need to allocate `OfTypeIterator` because there might be a backing array of `Attribute`. I don't know what contexts this would be true under.

image



tbh I initially hoped `GetCustomAttribute` returned an array so I'd be able to pre allocate an array and copy stuff into it manually liked you suggested😅

wallstop: Looks like `GetCustomAttributes` is TUnit's own special-sauce that does return an array, but as an IEnumerable - might be worth using something internal/tweaking this/making a variant that is more performant?

https://github.com/thomhurst/TUnit/blob/61ac7964cf5cffb290142bb61d740327a8782813/TUnit.Core/Extensions/MetadataExtensions.cs#L27

Unless I'm wrong? [C#'s built-in](https://learn.microsoft.com/en-us/dotnet/api/system.attribute.getcustomattributes?view=net-10.0) looks like it returns arrays.

thomhurst: That exists because the C# built in throws on generic attributes in .NET Framework. Feel free to rework it to return an array :smile:

TimothyMakkison: Turns out the `OfType<Attribute>` was never needed 😆 

image

> Looks like GetCustomAttributes is TUnit's own special-sauce that does return an array

According to Rider this is only used in two other select areas. Here we are calling `System.Reflection.GetCustomAttribute`, I'm not sure what TUnit uses that method for 🤔

TimothyMakkison: tbh I might bite the bullet and introduce `ValueListBuilder` and hope that this a hot path🤷 I feel like thomhurst would know where the problematic areas are


Edit: this is a silly idea, I should figure out how to attach a profiler to TUnit, or actually understand what TUnit calls and when before doing that. 

It may be possible to just preallocate an array and skip that issue

thomhurst: Tunit test suites are just normal console apps so normal profiling should work. For me I've used the dotnet trace tool to collect trace files and then analyse after

thomhurst:
APPROVE - Clean performance optimization. Avoids List allocation when TestClassType is null and removes redundant .OfType conversions. Both AotTestDataCollector and ReflectionTestDataCollector updated identically (dual-mode Rule #1). Reduces allocations in test discovery hot path (Rule #4 Performance First). No critical issues found.

PR #4208, opened 2026-01-01
fix: add TODO comments for unsupported NUnit TestCase properties

thomhurst:
## Summary
- Fixes crash when Roslyn formatter processes inline comments in attributes
- Properly adds trailing TODO comments to attribute lists for unsupported TestCase properties (TestName, Description, Category, Author, etc.)
- Maps NUnit's Ignore/IgnoreReason to TUnit's Skip property
- Adds 7 comprehensive tests covering all TestCase property scenarios

## What was fixed
1. **Crash issue**: Adding `/* TODO: ... */` as leading trivia to `AttributeArgument` created invalid syntax that crashed Roslyn's formatter
2. **Lost properties**: Unsupported properties were being dropped silently without any TODO comment

## Changes
- `NUnitAttributeRewriter.VisitAttributeList`: Override to handle TestCase attributes specially and add TODO comments as trailing trivia on `AttributeListSyntax`
- `NUnitExpectedResultRewriter.TransformTestCaseAttribute`: Return tuple with unsupported properties for proper comment generation
- 7 new tests covering: TestName, multiple properties, Ignore, IgnoreReason, ExpectedResult with properties, mixed properties, no comment case

## Test plan
- [x] All 38 NUnit migration tests pass
- [x] All 21 MSTest migration tests pass  
- [x] All 33 xUnit migration tests pass (1 pre-existing skip)
- [x] New tests verify TODO comments appear correctly

### DIFF
--- a/TUnit.Analyzers.CodeFixers/NUnitExpectedResultRewriter.cs
+++ b/TUnit.Analyzers.CodeFixers/NUnitExpectedResultRewriter.cs
@@ -391,6 +391,7 @@ public class NUnitExpectedResultRewriter : CSharpSyntaxRewriter
         foreach (var attrList in attributeLists)
         {
             var newAttributes = new List<AttributeSyntax>();
+            var unsupportedPropertiesList = new List<string>();
 
             foreach (var attr in attrList.Attributes)
             {
@@ -404,8 +405,9 @@ public class NUnitExpectedResultRewriter : CSharpSyntaxRewriter
                 else if (testCaseAttributes.Contains(attr))
                 {
                     // Transform TestCase with ExpectedResult
-                    var transformed = TransformTestCaseAttribute(attr);
+                    var (transformed, unsupportedProperties) = TransformTestCaseAttribute(attr);
                     newAttributes.Add(transformed);
+                    unsupportedPropertiesList.AddRange(unsupportedProperties);
                 }
                 else
                 {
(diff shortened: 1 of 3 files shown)


thomhurst:
## Summary
Fixes crash when adding TODO comments for unsupported NUnit TestCase properties by changing from inline block comments to trailing line comments on attribute lists.

## Critical Issues
None found ✅

## Suggestions

### Code Quality: Duplicate logic in VisitAttributeList override

The new `VisitAttributeList` override in `NUnitAttributeRewriter` (lines 80-141) duplicates significant logic from the base class `AttributeRewriter.VisitAttributeList` (lines 225-261 in BaseMigrationCodeFixProvider.cs).

Specifically:
- Both iterate through attributes
- Both check `ShouldRemoveAttribute` and `IsHookAttribute`
- Both call `ConvertAttribute` for non-special cases

The only unique behavior is tracking `droppedPropertiesList` for TestCase attributes. Consider refactoring to minimize duplication:

**Option 1:** Add a hook method in base class that derived classes can override to add trailing trivia
**Option 2:** Extract the TestCase-specific logic into a helper method and call it from the existing flow

This would make the code more maintainable and reduce risk of divergence between the two implementations.

### Minor: ConvertAttributeArguments signature change impact

Line 72 now accesses `.arguments` on the tuple return from `ConvertTestCaseArguments`. While this works correctly, it means `ConvertAttributeArguments` for TestCase has a different call pattern than other attribute types.

This is correct but slightly inconsistent - consider documenting why TestCase returns a tuple if not already clear from context.

## Verdict

The fix correctly solves the Roslyn formatter crash by using trailing line comments instead of leading block comments. The comprehensive test coverage (7 new tests) validates all scenarios. The duplicate logic is a code quality concern but not blocking.

thomhurst:
Closing to implement proper property mapping instead of TODO comments. NUnit TestCase properties can be properly converted to TUnit equivalents:
- TestName → [DisplayName]
- Category → [Category]  
- Description → [Property("Description", ...)]
- Author → [Property("Author", ...)]
- Explicit → [Explicit]

PR #4221, opened 2026-01-03
fix: handle ref struct parameters in [GenerateAssertion] source generator

### LINKED ISSUE #4192: [Bug]: Fix Source Generation for methods using `ref struct` (e.g., `DefaultInterpolatedStringHandler`
### Description

The current TUnit source generator fails when it encounters extension methods that use types marked with `[InterpolatedStringHandlerAttribute]`, such as `DefaultInterpolatedStringHandler`.

Since these types are `ref struct`s, the generated assertion classes are invalid because they attempt to store a `ref struct` as a field in a regular `class`, which is prohibited by C# language rules.

#### **Example of the problematic source code**
```csharp
[EditorBrowsable(EditorBrowsableState.Never)]
[GenerateAssertion(ExpectationMessage = "to be contains log {message}")]
internal static bool Contains(this FakeLogger logger, ref DefaultInterpolatedStringHandler message)
{
    // ...
}
```

```csharp
public sealed class FakeLogger_Contains_DefaultInterpolatedStringHandler_Assertion : Assertion<Microsoft.Extensions.Logging.Testing.FakeLogger>
{
    private readonly System.Runtime.CompilerServices.DefaultInterpolatedStringHandler _message; // error

    public FakeLogger_Contains_DefaultInterpolatedStringHandler_Assertion(AssertionContext<Microsoft.Extensions.Logging.Testing.FakeLogger> context, System.Runtime.CompilerServices.DefaultInterpolatedStringHandler message)
        : base(context)
    {
        _message = message; 
    }

    protected override Task<AssertionResult> CheckAsync(EvaluationMetadata<Microsoft.Extensions.Logging.Testing.FakeLogger> metadata)
    {
        var value = metadata.Value;
        var exception = metadata.Exception;

        if (exception != null)
        {
            return Task.FromResult(AssertionResult.Failed($"threw {exception.GetType().FullName}"));
        }

        if (value is null)
        {
            return Task.FromResult(AssertionResult.Failed("Actual value is null"));
        }

        var result = value!.Contains(_message); // error
        return Task.FromResult(result
            ? AssertionResult.Passed
            : AssertionResult.Failed($"found {value}"));
    }

    protected override string GetExpectation()
    {
        return $"to be contains log {_message}"; // error
    }
}
```

#### **Observed Issues in Generated Code**
1.  **Field violation**: The generator creates a field `private readonly DefaultInterpolatedStringHandler _message;` inside a class. This causes: `CS8345: Field cannot be of byref-like type unless it is an instance member of a 'ref' struct`.
2.  **Argument mismatch**: It tries to pass the handler by value instead of `ref`.
3.  **Generic constraints**: It attempts to use the handler as a generic type argument, which fails for `ref struct`s.

#### **Proposed Solution**
The source generator should detect if a parameter type is an interpolated string handler. The best way to achieve this is to check for the presence of the `System.Runtime.CompilerServices.InterpolatedStringHandlerAttribute` on the parameter's type.

**Refinement Logic:**
1.  **Detection**: During the metadata analysis phase of the Source Generator, check the parameter type:
```csharp
var isInterpolatedHandler = parameterSymbol.Type.GetAttributes()
        .Any(attr => attr.AttributeClass?.ToDisplayString() == "System.Runtime.CompilerServices.InterpolatedStringHandlerAttribute");
```

2.  **Handling**: If a type is an interpolated string handler (or any `ref struct`), the generator should:
    *   Avoid storing the raw `ref struct` in the assertion class.
    *   **Convert to String**: Instead of storing the handler, the generator should capture the result of `.ToStringAndClear()` or convert the argument to a `string` before passing it to the assertion constructor.
    *   Adjust the generated constructor and field to use `string` instead of the handler type.

### Expected Behavior

The Source Generator should detect that the parameter type is a ref struct (e.g., by checking for InterpolatedStringHandlerAttribute) and handle it accordingly.

Since ref struct types cannot be stored as fields in classes, the generator should:
* Convert the interpolated string handler to a standard string before passing it to the assertion class constructor;
* Generate the assertion class with a string field instead of a ref struct field;
* Ensure the generated code is compliant with C# safety rules regarding stack-only types.

### Actual Behavior

The Source Generator attempts to create a standard Assertion<T> class that stores the ref struct directly as a field. 

This leads to multiple compilation errors in the generated code:
* `CS8345`: Field cannot be of byref-like type 'DefaultInterpolatedStringHandler' unless it is an instance member of a 'ref' struct;
* `CS1615`: Argument is passed as 'value' while the parameter is declared as 'ref';
* `CS9244`: The type may not be a ref struct in order to use it as a type argument.

Essentially, the generated code is invalid because it tries to "box" or store a stack-only type within a heap-allocated class.

### Steps to Reproduce

* Add package `Microsoft.Extensions.Diagnostics.Testing`;
* Define an extension method intended for assertion generation that takes a ref struct parameter (specifically an interpolated string handler). Example:
``` csharp
   [GenerateAssertion(ExpectationMessage = "to be contains log {message}")]
   internal static bool Contains(this FakeLogger logger, ref DefaultInterpolatedStringHandler message)
   {
       var stringMessage = message.ToStringAndClear();
       return logger.Collector.GetSnapshot().Select(static x => x.Message).Contains(stringMessage);
   }
```
* Observe the generated code in `TUnit.Assertions.Extensions`.

### TUnit Version

1.7.7

### .NET Version

.net9.0

### Operating System

Windows

### IDE / Test Runner

JetBrains Rider

```shell

```

### Additional Context

_No response_

### IDE-Specific Issue?

thomhurst:
## Summary

Fixes #4192

The source generator was failing when encountering methods with `ref struct` parameters (like `DefaultInterpolatedStringHandler`) because it tried to store them as class fields, which is illegal in C#.

### Changes

- **Add ref struct detection methods**: `IsRefStruct()` and `IsInterpolatedStringHandler()` to detect ref-like types using the `IsRefLikeType` property and fallback pattern matching
- **Store ref struct parameters as string**: Fields for ref struct parameters now use `string` type instead of the original type
- **Convert in extension method**: Interpolated string handlers are converted to string via `.ToStringAndClear()` before passing to the assertion constructor; other ref structs use `.ToString()`
- **Fix inlined method bodies**: Remove redundant `.ToStringAndClear()`/`.ToString()` calls in inlined method bodies since the field is already a string
- **Add diagnostic TUNITGEN004**: Error when methods with ref struct parameters don't use `InlineMethodBody = true` (required because we can't call the original method with a string when it expects a ref struct)

### Example

For a method like:
```csharp
[GenerateAssertion(ExpectationMessage = "to contain {message}", InlineMethodBody = true)]
public static bool ContainsMessage(this string value, ref DefaultInterpolatedStringHandler message)
{
    var stringMessage = message.ToStringAndClear();
    return value.Contains(stringMessage);
}
```

The generator now produces:
1. A field: `private readonly string _message;` (instead of storing the handler)
2. Extension method converts: `message.ToStringAndClear()` before passing to constructor
3. Inlined body: `value!.Contains(_message)` (with `.ToStringAndClear()` removed since it's already a string)

## Test plan

- [x] Added test data file `RefStructParameterAssertion.cs` with interpolated string handler parameter examples
- [x] Added test case `RefStructParameter()` to verify the generated code
- [x] Full solution builds successfully
- [ ] Run source generator tests when test infrastructure version mismatch is resolved

### DIFF
--- a/TUnit.Assertions.SourceGenerator.Tests/MethodAssertionGeneratorTests.cs
+++ b/TUnit.Assertions.SourceGenerator.Tests/MethodAssertionGeneratorTests.cs
@@ -201,4 +201,37 @@ internal class MethodAssertionGeneratorTests : TestsBase<MethodAssertionGenerato
             // Snapshot test - the actual verification is done by snapshot comparison
             await Assert.That(generatedFiles.Count).IsGreaterThanOrEqualTo(1);
         });
+
+#if NET6_0_OR_GREATER
+    [Test]
+    public Task RefStructParameter() => RunTest(
+        Path.Combine(Sourcy.Git.RootDirectory.FullName,
+            "TUnit.Assertions.SourceGenerator.Tests",
+            "TestData",
+            "RefStructParameterAssertion.cs"),
+        async generatedFiles =>
+        {
+            await Assert.That(generatedFiles).HasCount(1);
+
+            var mainFile = generatedFiles.First();
+            await Assert.That(mainFile).IsNotNull();
+
+            // Verify that the field type is string, not the ref struct
+            await Assert.That(mainFile).Contains("private readonly string _message;");
+            await Assert.That(mainFile).Contains("private readonly string _suffix;");
+
+            // Verify that the extension method converts the ref struct to string
+            await Assert.That(mainFile).Contains("message.ToStringAndClear()");
+            await Assert.That(mainFile).Contains("suffix.ToStringAndClear()");
+
+            // Verify the constructor takes string, not the ref struct
+            await Assert.That(mainFile).Contains("string message)");
+            await Assert.That(mainFile).Contains("string suffix)");
+
+            // Verify that .ToStringAndClear() is removed in the inlined body
+            // (since the field is already a string)
+            // The inlined body should use _message directly, not _message.ToStringAndClear()
+            await Assert.That(mainFile).Contains("value!.Contains(_message)");
+        });
+#endif
 }
--- a/TUnit.Assertions.SourceGenerator.Tests/TestData/RefStructParameterAssertion.cs
+++ b/TUnit.Assertions.SourceGenerator.Tests/TestData/RefStructParameterAssertion.cs
@@ -1,0 +1,30 @@
+#if NET6_0_OR_GREATER
+using System.Runtime.CompilerServices;
+using TUnit.Assertions.Attributes;
+
+namespace TUnit.Assertions.Tests.TestData;
+
+/// <summary>
+/// Test case: Method with ref struct parameter (DefaultInterpolatedStringHandler)
+/// The generator should convert the ref struct to string before storing it
+/// </summary>
+public static class RefStructParameterAssertions
+{
+    /// <summary>
+    /// Test that interpolated string handlers are properly converted to strings
+    /// </summary>
+    [GenerateAssertion(ExpectationMessage = "to contain {message}", InlineMethodBody = true)]
+    public static bool ContainsMessage(this string value, ref DefaultInterpolatedStringHandler message)
+    {
+        var stringMessage = message.ToStringAndClear();
+        return value.Contains(stringMessage);
+    }
+
+    /// <summary>
+    /// Test with a simpler expression body
+    /// </summary>
+    [GenerateAssertion(ExpectationMessage = "to end with {suffix}", InlineMethodBody = true)]
+    public static bool EndsWithMessage(this string value, ref DefaultInterpolatedStringHandler suffix)
+        => value.EndsWith(suffix.ToStringAndClear());
+}
+#endif
--- a/TUnit.Assertions.SourceGenerator/Generators/MethodAssertionGenerator.cs
+++ b/TUnit.Assertions.SourceGenerator/Generators/MethodAssertionGenerator.cs
@@ -47,6 +47,15 @@ public sealed class MethodAssertionGenerator : IIncrementalGenerator
         isEnabledByDefault: true,
         description: "Methods decorated with [GenerateAssertion] must return bool, AssertionResult, Task<bool>, or Task<AssertionResult>.");
 
+    private static readonly DiagnosticDescriptor RefStructRequiresInliningRule = new DiagnosticDescriptor(
+        id: "TUNITGEN004",
+        title: "Ref struct parameter requires method body inlining",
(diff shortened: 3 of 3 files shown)


thomhurst:
## Summary
Fixes ref struct parameter handling in the assertion source generator by converting ref structs (like `DefaultInterpolatedStringHandler`) to strings before storage, since ref structs cannot be class fields.

## Critical Issues
None found ✅

## Suggestions

### 1. Reflection-based detection could be more robust
**Location:** `TUnit.Assertions.SourceGenerator/Generators/MethodAssertionGenerator.cs:1076-1079`

The `IsRefStruct` method uses reflection to access `IsRefLikeType`:
```csharp
var isRefLikeProperty = namedType.GetType().GetProperty("IsRefLikeType");
if (isRefLikeProperty?.GetValue(namedType) is bool isRefLike && isRefLike)
```

While you have a fallback, consider checking the Roslyn version you're targeting. Modern Roslyn versions (4.0+) expose `INamedTypeSymbol.IsRefLikeType` directly. If you're already on a modern version, you could use it directly without reflection:
```csharp
if (namedType.IsRefLikeType)
    return true;
```

### 2. Missing snapshot verification files
The new test `RefStructParameter` uses Verify's snapshot testing, but no `.verified.txt` files are included in the PR. According to TUnit rule #2: "Changes to source generator output... require running snapshot tests. Commit `.verified.txt` files."

After the test infrastructure version mismatch is resolved and tests run successfully, ensure you commit the generated `.verified.txt` files for the new test case.

### 3. Regex edge cases
**Location:** `TUnit.Assertions.SourceGenerator/Generators/MethodAssertionGenerator.cs:770-779`

The regex patterns remove `.ToStringAndClear()` and `.ToString()` from field references:
```csharp
inlinedBody = Regex.Replace(
    inlinedBody,
    $@"{Regex.Escape(fieldName)}\.ToStringAndClear\(\)",
    fieldName);
```

This looks correct, but be aware that it will match even in string literals or comments. Since you're working with parsed method bodies this is likely safe, but if edge cases arise, you might need to use Roslyn's semantic model for more precise replacement.

## Previous Review Status
Unable to check previous comments due to GitHub token scope limitations.

## Verdict
✅ **APPROVE** - No critical issues. The implementation correctly handles the ref struct constraint, adds appropriate diagnostics, and follows TUnit patterns. The suggestions are minor improvements that don't block merging.

PR #4225, opened 2026-01-03
perf: skip `Enumerator`, `Hashset` and state machine

TimothyMakkison:
- Add early return when there aren't any objects to be tracked. Skipping `Enumerator` and `HashSet` allocations
- Defer creation of `async` state machine when no objects are tracked

See #4215

#### Before 
image

Note that `ObjectTracker+<UntrackObjects>d__11` is the state machine for `UntrackObjects`

#### After
image

`ObjectTracker+<Inner>d__12` is the state machine for the method now called `UntrackObjectsAsync`

### DIFF
--- a/TUnit.Core/Tracking/ObjectTracker.cs
+++ b/TUnit.Core/Tracking/ObjectTracker.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Concurrent;
+using System.Collections.Immutable;
 using System.Diagnostics.CodeAnalysis;
 using TUnit.Core.Helpers;
 
@@ -48,12 +49,17 @@ internal class ObjectTracker(TrackableObjectGraphProvider trackableObjectGraphPr
         s_trackedObjects.GetOrAdd(obj, static _ => new Counter());
 
     /// <summary>
-    /// Flattens a ConcurrentDictionary of depth-keyed HashSets into a single HashSet.
+    /// Flattens a ConcurrentDictionary of depth-keyed HashSets into a single ISet.
     /// Thread-safe: locks each HashSet while copying.
     /// Pre-calculates capacity to avoid HashSet resizing during population.
     /// </summary>
-    private static HashSet<object> FlattenTrackedObjects(ConcurrentDictionary<int, HashSet<object>> trackedObjects)
+    private static ISet<object> FlattenTrackedObjects(ConcurrentDictionary<int, HashSet<object>> trackedObjects)
     {
+        if (trackedObjects.IsEmpty)
+        {
+            return ImmutableHashSet<object>.Empty;
+        }
+
 #if NETSTANDARD2_0
         // .NET Standard 2.0 doesn't support HashSet capacity constructor
         var result = new HashSet<object>(Helpers.ReferenceEqualityComparer.Instance);
@@ -91,8 +97,13 @@ internal class ObjectTracker(TrackableObjectGraphProvider trackableObjectGraphPr
         var alreadyTracked = FlattenTrackedObjects(testContext.TrackedObjects);
 
         // Get new trackable objects
-        var newTrackableObjects = new HashSet<object>(Helpers.ReferenceEqualityComparer.Instance);
         var trackableDict = trackableObjectGraphProvider.GetTrackableObjects(testContext);
+        if (trackableDict.IsEmpty && alreadyTracked.Count == 0)
+        {
+            return;
+        }
+
+        var newTrackableObjects = new HashSet<object>(Helpers.ReferenceEqualityComparer.Instance);
         foreach (var kvp in trackableDict)
         {
             lock (kvp.Value)
@@ -113,11 +124,20 @@ internal class ObjectTracker(TrackableObjectGraphProvider trackableObjectGraphPr
         }
     }
 
-    public async ValueTask UntrackObjects(TestContext testContext, List<Exception> cleanupExceptions)
+    public ValueTask UntrackObjects(TestContext testContext, List<Exception> cleanupExceptions)
     {
         // Get all objects to untrack (DRY: use helper method)
         var objectsToUntrack = FlattenTrackedObjects(testContext.TrackedObjects);
+        if (objectsToUntrack.Count == 0)
+        {
+            return ValueTask.CompletedTask;
+        }
 
+        return UntrackObjectsAsync(cleanupExceptions, objectsToUntrack);
+    }
+
+    private async ValueTask UntrackObjectsAsync(List<Exception> cleanupExceptions, ISet<object> objectsToUntrack)
+    {
         foreach (var obj in objectsToUntrack)
         {
             try


thomhurst:
## Summary
Performance optimization to skip allocations in ObjectTracker when there are no objects to track.

## Critical Issues
None found ✅

## Suggestions

### 1. Consider NetStandard2.0 compatibility
The PR adds ImmutableHashSet.Empty which requires System.Collections.Immutable. Verify this package is available for NetStandard2.0 target.

### 2. Update method documentation
Line 51-52: XML doc says method returns HashSet but now returns ISet. Consider updating the comment.

## Performance Impact
Optimizations are well-targeted:
- Avoids Enumerator allocation when dictionary is empty
- Avoids HashSet allocation when no objects to track  
- Defers async state machine allocation when no objects to untrack
- Returns cached ValueTask.CompletedTask

Aligns perfectly with TUnit's Performance First principle for hot paths.

## Verdict
✅ APPROVE - Solid performance optimization with no critical issues.

TimothyMakkison:
Before I forget, it might be possible to only create the state machine in `UntrackObject` when it actually needs to be `async` using `ValueTask`. Can't remember how to do this sanely though

TimothyMakkison:
Ok, updated the comment and I did verify that `TUnit.Core` references `System.Collections.Immutable`. If this were an issue I could create a single static empty `HashSet`.

The only quibble I can think of, is using a condtional operator on line 131

thomhurst:
Not an issue. Those reviews are AI assisted and can be overly picky!

thomhurst:
## Summary
Performance optimization that adds early returns to skip allocations when there are no objects to track/untrack, and defers async state machine creation in the hot path.

## Critical Issues
None found ✅

## Suggestions

### 1. Consider removing the capacity pre-calculation entirely
The PR adds an early return for empty collections, which is good. However, the `FlattenTrackedObjects` method still has the double-pass capacity pre-calculation on non-NETSTANDARD2_0 targets (lines 62-72 in the original file).

Based on issue #4215's rationale, you might want to consider removing this pre-calculation entirely in a follow-up. The cost of:
- Second `ConcurrentDictionary.Enumerator` allocation
- Second round of locking each HashSet
- Second iteration through all entries

...likely outweighs the benefit of pre-sizing the HashSet to avoid a few resize operations. Modern HashSets resize efficiently, and the double enumeration + locking overhead is measurable.

**Suggested approach** (for a follow-up PR, not blocking this one):
```csharp
private static ISet<object> FlattenTrackedObjects(ConcurrentDictionary<int, HashSet<object>> trackedObjects)
{
    if (trackedObjects.IsEmpty)
    {
        return ImmutableHashSet<object>.Empty;
    }

    var result = new HashSet<object>(Helpers.ReferenceEqualityComparer.Instance);
    foreach (var kvp in trackedObjects)
    {
        lock (kvp.Value)
        {
            foreach (var obj in kvp.Value)
            {
                result.Add(obj);
            }
        }
    }
    return result;
}
```

This would be consistent with the performance-first philosophy in this PR.

### 2. `ImmutableHashSet.Empty` usage is appropriate
The use of `ImmutableHashSet<object>.Empty` for the empty case is good - it's a singleton with zero allocations. This is better than allocating an empty `HashSet`.

### 3. Performance impact looks solid
Based on the screenshots in the PR description:
- Before: 116,224 `HashSet` allocations, 14,400 `Enumerator` allocations, 14,400 state machines
- After: 101,824 `HashSet` allocations, 0 `Enumerator` allocations, 0 state machines

This confirms the optimization is working as intended in the hot path (tests with no tracked objects).

## Verdict
✅ **APPROVE** - No critical issues. The early returns avoid allocations effectively, and deferring the async state machine is a textbook hot-path optimization.

thomhurst:
## Summary
Performance optimization that adds early returns and defers async state machine creation in ObjectTracker to avoid unnecessary allocations when no objects are tracked.

## Critical Issues
None found ✅

## Suggestions

### 1. Consider ImmutableHashSet comparer compatibility (TUnit.Core/Tracking/ObjectTracker.cs:60)

The change returns ImmutableHashSet Empty when there are no tracked objects, but ImmutableHashSet Empty uses the default equality comparer, not ReferenceEqualityComparer.Instance.

While this is fine for the empty case (no comparisons will occur), it creates an inconsistency where sometimes the returned ISet uses reference equality and sometimes it uses default equality. This could be confusing and potentially lead to subtle bugs if the code is extended later.

**Options:**
- Accept the inconsistency (performance benefit likely outweighs the risk for empty sets)
- Use ImmutableHashSet.Create with ReferenceEqualityComparer.Instance instead for consistency
- Add a comment explaining why the different comparer is acceptable for the empty case

### 2. Minor: Missing using directive location (TUnit.Core/Tracking/ObjectTracker.cs:2)

The new using System.Collections.Immutable is added after the existing usings. While this works, the existing code appears to keep System.* usings together at the top. Consider keeping consistent ordering.

## Verdict

The performance optimizations are sound and address legitimate allocations in the hot path. The early returns are properly placed and the async state machine deferral is a good optimization pattern.

thomhurst:
## Summary
Performance optimization to skip creating enumerators, HashSets, and async state machines in ObjectTracker when there are no objects to track.

## Critical Issues
None found ✅

## Suggestions

### Performance Enhancement Opportunity
The change at line 58-61 uses `ImmutableHashSet<object>.Empty` for the empty case. While this avoids allocating a HashSet, there's a subtle performance consideration:

**Current implementation:**
```csharp
if (trackedObjects.IsEmpty)
{
    return ImmutableHashSet<object>.Empty;
}
```

**Consideration:** `ImmutableHashSet<object>.Empty` doesn't use `ReferenceEqualityComparer`, while the non-empty path creates a HashSet with `ReferenceEqualityComparer.Instance`. This is actually fine because an empty set's comparer never gets used, but for consistency and to avoid any potential confusion, you could consider:

```csharp
// Alternative: static readonly field
private static readonly ISet<object> EmptySet = ImmutableHashSet<object>.Empty;
```

However, the current implementation is perfectly correct - the comparer difference doesn't matter for an empty collection, and `ImmutableHashSet<object>.Empty` is a singleton so there's no allocation. This is just a note about the comparer difference, not a blocking issue.

### Code Quality
The split between `UntrackObjects` (sync) and `UntrackObjectsAsync` (async) at lines 127-152 is excellent - this is exactly the right pattern for avoiding ValueTask state machines on the happy path (early return). Well done following TUnit's "Performance First" principle!

## TUnit Rules Compliance
✅ **Performance First** - Excellent adherence. Multiple allocation reductions:
  - Skips HashSet allocation when empty (using ImmutableHashSet.Empty singleton)
  - Avoids async state machine when returning early
  - Early exits prevent unnecessary enumeration

✅ **No VSTest** - Not applicable (no test platform changes)

✅ **AOT Compatible** - No reflection changes

✅ **Snapshot Testing** - Not applicable (no source generator or public API changes)

✅ **Dual-Mode** - Not applicable (no source-gen/reflection dual implementation)

## Verdict
✅ **APPROVE** - No critical issues. Solid performance improvements aligned with TUnit's performance-first philosophy.

PR #4231, opened 2026-01-04
perf: use `ValueStringBuilder` to avoid allocations

TimothyMakkison:
Uses `ValueListBuilder` and `ValueStringBuilder` to drastically reduce allocations in `TestIdentifierService`

- Note that `ValueListBuilder` has a lot of potential in other areas of `TUnit`; I see a lot of temporary Lists in use.
- `typeSb` could be a `VLS` but it's used so infrequently I don't see any point - Happy to change this.
    - Edit switched this to use `VLS` as it may see some use if someone uses generic tests. 
- removes a lot of intermediary `string` allocations and `ArrayPool` usage

### Bef ore
image

### After

image


thomhurst:
## Summary
Optimizes TestIdentifierService by replacing StringBuilder and ArrayPool with ValueStringBuilder and ValueListBuilder to reduce heap allocations during test ID generation.

## Critical Issues
None found ✅

## Suggestions

1. **Consider stack allocation size for generic type names** (TUnit.Engine/Services/TestIdentifierService.cs:108)
   - The `typeVsb` uses `stackalloc char[128]` which may be insufficient for deeply nested generic types with long namespace names
   - Suggestion: Consider increasing to 256 chars or add a comment about the size choice

2. **Minor: ValueListBuilder initial capacity** (TUnit.Engine/Services/TestIdentifierService.cs:91)
   - The initial capacity `[null, null, null, null]` (4 elements) may be too small for deeply nested types
   - However, this is likely rare in practice and the builder will grow automatically, so this is very minor

## Verdict

This is an excellent performance optimization that follows TUnit's Rule 4 (Performance First). The implementation:
- Eliminates heap allocations in a hot path (test discovery)
- Properly disposes of ref structs
- Maintains the same logic while improving efficiency
- Is AOT-compatible

The benchmarks show significant improvement (124 B → 56 B allocated). Great work!

thomhurst:
This seems to have broken execution for a bunch of tests. I think using a custom collection is quite risky too. Maybe we should leave this one.

TimothyMakkison:
> This seems to have broken execution for a bunch of tests.

What are the errors? I wouldn't be surprised if I messed up the logic without any tests: 😊 

> I think using a custom collection is quite risky too. Maybe we should leave this one.

Fair enough, I had a lot of plans to use these but I can adapt. For what it's worth the collections are both internally used by Microsoft in the .Net runtime and IMO are easier to use than `ArrayPools`

thomhurst:
How come they don't add it into the core library?

I haven't investigated properly, but it's expected things like 84 tests in a run and only getting 13. Click on the Ubuntu pipeline failures on the pr checks

TimothyMakkison:
The failed test are mostly `TUnit.Engine` so I probably messed up some logic when rewriting this method. Going to be a pain debugging this without local tests 😅 Not sure why `TUnit.TestProject.Fail4` is failing.

> How come they don't add it into the core library?

There is a [proposal ](https://github.com/dotnet/runtime/issues/25587) for this but it's blocked until they add a [safety tool](https://github.com/dotnet/runtime/issues/50389). They are worried that people will accidentally misuse it by; copying the `ValueStringBuilder` (instead of passing by `ref`), and then accidentally using an array that was returned to the `ArrayPool`. Andrew lock has a better explanation on his [blog.](https://andrewlock.net/a-deep-dive-on-stringbuilder-part-6-vaulestringbuilder-a-stack-based-string-builder/#so-can-i-actually-use-valuestringbuilder-)

thomhurst:
## Summary
This PR optimizes TestIdentifierService to reduce allocations using ValueStringBuilder and ValueListBuilder, improving performance.

## Critical Issues

**1. Code Duplication: ValueStringBuilder already exists**

The PR adds ValueStringBuilder to TUnit.Engine/Helpers/, but this type already exists at TUnit.Core/Helpers/ValueStringBuilder.cs (line 9). Since TUnit.Engine.csproj already references TUnit.Core (line 34 of TUnit.Engine.csproj), you can import and use the existing type instead of duplicating the code.

Recommendation:
- Remove TUnit.Engine/Helpers/ValueStringBuilder.cs
- Add using TUnit.Core.Helpers; to TestIdentifierService.cs
- Consider moving ValueListBuilder to TUnit.Core/Helpers/ as well for reusability (the PR description mentions it has potential in other areas of TUnit)

**2. Potential Bug in WriteTypeNameWithGenerics**

In WriteTypeNameWithGenerics (TestIdentifierService.cs:87), there is a ValueListBuilder initialized with a stack-allocated array of 4 nulls. This may cause the builder to start with Length=4 instead of Length=0, potentially adding null strings to the output when iterating backwards at line 139.

## Suggestions

1. Consider DynamicallyAccessedMembers annotation for reflection usage (line 113) per TUnit Rule 5
2. Good job disposing the builders correctly
3. Performance validation looks good based on the PR description

## Verdict
REQUEST CHANGES - Critical code duplication and potential bug in ValueListBuilder initialization

TimothyMakkison:
Should have fixed the issue, on line 138 I looped the list in reverse but didn't update line 140 to reflect this.


Note that it might be possible to remove the allocation on line 124 by changing `var typeHierarchy = new ValueListBuilder<Range>` where it keeps track of strings in `typeVSb`.


Edit: I think there is a second bug 😭

thomhurst:
## Summary
Uses ValueStringBuilder and ValueListBuilder to drastically reduce allocations in TestIdentifierService, replacing ArrayPool and StringBuilder usage.

## Critical Issues

### 1. Missing Dispose calls - Memory leak ❌
The PR creates multiple `ValueStringBuilder` and `ValueListBuilder` instances but doesn't dispose them properly in several places:

**In `WriteTypeNameWithGenerics` (TUnit.Engine/Services/TestIdentifierService.cs:80-145)**:
- `typeVsb` is only disposed at the end, but if an exception occurs during the loop, it won't be disposed
- `typeHierarchy` is disposed, which is good

**Recommendation**: Wrap in try-finally blocks or ensure disposal happens even on exceptions.

### 2. Critical bug in reversal logic ❌
**In `WriteTypeNameWithGenerics` (TUnit.Engine/Services/TestIdentifierService.cs:134-141)**:

The original code:
```csharp
typeHierarchy.Reverse();
for (var i = 0; i < typeHierarchy.Count; i++)
{
    if (i > 0) sb.Append('+');
    sb.Append(typeHierarchy[i]);
}
```

The new code:
```csharp
// Reverse to get outer-to-inner order
// Append all types with + separator (matching .NET Type.FullName convention for nested types)
for (var i = typeHierarchy.Length - 1; i >= 0; i--)
{
    if (i < typeHierarchy.Length - 1) vsb.Append('+');
    vsb.Append(typeHierarchy[i]);
}
```

**Problem**: The separator logic is wrong. The condition `if (i < typeHierarchy.Length - 1)` will add '+' **before** every element except the last one (when traversing backwards), which means '+' will be added **after** the first element in the final output. This reverses the separator placement.

**Expected**: `OuterType+InnerType`
**Actual**: `InnerType+OuterType` (incorrect)

The condition should be `if (i != typeHierarchy.Length - 1)` to match the original behavior.

### 3. Incorrect disposal of `typeVsb` ❌
**In `WriteTypeNameWithGenerics` (TUnit.Engine/Services/TestIdentifierService.cs:123)**:

```csharp
typeHierarchy.Append(typeVsb.AsSpan().ToString());
typeVsb.Length = 0;  // This is NOT disposal
```

Setting `Length = 0` doesn't return the pooled array. You must call `typeVsb.Dispose()` and then recreate it if you need to reuse it. Better approach: create `typeVsb` inside the loop scope.

### 4. Non-idiomatic ValueStringBuilder initialization
**In multiple places**: 
```csharp
var vsb = new ValueStringBuilder(stackalloc char[256]);
```

This creates the builder but the `ToString()` method calls `Dispose()` internally (line 85 in ValueStringBuilder.cs). However, if any exception occurs before `ToString()`, the stackalloc buffer won't be cleaned up. While stackalloc is safe since it's on the stack, if the builder rented from ArrayPool due to growth, those arrays won't be returned.

**Recommendation**: Always use try-finally or using pattern for ValueStringBuilder when it might grow beyond initial capacity.

## Suggestions

### Good: ValueStringBuilder bug fix ✅
The change to `AppendSpanFormattable` in ValueStringBuilder.cs:156 is actually a **bug fix** in the existing code:
```csharp
-if (value.TryFormat(_chars, out int charsWritten, format: default, provider: null))
+if (value.TryFormat(_chars[_pos..], out int charsWritten, format: default, provider: null))
```
This correctly passes the remaining buffer instead of the entire buffer. Good catch!

### Consider: Pre-calculating capacity
For `WriteTypeWithParameters`, you could pre-calculate the required capacity to avoid growth:
```csharp
// Estimate: '(' + params + ', ' * (n-1) + ')'
int estimatedLength = 2 + parameterTypes.Length * 20; // rough estimate
```

## Verdict
⚠️ **REQUEST CHANGES** - Critical bugs in disposal and reversal logic that will cause incorrect test IDs and potential memory leaks

TimothyMakkison:
Okay, should be good now.

> 2. Critical bug in reversal logic

I think the ai has missed that the original code calles `List.Reverse()` whereas I'm manually iterating in reverse. I don't think this is a bug and it's just confused, WDYT?

> The change to AppendSpanFormattable in ValueStringBuilder.cs:156 is actually a bug fix in the existing code:

Yup, glad the tests were failing, I'm really annoyed I missed it

Should I wrap everything in `try finally` to be safe? Failiure to [return an object isn't a massive issue](https://learn.microsoft.com/en-us/dotnet/api/system.buffers.arraypool-1.rent?view=net-10.0#remarks) and unlikely to happen as the VSB would have to exceeed 256 items and have an exception thrown.

thomhurst:
> Should I wrap everything in try finally to be safe?

If its IDisposable can we just use `using` when newing it up?

TimothyMakkison:
> > Should I wrap everything in try finally to be safe?

It's probably not necessary, but I've added `try finally`. Microsoft internally bother when they use, but their code is very battle hardened and reviewed by several people. Failiure to return an array causes it to be eventually garbage collected.

> If its IDisposable can we just use `using` when newing it up?

Can't use `using` on a value passed by `ref` into methods 🤷

PR #4273, opened 2026-01-09
perf: replace array and where clause with `ValueStringBuilder`

TimothyMakkison:
Removes `string[]` and `WhereArrayIterator` allocations by using `ValueStringBuilder`

### Allocations that are eliminated
image
image

### DIFF
--- a/TUnit.Core/Helpers/ValueStringBuilder.cs
+++ b/TUnit.Core/Helpers/ValueStringBuilder.cs
@@ -1,0 +1,301 @@
+﻿using System.Buffers;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace TUnit.Core.Helpers;
+
+// From https://raw.githubusercontent.com/dotnet/runtime/refs/heads/main/src/libraries/Common/src/System/Text/ValueStringBuilder.cs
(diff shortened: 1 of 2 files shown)


thomhurst:
## Summary
Eliminates array and LINQ allocations in test output methods by using stack-based `ValueStringBuilder`.

## Critical Issues
1. **Missing `static` modifier** (`TestContext.Output.cs:107`): The `AppendIfNotNullOrEmpty` method should be declared as `static` since it doesn't access instance state:
   ```csharp
   static void AppendIfNotNullOrEmpty(ref ValueStringBuilder builder, string? value)
   ```

## Suggestions
None - the performance optimization is well-targeted and the `ValueStringBuilder` implementation from dotnet/runtime is battle-tested.

## TUnit Rules Compliance
✅ **Performance First** - Directly addresses this rule by eliminating hot-path allocations  
✅ **AOT Compatible** - Ref struct with no reflection concerns  
✅ **Modern C#** - Uses modern span-based APIs

## Verdict
⚠️ **REQUEST CHANGES** - One minor but required fix (missing `static` modifier)

TimothyMakkison:
> Missing static modifier

Fixed 👍 

Hopefully this PR proves that I messed up the logic on #4231. I'll give it another try if this PR passes the CI

PR #4279, opened 2026-01-10
perf: reduce allocations with `Span.Split` and `VSB`

TimothyMakkison:
Use `ValueStringBuilder` and `Span.Split` to avoid allocating `string[]` and `string.Substring`

Note that while the savings are small for modern versions of .NET - 2.5MB, this represents a substantial speedup for .NET framework where `string.Split` would allocate 60MB of `int[]`

### Before
image

### After
image

### DIFF
--- a/TUnit.Engine/Exceptions/TUnitFailedException.cs
+++ b/TUnit.Engine/Exceptions/TUnitFailedException.cs
@@ -1,4 +1,5 @@
 ﻿using TUnit.Core.Exceptions;
+using TUnit.Core.Helpers;
 
 namespace TUnit.Engine.Exceptions;
 
@@ -23,9 +24,21 @@ public abstract class TUnitFailedException : TUnitException
             return string.Empty;
         }
 
-        var lines = stackTrace!.Split([Environment.NewLine], StringSplitOptions.None);
+        var vsb = new ValueStringBuilder(stackalloc char[256]);
 
-        return string.Join(Environment.NewLine,
-            lines.TakeWhile(x => !x.Trim().StartsWith("at TUnit")));
+        var added = false;
+        foreach(var range in stackTrace.AsSpan().Split(Environment.NewLine))
+        {
+            var slice = stackTrace.AsSpan()[range];
+            if (slice.Trim().StartsWith("at TUnit"))
+            {
+                break;
+            }
+            vsb.Append(added ? Environment.NewLine : "");
+            vsb.Append(slice);
+            added = true;
+        }
+
+        return vsb.ToString();
     }
 }


thomhurst:
## Summary
Optimizes stack trace filtering to reduce allocations using `ValueStringBuilder` and `Span.Split` instead of `string.Split` and LINQ.

## Critical Issues
None found ✅

## Suggestions

**Minor optimization opportunity:**
The current implementation appends `Environment.NewLine` conditionally on each iteration. You could simplify this slightly by appending the newline **after** the slice (except for the last element), avoiding the `added` boolean:

```csharp
var vsb = new ValueStringBuilder(stackalloc char[256]);
var enumerator = stackTrace.AsSpan().Split(Environment.NewLine);
var isFirst = true;

foreach (var range in enumerator)
{
    var slice = stackTrace.AsSpan()[range];
    if (slice.Trim().StartsWith("at TUnit"))
    {
        break;
    }
    
    if (!isFirst)
    {
        vsb.Append(Environment.NewLine);
    }
    
    vsb.Append(slice);
    isFirst = false;
}
```

However, the current implementation is clear and correct, so this is entirely optional.

## Alignment with TUnit Rules

✅ **Performance First (Rule 4):** This change directly aligns with TUnit's performance-first principle. Exception handling is in the execution hot path (every test failure creates a `TUnitFailedException`), making allocation reduction valuable.

✅ **AOT Compatible (Rule 5):** Uses span-based APIs compatible with Native AOT.

✅ **Modern C# (Code Principle):** Leverages latest .NET features (Span APIs, stackalloc, ValueStringBuilder).

The Polyfill package ensures `MemoryExtensions.Split` is available on netstandard2.0, so this will work across all target frameworks.

## Verdict
✅ **APPROVE** - Great performance improvement with correct implementation!

thomhurst:
How does the stackalloc work if there's more than 256 chars?

TimothyMakkison:
> How does the stackalloc work if there's more than 256 chars?

We initialise `ValueStringBuilder` with a starting buffer, in this case we used `stackalloc` to create an allocation free initial buffer, however any compatible collection would be a valid starting buffer. If we need more than 256 characters; `void Grow(int additionalCapacityBeyondPos)` is called which copies the current buffer into a new array rented from `ArrayPool`, returning any previously rented arrays. Finally, `Dispose` or `ToString` will return any rented arrays when we are done with the `ValueStringBuilder`. 

Just like any other value on the stack, the `stackalloc` buffer will be dissappear when the stack frame pops at the end of the functions scope.

This is true for `ValueListBuilder` as well.

PR #4280, opened 2026-01-10
perf: replace `SubsString` with alternative lookup

TimothyMakkison:
Use `GetAlternateLookup` to avoid allocaing a `string` via `SubString`. 

Edit: use preprocessor directive, apparently you can't polyfill `GetAlternativeLookup`

Small improvement saves around 200KB

### DIFF
--- a/TUnit.Core/Discovery/ObjectGraphDiscoverer.cs
+++ b/TUnit.Core/Discovery/ObjectGraphDiscoverer.cs
@@ -651,8 +651,13 @@ internal sealed class ObjectGraphDiscoverer : IObjectGraphTracker
             return true; // Malformed key, treat as direct
         }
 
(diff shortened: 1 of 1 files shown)


thomhurst:
## Summary
Replaces Substring allocation with GetAlternateLookup on .NET 9+ in test discovery hot path, saving approximately 200KB.

## Critical Issues
None found - approved

## Suggestions
None - this is a well-executed performance optimization that targets a hot path per Rule 4, uses preprocessor directive for backward compatibility, maintains semantic equivalence with zero-allocation lookup, and doesn't require snapshot updates.

The use of GetAlternateLookup with ReadOnlySpan is the correct modern approach for span-based HashSet lookups in .NET 9+.

## Verdict
APPROVE - Clean performance win with no downsides

PR #4292, opened 2026-01-10
perf: cache CreateExecutableTestFactory delegate

### LINKED ISSUE #4281: perf: CreateExecutableTestFactory creates new lambdas on every property access
## Summary
The `CreateExecutableTestFactory` property getter in `TestMetadata<T>` creates new lambda closures every time it is accessed, causing unnecessary heap allocations and GC pressure.

## Evidence
From dotnet-trace profiling of the TUnit.PerformanceBenchmarks project (3700+ tests):
- `<get_CreateExecutableTestFactory>b__2` shows **5.41% inclusive** CPU time
- `<set_InstanceFactory>b__0` shows **2.2% inclusive** CPU time

## Root Cause
In `TUnit.Core/TestMetadata\`1.cs` lines 62-116, the property getter creates new lambdas on every access:

```csharp
public override Func<ExecutableTestCreationContext, TestMetadata, AbstractExecutableTest> CreateExecutableTestFactory
{
    get
    {
        if (InstanceFactory != null && InvokeTypedTest != null)
        {
            return (context, metadata) =>  // NEW LAMBDA EVERY ACCESS!
            {
                var typedMetadata = (TestMetadata<T>)metadata;
                Func<TestContext, Task<object>> createInstance = async testContext => { ... };
                Func<object, object?[], TestContext, CancellationToken, Task> invokeTest = async (...) => { ... };
                return new ExecutableTest(createInstance, invokeTest) { ... };
            };
        }
        throw new InvalidOperationException(...);
    }
}
```

## Suggested Fix
Cache the delegate after first creation:

```csharp
private Func<ExecutableTestCreationContext, TestMetadata, AbstractExecutableTest>? _cachedExecutableTestFactory;

public override Func<ExecutableTestCreationContext, TestMetadata, AbstractExecutableTest> CreateExecutableTestFactory
{
    get
    {
        if (_cachedExecutableTestFactory != null)
            return _cachedExecutableTestFactory;

        if (InstanceFactory != null && InvokeTypedTest != null)
        {
            _cachedExecutableTestFactory = (context, metadata) => { ... };
            return _cachedExecutableTestFactory;
        }
        throw new InvalidOperationException(...);
    }
}
```

## Impact
- Reduces heap allocations per test by eliminating redundant lambda creation
- Expected improvement: Noticeable reduction in GC pressure for large test suites
- Affects every test execution path

## Related Files
- `TUnit.Core/TestMetadata\`1.cs`

thomhurst:
## Summary
- Caches the factory delegate in `TestMetadata<T>.CreateExecutableTestFactory` instead of creating new lambdas on every property access
- Avoids repeated delegate allocations during test discovery and execution

Fixes #4281

## Test plan
- [x] Unit tests pass (94/94)
- [ ] Benchmarks show reduced allocations during test execution

### DIFF
--- a/TUnit.Core/TestMetadata`1.cs
+++ b/TUnit.Core/TestMetadata`1.cs
@@ -14,6 +14,7 @@ public class TestMetadata<
 {
     private Func<Type[], object?[], T>? _instanceFactory;
     private Func<T, object?[], Task>? _testInvoker;
+    private Func<ExecutableTestCreationContext, TestMetadata, AbstractExecutableTest>? _cachedExecutableTestFactory;
 
     /// <summary>
     /// Strongly typed instance factory
@@ -63,10 +64,16 @@ public class TestMetadata<
     {
         get
         {
+            // Return cached factory if available
+            if (_cachedExecutableTestFactory != null)
+            {
+                return _cachedExecutableTestFactory;
+            }
+
             // For AOT mode, create delegates from the strongly-typed ones
             if (InstanceFactory != null && InvokeTypedTest != null)
             {
-                return (context, metadata) =>
+                _cachedExecutableTestFactory = (context, metadata) =>
                 {
                     var typedMetadata = (TestMetadata<T>)metadata;
 
@@ -109,6 +116,8 @@ public class TestMetadata<
                         Context = context.Context
                     };
                 };
+
+                return _cachedExecutableTestFactory;
             }
 
             throw new InvalidOperationException($"InstanceFactory and InvokeTypedTest must be set for {typeof(T).Name}");


thomhurst:
## Summary
Caches the `CreateExecutableTestFactory` delegate to avoid repeated allocations during test discovery and execution.

## Critical Issues
None found ✅

## Suggestions

### Thread Safety (Minor)
The caching pattern in `TestMetadata<T>.cs:66-72` has a benign race condition - multiple threads accessing `CreateExecutableTestFactory` concurrently could create multiple delegate instances before caching. However, this is acceptable because:

1. The reflection mode (`ReflectionTestMetadata.cs:38`) uses `field ??=` which has the same race condition
2. The race is benign - all created delegates are functionally identical
3. Worst case is a few extra allocations during initialization, which is negligible compared to the allocations saved

If you want to eliminate the race entirely, consider using `Lazy<T>` or `LazyInitializer.EnsureInitialized`, but given the benign nature and consistency with the reflection mode, the current implementation is acceptable.

### .gitignore Change
The `.gitignore` change adding `.worktrees/` is unrelated to the performance fix but is a reasonable developer experience improvement. Consider splitting unrelated changes into separate PRs in the future for clearer history.

## Verdict
✅ **APPROVE** - Performance improvement with dual-mode consistency. The thread safety consideration is minor and acceptable given the benign race and consistency with existing reflection mode implementation.

PR #4304, opened 2026-01-11
perf: use `vlb` in `FilterScopedAttributes` and defer creation.

TimothyMakkison:
Small improvement, saves around 2 MB in `ScopedAttributeFilter.FilterScopedAttributes` by deferring `Dictionary` creation and using `ValueListBuilder`

### Before
image

### After

image

### DIFF
--- a/TUnit.Engine/Utilities/ScopedAttributeFilter.cs
+++ b/TUnit.Engine/Utilities/ScopedAttributeFilter.cs
@@ -1,7 +1,5 @@
-using System.Diagnostics.CodeAnalysis;
-using Polyfills;
 using TUnit.Core;
-using TUnit.Engine.Helpers;
+using TUnit.Core.Helpers;
 
 namespace TUnit.Engine.Utilities;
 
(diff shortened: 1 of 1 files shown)


thomhurst:
## Summary
Performance optimization that uses ValueListBuilder and defers Dictionary allocation in ScopedAttributeFilter, saving ~2 MB.

## Critical Issues
None found ✅

The changes look good:
- Proper use of ValueListBuilder with stack allocation
- Dictionary deferred until scoped attributes are actually encountered
- Return type change from List<T> to T[] is safe (internal API, all callers just enumerate)
- Dispose() correctly called on ValueListBuilder
- Aligns with Rule 4 (Performance First) - minimizes allocations in hot path

## Suggestions

**Minor: Stack allocation size**
The initial stack buffer is `[null,null,null,null]` (4 elements). Consider using `stackalloc T[8]` instead for a more typical initial capacity that matches common scenarios, though 4 is fine if profiling shows it's sufficient.

**Note: OrderBy allocation**
One call site (EventReceiverOrchestrator.cs:244) uses `.OrderBy()` which allocates anyway. This isn't a blocker, but if optimizing that path later, consider sorting the array in-place or using ValueListBuilder's ability to insert in sorted order.

## Verdict
✅ **APPROVE** - Solid performance improvement with no critical issues

PR #4312, opened 2026-01-12
perf: convert synchronous TestStateManager methods to void

thomhurst:
## Summary

- Convert TestStateManager methods from `Task`-returning to `void` since they are purely synchronous in-memory operations
- Make `InProgress` message fire-and-forget in TestCoordinator since it's informational and doesn't need to block test execution

### Changes

| Old Method | New Method |
|------------|------------|
| `MarkRunningAsync` | `MarkRunning` |
| `MarkCompletedAsync` | `MarkCompleted` |
| `MarkFailedAsync` | `MarkFailed` |
| `MarkSkippedAsync` | `MarkSkipped` |
| `MarkCircularDependencyFailedAsync` | `MarkCircularDependencyFailed` |
| `MarkDependencyResolutionFailedAsync` | `MarkDependencyResolutionFailed` |

### Benefits

- Eliminates async state machine overhead at call sites
- Zero allocation for synchronous operations (previously used `Task.CompletedTask`)
- Cleaner API that accurately represents the synchronous nature of these operations

## Test plan

- [x] All 1013 profile tests pass
- [x] No performance regression (~251-263ms same as baseline)
- [x] Build succeeds for all target frameworks

### DIFF
--- a/TUnit.Engine/Scheduling/TestRunner.cs
+++ b/TUnit.Engine/Scheduling/TestRunner.cs
@@ -77,7 +77,7 @@ public sealed class TestRunner
 
                 if (dependency.Test.State == TestState.Failed && !dependency.ProceedOnFailure)
                 {
-                    await _testStateManager.MarkSkippedAsync(test, "Skipped due to failed dependencies").ConfigureAwait(false);
+                    _testStateManager.MarkSkipped(test, "Skipped due to failed dependencies");
                     await _tunitMessageBus.Skipped(test.Context, "Skipped due to failed dependencies").ConfigureAwait(false);
                     return;
                 }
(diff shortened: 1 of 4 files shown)


thomhurst:
## Summary
Converts synchronous TestStateManager methods from Task-returning to void to eliminate async overhead.

## Critical Issues

Fire-and-forget pattern introduces potential issues (TestCoordinator.cs:65)

The change to fire-and-forget for InProgress messages introduces several problems:

1. Silent exception swallowing: If InProgress throws, the exception will be unobserved
2. Race conditions: InProgress message may not complete before subsequent state changes, causing out-of-order test state reporting
3. Inconsistent pattern: All other message bus calls are awaited

Recommendation: Keep the await for InProgress. The performance gain is negligible compared to the reliability risk.

## Suggestions

The TestStateManager void conversion itself is good - these are genuinely synchronous operations.

## Verdict
REQUEST CHANGES - Fire-and-forget pattern needs to be addressed

PR #4322, opened 2026-01-12
fix: handle IgnoringType for types nested in ValueTypes

### LINKED ISSUE #4320: [Bug]: IgnoringType is ignored if type is in a ValueType
### Description

When a type is stored inside a ValueType, like a tuple, it is *not* ignored when checking structural equivalency (`IsEquivalentTo`), when it has explicitly been excluded via `IgnoringType`.

This has been caused by https://github.com/thomhurst/TUnit/pull/4263

### Expected Behavior

Types that are specified in `IgnoringType` are ignored, even when they are stored in a ValueType.

### Actual Behavior

Equivalency for these types is checked and tests unexpectedly fail.

### Steps to Reproduce

See my reproducer at https://github.com/davidmehren/TUnitIgnoreTypeRepro/blob/main/Tests.cs

I would expect both tests to succeed.

### TUnit Version

1.9.85

### .NET Version

.NET 10.0.101

### Operating System

Linux

### IDE / Test Runner

```shell
████████╗██╗   ██╗███╗   ██╗██╗████████╗
╚══██╔══╝██║   ██║████╗  ██║██║╚══██╔══╝
   ██║   ██║   ██║██╔██╗ ██║██║   ██║
   ██║   ██║   ██║██║╚██╗██║██║   ██║
   ██║   ╚██████╔╝██║ ╚████║██║   ██║
   ╚═╝    ╚═════╝ ╚═╝  ╚═══╝╚═╝   ╚═╝

   TUnit v1.9.85.0 | 64-bit | Fedora Linux 43 (Workstation Edition) | linux-x64 | .NET 10.0.1 | Microsoft Testing Platform v2.0.2

   Engine Mode: SourceGenerated

fehlerhaft Fails (18ms)
  TUnit.Engine.Exceptions.TestFailedException: AssertionException: Expected actual to be equivalent to expected
  but Property Ignores did not match
  Expected: (IgnoreMe { Message = baz }, IgnoreMe { Message = baz })
  Received: (IgnoreMe { Message = foobar }, IgnoreMe { Message = foobar })
  
  at Assert.That(actual).IsEquivalentTo(expected).IgnoringType<IgnoreMe>()

Testlaufzusammenfassung: Fehler! - TUnitIgnoreTypeRepro.dll (net10.0|x64)
  gesamt: 2
  fehlerhaft: 1
  erfolgreich: 1
  übersprungen: 0
  Dauer: 267ms


Process finished with exit code 2.
```

### Additional Context

_No response_

### IDE-Specific Issue?

thomhurst:
## Summary

- Fix `IgnoringType` being bypassed when the ignored type is nested inside a `ValueTuple` or other value type implementing `IEquatable<T>`
- Add `ContainsIgnoredGenericArgument()` helper method that recursively checks if a generic type contains ignored types
- Add comprehensive tests for tuple scenarios

## Problem

When a type is stored inside a `ValueTuple` (like `(IgnoreMe, IgnoreMe)`), the tuple is treated as a "primitive" type by `TypeHelper.IsPrimitiveOrWellKnownType()` because it implements `IEquatable<T>`. This causes the comparison to use `Equals()` directly instead of recursively comparing members, bypassing the `ShouldIgnoreType` check entirely.

## Solution

Before treating a value type as primitive, check if it's a generic type containing any ignored types as generic arguments. If so, fall through to structural comparison which properly handles the ignored types.

## Test plan

- [x] Added test: `IgnoringType_In_Tuple_Properties_Are_Ignored` - Basic tuple case
- [x] Added test: `IgnoringType_In_Nested_Tuple_Properties_Are_Ignored` - Nested tuples  
- [x] Added test: `IgnoringType_In_Mixed_Tuple_Still_Compares_Non_Ignored_Parts` - Verifies non-ignored parts are still compared
- [x] Added test: `IgnoringType_In_Mixed_Tuple_Passes_When_NonIgnored_Parts_Match` - Mixed tuple with matching non-ignored values
- [x] All existing `IgnoringTypeEquivalentTests` pass (12 tests)
- [x] All `CollectionStructuralEquivalenceTests` pass (20 tests)
- [x] All `EquivalentAssertionTests` pass (29 tests)

Fixes #4320

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/TUnit.Assertions.Tests/IgnoringTypeEquivalentTests.cs
+++ b/TUnit.Assertions.Tests/IgnoringTypeEquivalentTests.cs
@@ -227,4 +227,124 @@ public class IgnoringTypeEquivalentTests
         public string Name { get; set; } = string.Empty;
         public int Value { get; set; }
     }
+
+    // Test classes for ValueType/Tuple tests
+    private class IgnoreMe
+    {
+        public string Message { get; set; } = string.Empty;
+
+        public IgnoreMe() { }
+        public IgnoreMe(string message) => Message = message;
+    }
+
+    private class ClassWithTupleProperty
+    {
+        public string Name { get; set; } = string.Empty;
+        public (IgnoreMe, IgnoreMe) Ignores { get; set; }
+        public int Value { get; set; }
+    }
+
+    private class ClassWithNestedTupleProperty
+    {
+        public string Name { get; set; } = string.Empty;
+        public ((IgnoreMe, int), string) NestedIgnores { get; set; }
+        public int Value { get; set; }
+    }
+
+    private class ClassWithMixedTupleProperty
+    {
+        public string Name { get; set; } = string.Empty;
+        public (IgnoreMe, int) MixedTuple { get; set; }
+        public int Value { get; set; }
+    }
+
+    [Test]
+    public async Task IgnoringType_In_Tuple_Properties_Are_Ignored()
+    {
+        var object1 = new ClassWithTupleProperty
+        {
+            Name = "Test",
+            Ignores = (new IgnoreMe("foobar"), new IgnoreMe("foobar")),
+            Value = 123
+        };
+
+        var object2 = new ClassWithTupleProperty
+        {
+            Name = "Test",
+            Ignores = (new IgnoreMe("baz"), new IgnoreMe("baz")),
+            Value = 123
+        };
+
+        await TUnitAssert.That(object1)
+            .IsEquivalentTo(object2)
+            .IgnoringType<IgnoreMe>();
+    }
+
+    [Test]
+    public async Task IgnoringType_In_Nested_Tuple_Properties_Are_Ignored()
+    {
+        var object1 = new ClassWithNestedTupleProperty
+        {
+            Name = "Test",
+            NestedIgnores = ((new IgnoreMe("foobar"), 1), "hello"),
+            Value = 123
+        };
+
+        var object2 = new ClassWithNestedTupleProperty
+        {
+            Name = "Test",
+            NestedIgnores = ((new IgnoreMe("baz"), 1), "hello"),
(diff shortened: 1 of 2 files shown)


thomhurst:
## Summary
Fixes a bug where `IgnoringType` was bypassed when ignored types were nested inside ValueTuples, by checking if generic types contain ignored type arguments before treating them as primitives.

## Critical Issues
None found ✅

## Suggestions

### Performance: Hot path optimization
The new `ContainsIgnoredGenericArgument()` call at line 123 is in a hot path (called for every object comparison). Consider optimizing:

```csharp
// Current: Always calls the method
if (TypeHelper.IsPrimitiveOrWellKnownType(actualType) && !ContainsIgnoredGenericArgument(actualType))

// Suggested: Only check generic types
if (TypeHelper.IsPrimitiveOrWellKnownType(actualType) && 
    !(actualType.IsGenericType && _ignoredTypes.Count > 0 && ContainsIgnoredGenericArgument(actualType)))
```

This avoids the method call overhead for non-generic types and when no types are being ignored (the common case). The early exits are already in `ContainsIgnoredGenericArgument()`, but pulling them inline removes the call overhead.

Alternatively, consider caching results or only checking value types that implement IEquatable (since that's the only category from TypeHelper that would include tuples).

### Consider edge case: What about generic classes implementing IEquatable?
The fix specifically targets value types (which include ValueTuple), but could a similar issue occur with a generic class that implements `IEquatable<T>` and contains ignored types? Based on TypeHelper.cs:95-98, classes aren't treated as primitive even if they implement IEquatable, so this should be fine - but worth confirming this edge case is handled correctly.

## Verdict
✅ **APPROVE** - No critical issues. The fix correctly solves the reported bug with comprehensive tests. The performance suggestion is optional optimization.

PR #4257, opened 2026-01-07
fix: Duplicated output in "Detailed" mode

thomhurst:
Fixes #4256

PR #4327, opened 2026-01-12
fix: always include StandardOutputProperty in test nodes for IDE support

### LINKED ISSUE #4325: [Bug]: 1.9.42 broke test output
### Description

Since 1.9.42, I can't see Console.Write in the test output.

### Expected Behavior

Console.Write should redirected to the test output.

### Actual Behavior

Can't see the Console.Write anywhere.

### Steps to Reproduce

```
public class C
{
    [Test]
    public void M()
    {
        Console.WriteLine("Hi");
    }
}
```

### TUnit Version

1.9.85

### .NET Version

.NET 10.0

### Operating System

macOS

### IDE / Test Runner

JetBrains Rider

```shell

```

### Additional Context

_No response_

### IDE-Specific Issue?

thomhurst:
## Summary

- Fixes #4325 - Console.WriteLine output not visible in Rider since v1.9.42
- Always adds `StandardOutputProperty` and `StandardErrorProperty` to test nodes regardless of output mode
- Removes the unused `IsDetailedOutput` check that was preventing IDE test output display

## Root Cause

PR #4257 fixed duplicate output in "Detailed" mode by skipping `StandardOutputProperty` when `IsDetailedOutput` was true. However, IDEs like Rider are detected as "detailed" mode (because their client ID doesn't contain "console") but still need these properties to display test output in their results panel.

The real-time console output duplication is already handled separately by `HideTestOutput` in `OptimizedConsoleInterceptor`, so `StandardOutputProperty` should always be included for IDE consumption.

## Test plan

- [x] Build succeeds
- [x] `ConsoleTests` pass and show output in "Standard output" section
- [x] `CaptureOutputTests` pass with correct output capture
- [ ] Manual verification in Rider that Console.WriteLine output appears in test results

### DIFF
--- a/TUnit.Engine/Extensions/TestExtensions.cs
+++ b/TUnit.Engine/Extensions/TestExtensions.cs
@@ -6,7 +6,6 @@ using Microsoft.Testing.Platform.Extensions.Messages;
 using TUnit.Core;
 using TUnit.Core.Extensions;
 using TUnit.Engine.Capabilities;
-using TUnit.Engine.Services;
 #pragma warning disable TPEXP
 
 namespace TUnit.Engine.Extensions;
@@ -14,7 +13,6 @@ namespace TUnit.Engine.Extensions;
 internal static class TestExtensions
 {
     private static bool? _cachedIsTrxEnabled;
-    private static bool? _cachedIsDetailedOutput;
 
     private static readonly ConcurrentDictionary<Assembly, string> AssemblyFullNameCache = new();
     private static readonly ConcurrentDictionary<string, CachedTestNodeProperties> TestNodePropertiesCache = new();
@@ -34,7 +32,6 @@ internal static class TestExtensions
         AssemblyFullNameCache.Clear();
         TestNodePropertiesCache.Clear();
         _cachedIsTrxEnabled = null;
-        _cachedIsDetailedOutput = null;
     }
 
     private static string GetCachedAssemblyFullName(Assembly assembly)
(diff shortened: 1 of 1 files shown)


thomhurst:
## Summary
Fixes Console.WriteLine output not appearing in Rider by always including StandardOutputProperty and StandardErrorProperty in test nodes.

## Critical Issues
None found ✅

## Analysis

The fix correctly addresses the root cause identified in #4325. Here's why this change is correct:

**The Problem:**
- PR #4257 added a check to skip `StandardOutputProperty` when `IsDetailedOutput` was true to prevent duplicate console output
- However, IDEs like Rider are detected as "detailed" mode (via `VerbosityService.IsDetailedOutput` at line 58 - returns true when client ID doesn't contain "console")
- IDEs need these properties to display output in their test results panel, even though they're classified as "detailed"

**The Solution:**
- Real-time console output duplication is already handled by `HideTestOutput` in `OptimizedConsoleInterceptor.cs:66-70`, which checks `VerbosityService.HideTestOutput` (not `IsDetailedOutput`)
- The `IsDetailedOutput` check in `TestExtensions.cs:162` was redundant and breaking IDE output display
- The fix correctly removes this check and always includes the properties when output exists

**Code Quality:**
- Properly removes the now-unused `IsDetailedOutput()` method and `_cachedIsDetailedOutput` field
- Removes the unused `using TUnit.Engine.Services;` import
- Test coverage exists via `ConsoleTests` and `CaptureOutputTests` mentioned in the test plan

**TUnit Rules Compliance:**
- ✅ No source generator changes - no snapshot updates needed
- ✅ No public API changes
- ✅ No VSTest usage
- ✅ No performance regressions (actually removes unnecessary method calls)
- ✅ No reflection usage

## Verdict
✅ **APPROVE** - Clean fix that correctly addresses the root cause without side effects.

PR #4329, opened 2026-01-12
perf: `AssertAsync` return synchronous  `Task`

TimothyMakkison:
Make `AssertAsync` synchronous  return a `Task` without awaiting. 

- As usual this will mean stack traces thrown by the returned task won't include this method.

### Before
image

### After
No longer creating a state machine or allocating
image


thomhurst:
I'd like to keep AssertAsync in the stacktrace I think

TimothyMakkison:
Fair enough, I thought this was stripped out before the user saw the error?

thomhurst:
> Fair enough, I thought this was stripped out before the user saw the error?

The assertions library is technically standalone so may not be stripped out if used elsewhere. Also can keep the full stacktrace with --detailed-stacktrace 😄

PR #4346, opened 2026-01-12
fix: convert xUnit Assert.All to TUnit All(predicate) instead of AllSatisfy

### LINKED ISSUE #4335: xUnit Migration: Assert.All converted to non-existent AllSatisfy method
## Description

When using the TUnit xUnit migration code fixer (TUXU0001), the fixer converts xUnit's `Assert.All()` to `.AllSatisfy()`, but this method doesn't exist on TUnit's collection assertions, resulting in CS1061 compilation errors.

## Steps to Reproduce

1. Create an xUnit test using `Assert.All()`:
```csharp
using Xunit;

[Fact]
public void Test_All_Items()
{
    var items = new[] { 1, 2, 3 };
    Assert.All(items, item => Assert.True(item > 0));
}
```

2. Install TUnit package and run `dotnet format analyzers --severity info --diagnostics TUXU0001`

## Expected Behavior

The code should be converted to valid TUnit syntax. If `.AllSatisfy()` doesn't exist, either:
1. The method should be added to TUnit, or
2. The fixer should convert to an existing TUnit pattern

## Actual Behavior

The code is converted to:
```csharp
[Test]
public async Task Test_All_Items()
{
    var items = new[] { 1, 2, 3 };
    await Assert.That(items).AllSatisfy(item => Assert.True(item > 0));
}
```

This causes:
```
error CS1061: 'CollectionAssertion<int>' does not contain a definition for 'AllSatisfy' and no accessible extension method 'AllSatisfy' accepting a first argument of type 'CollectionAssertion<int>' could be found
```

## Affected Assertion Types

The issue affects multiple collection assertion types:
- `CollectionAssertion<T>`
- `ReadOnlyListAssertion<T>`

## Impact

This affects tests that validate all items in a collection satisfy a condition. In the Moq test suite, this resulted in 28+ compilation errors.

## Environment

- TUnit Version: 1.9.91
- .NET SDK: 10.0.101
- OS: Windows

thomhurst:
## Summary
- Fixes #4335
- xUnit `Assert.All(collection, action)` now converts to `await Assert.That(collection).All(predicate)` instead of the non-existent `AllSatisfy` method
- Added `TryConvertActionToPredicate` to extract predicates from common xUnit assertion patterns
- Also fixed MSTest migration to use `All()` instead of non-existent `AllSatisfy()`

## Test plan
- [x] Added test for xUnit Assert.All conversion
- [x] All xUnit migration tests pass (41 total)
- [x] All MSTest migration tests pass (23 total)

### DIFF
--- a/TUnit.Analyzers.CodeFixers/XUnitMigrationCodeFixProvider.cs
+++ b/TUnit.Analyzers.CodeFixers/XUnitMigrationCodeFixProvider.cs
@@ -569,9 +569,9 @@ public class XUnitMigrationCodeFixProvider : BaseMigrationCodeFixProvider
 
                 // Reference assertions
                 "Same" when arguments.Count >= 2 =>
-                    CreateTUnitAssertion("IsSameReference", arguments[1].Expression, arguments[0]),
+                    CreateTUnitAssertion("IsSameReferenceAs", arguments[1].Expression, arguments[0]),
                 "NotSame" when arguments.Count >= 2 =>
-                    CreateTUnitAssertion("IsNotSameReference", arguments[1].Expression, arguments[0]),
+                    CreateTUnitAssertion("IsNotSameReferenceAs", arguments[1].Expression, arguments[0]),
 
                 // String/Collection contains
                 "Contains" when arguments.Count >= 2 =>
--- a/TUnit.Analyzers.Tests/XUnitMigrationAnalyzerTests.cs
+++ b/TUnit.Analyzers.Tests/XUnitMigrationAnalyzerTests.cs
@@ -862,6 +862,82 @@ public class XUnitMigrationAnalyzerTests
     }
 
     [Test]
+    public async Task Assert_Same_Can_Be_Converted()
+    {
+        await CodeFixer
+            .VerifyCodeFixAsync(
+                """
+                {|#0:using TUnit.Core;
+
+                public class MyClass
+                {
+                    [Fact]
+                    public void MyTest()
+                    {
+                        var expected = new object();
+                        var actual = expected;
+                        Assert.Same(expected, actual);
+                    }
+                }|}
+                """,
+                Verifier.Diagnostic(Rules.XunitMigration).WithLocation(0),
+                """
+                using TUnit.Core;
+
+                public class MyClass
+                {
+                    [Test]
+                    public async Task MyTest()
+                    {
+                        var expected = new object();
+                        var actual = expected;
+                        await Assert.That(actual).IsSameReferenceAs(expected);
+                    }
+                }
+                """,
+                ConfigureXUnitTest
+            );
+    }
+
+    [Test]
+    public async Task Assert_NotSame_Can_Be_Converted()
+    {
+        await CodeFixer
+            .VerifyCodeFixAsync(
+                """
+                {|#0:using TUnit.Core;
+
+                public class MyClass
+                {
+                    [Fact]
+                    public void MyTest()
+                    {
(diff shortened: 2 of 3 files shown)


thomhurst:
## Summary
Fixes xUnit and MSTest migration to use the correct `All(predicate)` method instead of the non-existent `AllSatisfy` method.

## Critical Issues
None found ✅

## Suggestions
None - the implementation is clean and well-tested. The predicate extraction logic handles common xUnit assertion patterns (Assert.True, Assert.False, Assert.NotNull, Assert.Null) intelligently, and the fallback behavior for unrecognized patterns is appropriate.

## TUnit Rules Check
✅ No dual-mode concerns (not touching core engine or source generator)
✅ No snapshot test changes needed (analyzer uses different test approach)
✅ No public API changes
✅ No hot path performance concerns (code fixer, not runtime)
✅ No reflection or AOT compatibility issues
✅ Modern C# syntax used appropriately

## Verdict
✅ **APPROVE** - Clean bug fix with proper test coverage. The conversion logic correctly transforms both xUnit `Assert.All(collection, action)` and MSTest `CollectionAssert.AllItemsAreNotNull`/`AllItemsAreInstancesOfType` to use TUnit's `All(predicate)` API.

thomhurst:
## Summary
Fixes xUnit Assert.All and MSTest AllItemsAreNotNull/AllItemsAreInstancesOfType migration to use the correct TUnit `.All(predicate)` method instead of the non-existent `.AllSatisfy()` method.

## Critical Issues
None found ✅

## Suggestions

### 1. Consider edge cases in TryConvertActionToPredicate
The `TryConvertActionToPredicate` method handles common patterns like `Assert.True`, `Assert.False`, `Assert.NotNull`, and `Assert.Null`. However, there are other common xUnit assertion patterns that might appear in `Assert.All` calls:

- `Assert.Equal(expected, actual)` → `actual == expected`
- `Assert.NotEqual(expected, actual)` → `actual != expected`  
- `Assert.Empty(collection)` → `!collection.Any()`
- `Assert.NotEmpty(collection)` → `collection.Any()`

While the current implementation correctly falls back to returning the original expression (which will cause a compilation error and prompt manual conversion), adding these patterns could improve the migration experience.

**Impact**: Low priority enhancement - current behavior is safe (compilation error prompts manual fix).

### 2. Lambda formatting consistency
In `CreateIsInstanceOfTypeLambda` and `CreateNotNullLambda`, you're adding `.WithArrowToken(...WithTrailingTrivia(SyntaxFactory.Space))` for spacing, which is good. However, in `TryConvertActionToPredicate`, the same pattern is applied. This is consistent and correct.

**No action needed** - just confirming this is intentional and consistent.

### 3. Test coverage for fallback behavior
The test `Assert_All_Can_Be_Converted` covers the happy path where `Assert.True` is successfully converted to a predicate. Consider adding a test case for when conversion isn't possible (e.g., `Assert.All(items, item => item.DoSomething())`) to verify the fallback behavior produces expected results.

**Impact**: Would improve test coverage but not critical since fallback behavior is simple.

## Verdict

The fix correctly addresses issue #4335 by:
1. Using the actual TUnit `.All(predicate)` method signature
2. Intelligently converting common xUnit action patterns to predicates
3. Providing safe fallback behavior for unconvertible patterns
4. Adding test coverage for the conversion

The code is well-structured with good separation of concerns (`CreateAllAssertion` → `TryConvertActionToPredicate`). The MSTest fixes are a bonus cleanup that prevents the same issue there.

PR #4349, opened 2026-01-12
fix: prevent interface implementation methods from being converted to async

### LINKED ISSUE #4342: NUnit Migration: Interface implementation methods incorrectly converted to async
## Description

When running the TUnit NUnit migration code fixer, methods that implement interface members with `void` return types are incorrectly converted to `async Task` methods. This breaks the interface implementation because the interface still expects `void`.

## Steps to Reproduce

1. Create a project with classes that implement interfaces (like `IInterceptor`, `IProxyGenerationHook`) and use NUnit assertions
2. Add TUnit packages  
3. Run `dotnet format analyzers --severity info --diagnostics TUNU0001`
4. Build the project

## Expected Behavior

Methods that implement interface members should NOT have their return type changed. The assertions should either:
1. Use synchronous assertion patterns
2. Be left unchanged for manual migration

## Actual Behavior

Interface implementation methods are converted to async, causing build errors:
```
error CS0738: 'ChangeProxyTargetInterceptor' does not implement interface member 
'IInterceptor.Intercept(IInvocation)'. 'ChangeProxyTargetInterceptor.Intercept(IInvocation)' 
cannot implement 'IInterceptor.Intercept(IInvocation)' because it does not have the 
matching return type of 'void'.
```

## Example

**Interface definition (not modified by migration):**
```csharp
public interface IInterceptor
{
    void Intercept(IInvocation invocation);
}
```

**Before migration:**
```csharp
public class ChangeProxyTargetInterceptor : IInterceptor
{
    public void Intercept(IInvocation invocation)
    {
        var targetAccessor = invocation.Proxy as IProxyTargetAccessor;
        Assert.IsNotNull(targetAccessor);
        targetAccessor.DynProxySetTarget(target);
        invocation.Proceed();
    }
}
```

**After migration (current - broken):**
```csharp
public class ChangeProxyTargetInterceptor : IInterceptor
{
    public async Task Intercept(IInvocation invocation)  // <-- Breaks interface implementation
    {
        var targetAccessor = invocation.Proxy as IProxyTargetAccessor;
        await Assert.That(targetAccessor).IsNotNull();
        targetAccessor.DynProxySetTarget(target);
        invocation.Proceed();
    }
}
```

**Expected after migration:**

Option 1 - Use synchronous pattern:
```csharp
public class ChangeProxyTargetInterceptor : IInterceptor
{
    public void Intercept(IInvocation invocation)
    {
        var targetAccessor = invocation.Proxy as IProxyTargetAccessor;
        Assert.That(targetAccessor).IsNotNull().Wait();  // or synchronous alternative
        targetAccessor.DynProxySetTarget(target);
        invocation.Proceed();
    }
}
```

Option 2 - Leave unchanged for manual migration:
```csharp
public class ChangeProxyTargetInterceptor : IInterceptor
{
    public void Intercept(IInvocation invocation)
    {
        var targetAccessor = invocation.Proxy as IProxyTargetAccessor;
        Assert.IsNotNull(targetAccessor);  // Left for manual migration
        targetAccessor.DynProxySetTarget(target);
        invocation.Proceed();
    }
}
```

## Detection Suggestion

The code fixer could detect interface implementations by checking:
1. If the containing class implements any interfaces
2. If the method signature matches an interface member
3. If so, don't convert to async

## Impact

This is a critical issue because it generates code that cannot compile.

## Environment

- TUnit version: 1.9.91
- .NET version: .NET 8.0
- OS: Windows

thomhurst:
## Summary
Methods that implement interface members should NOT have their signatures converted to `async Task` because this would break the interface implementation contract. 

The fix uses a two-pass approach:
1. **Before syntax modifications**: Collect all methods that implement interface members using semantic analysis (via semantic model)
2. **During async signature rewriting**: Skip methods identified as interface implementations

Also adds `ref`/`out`/`in` parameter check to prevent async conversion of methods with those parameter types (which cannot be async).

## Changes
- Modified `AsyncMethodSignatureRewriter` to accept a set of interface-implementing method keys
- Added static method `CollectInterfaceImplementingMethods` that uses semantic analysis to identify interface implementations
- Updated `BaseMigrationCodeFixProvider` to collect interface-implementing methods before modifications
- Added test for interface implementation scenario

## Test plan
- [x] Run NUnit migration tests (66 passed)
- [x] Run xUnit migration tests (39 passed, 1 skipped)
- [x] Run MSTest migration tests (23 passed)

Fixes #4342

### DIFF
--- a/TUnit.Analyzers.CodeFixers/Base/AsyncMethodSignatureRewriter.cs
+++ b/TUnit.Analyzers.CodeFixers/Base/AsyncMethodSignatureRewriter.cs
@@ -10,6 +10,84 @@ namespace TUnit.Analyzers.CodeFixers.Base;
 /// </summary>
 public class AsyncMethodSignatureRewriter : CSharpSyntaxRewriter
 {
+    private readonly HashSet<string> _interfaceImplementingMethods;
+
+    public AsyncMethodSignatureRewriter() : this(new HashSet<string>())
+    {
+    }
+
+    public AsyncMethodSignatureRewriter(HashSet<string> interfaceImplementingMethods)
+    {
+        _interfaceImplementingMethods = interfaceImplementingMethods;
+    }
+
+    /// <summary>
+    /// Collects method signatures that implement interface members.
+    /// This should be called BEFORE syntax modifications while the semantic model is still valid.
+    /// </summary>
+    public static HashSet<string> CollectInterfaceImplementingMethods(
+        CompilationUnitSyntax compilationUnit,
+        SemanticModel semanticModel)
+    {
+        var methods = new HashSet<string>();
+
+        foreach (var methodDecl in compilationUnit.DescendantNodes().OfType<MethodDeclarationSyntax>())
+        {
+            // Check for explicit interface implementation syntax
+            if (methodDecl.ExplicitInterfaceSpecifier != null)
+            {
+                methods.Add(GetMethodKey(methodDecl));
+                continue;
+            }
+
+            var methodSymbol = semanticModel.GetDeclaredSymbol(methodDecl);
+            if (methodSymbol == null)
+            {
+                continue;
+            }
+
+            // Check if this method explicitly implements an interface
+            if (methodSymbol.ExplicitInterfaceImplementations.Length > 0)
+            {
+                methods.Add(GetMethodKey(methodDecl));
+                continue;
+            }
+
+            // Check if this method implicitly implements an interface member
+            var containingType = methodSymbol.ContainingType;
+            if (containingType != null)
+            {
+                foreach (var iface in containingType.AllInterfaces)
+                {
+                    foreach (var member in iface.GetMembers().OfType<IMethodSymbol>())
+                    {
+                        var impl = containingType.FindImplementationForInterfaceMember(member);
+                        if (SymbolEqualityComparer.Default.Equals(impl, methodSymbol))
+                        {
+                            methods.Add(GetMethodKey(methodDecl));
+                            break;
+                        }
+                    }
+                }
+            }
+        }
+
+        return methods;
+    }
+
+    /// <summary>
+    /// Gets a unique key for a method declaration based on its signature.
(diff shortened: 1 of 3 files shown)


thomhurst:
## Summary
Fixes interface implementation methods being incorrectly converted to async by the migration code fixer.

## Critical Issues
None found ✅

## Suggestions

### 1. Potential Performance Concern - HashSet allocation
The CollectInterfaceImplementingMethods method is called during code fix application. While this is not as hot as test discovery/execution, it still allocates a new HashSet and iterates through all methods.

Location: TUnit.Analyzers.CodeFixers/Base/AsyncMethodSignatureRewriter.cs:19-66

Consider:
- The implementation looks reasonable for a code fixer context (not a hot path)
- The two-pass approach is necessary since semantic model becomes invalid after syntax changes
- No action needed unless profiling shows issues

### 2. Method Key Stability
The GetMethodKey method uses string concatenation to create method signatures. The key format should remain stable across syntax tree modifications (before/after conversion), which appears to be the case.

Location: TUnit.Analyzers.CodeFixers/Base/AsyncMethodSignatureRewriter.cs:83-91

### 3. Test Coverage
The added test covers the implicit interface implementation case well. Consider adding edge cases for explicit interface implementations, generic interfaces, and multiple interface implementations, though the explicit case is already handled by checking ExplicitInterfaceSpecifier.

## Technical Review

✅ Dual-mode: N/A - This is analyzer/code fixer code, not engine metadata collection
✅ Snapshot Testing: N/A - No source generator output or public API changes
✅ No VSTest: No VSTest references
✅ Performance: Acceptable for code fixer context (not hot path)
✅ AOT Compatibility: Uses Roslyn semantic model appropriately (analyzers run at compile-time)

The two-pass approach (collect interface implementations before modifications, then check during rewriting) is the correct solution since the semantic model becomes invalid after syntax tree modifications.

## Verdict
✅ APPROVE - Well-implemented fix with appropriate test coverage. The solution correctly handles the semantic model invalidation problem and prevents breaking interface implementation contracts.

PR #4362, opened 2026-01-13
perf: reuse `List` and remove `ToArray` call

TimothyMakkison:
- Remove unneeded `ToArray` call in `ScheduleAndExecuteAsync`
- Reuse `List<AbstractExecutableTest>` in `CircularDependencyDetector`, clearing it each time.

Should save around 2MB in `List<AbstractExecutableTest>` and `AbstractExecutableTest[]` allocations

### DIFF
--- a/TUnit.Engine/Scheduling/TestScheduler.cs
+++ b/TUnit.Engine/Scheduling/TestScheduler.cs
@@ -117,8 +117,7 @@ internal sealed class TestScheduler : ITestScheduler
             }
         }
 
-        var executableTestsArray = executableTests.ToArray();
-        if (executableTestsArray.Length == 0)
+        if (executableTests.Count == 0)
         {
             await _logger.LogDebugAsync("No executable tests found after removing circular dependencies").ConfigureAwait(false);
             return true;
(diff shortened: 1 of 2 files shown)


thomhurst:
## Summary
Performance optimization that removes unnecessary array allocation and reuses a list buffer in circular dependency detection.

## Critical Issues
None found

## Suggestions

### CircularDependencyDetector.cs: Verify buffer reuse correctness
The buffer reuse optimization looks correct. The pathBuffer is mutated during the DFS traversal in HasCycleDfs. The PR correctly makes a defensive copy when a cycle is found.

The logic is sound because:
1. Each top-level iteration starts fresh with pathBuffer.Clear()
2. When a cycle is found, a copy is made immediately before returning
3. The buffer is properly isolated per top-level test iteration

### TestScheduler.cs: Consider consistency
At line 134 (after the PR), you're passing executableTests (a List) directly to GroupTestsByConstraintsAsync.

However, at lines 246 and 277 in the same file, the code still creates arrays with ToArray() before calling the same method. For consistency and to avoid the same allocation overhead, consider removing those ToArray() calls too in a follow-up PR.

## Verdict
APPROVE - No critical issues. Clean performance optimization that saves ~2MB in allocations as claimed.

Issue #4339, opened 2026-01-12
NUnit Migration: `[Platform]` attribute not handled


thomhurst:
## Description

When running the TUnit NUnit migration code fixer, the NUnit `[Platform]` attribute is not handled. There may not be a direct TUnit equivalent, but the migration should either convert it to an appropriate alternative or flag it for manual review.

## Steps to Reproduce

1. Create a project with NUnit tests that use `[Platform]` attribute
2. Add TUnit packages
3. Run `dotnet format analyzers --severity info --diagnostics TUNU0001`
4. Build the project

## Expected Behavior

Either:
1. Convert to a TUnit equivalent if one exists
2. Remove and add a comment indicating manual migration needed
3. Convert to an appropriate conditional attribute

## Actual Behavior

The `[Platform]` attribute is left unchanged, causing build errors:
```
error CS0246: The type or namespace name 'Platform' could not be found
error CS0246: The type or namespace name 'PlatformAttribute' could not be found
```

## Example

**Before migration:**
```csharp
[Test]
[Platform(Include = "Win,Mono")]
public void SimpleUsage()
{
    // test code
}

[Test]
[Platform(Exclude = "Mono", Reason = "Mono does not support this feature")]
public void AnotherTest()
{
    // test code
}
```

**After migration (current):**
```csharp
[Test]
[Platform(Include = "Win,Mono")]  // <-- Not handled
public async Task SimpleUsage()
{
    // test code
}
```

## Suggested Solutions

1. If TUnit has an equivalent attribute, convert to it
2. If not, consider converting to runtime checks using `Environment.OSVersion` or `RuntimeInformation.IsOSPlatform`
3. Or add a `// TODO: Manual migration required - [Platform] attribute not supported` comment

## Environment

- TUnit version: 1.9.91
- .NET version: .NET 8.0
- OS: Windows

thomhurst:
Fixed in PR #4348. The NUnit migration code fixer now converts \ attributes to TUnit's \ attribute for supported platforms (Win, Linux, MacOS), or adds a TODO comment for unsupported platform names.

Issue #4334, opened 2026-01-12
xUnit Migration: Exception property access (.Message, .ParamName) not properly converted


thomhurst:
## Description

When using the TUnit xUnit migration code fixer (TUXU0001), the fixer incorrectly attempts to access exception properties like `.Message`, `.ParamName`, `.InnerException` on `ThrowsAssertion<T>` objects rather than on the captured exception instance itself, resulting in CS1061 compilation errors.

## Steps to Reproduce

1. Create an xUnit test that captures and inspects an exception:
```csharp
using Xunit;

[Fact]
public void Test_Exception_Message()
{
    var ex = Assert.Throws<ArgumentException>(() => ThrowException());
    Assert.Equal("param", ex.ParamName);
    Assert.Contains("error occurred", ex.Message);
}
```

2. Install TUnit package and run `dotnet format analyzers --severity info --diagnostics TUXU0001`

## Expected Behavior

The code should be converted to properly capture the exception and access its properties:
```csharp
[Test]
public async Task Test_Exception_Message()
{
    var ex = await Assert.That(() => ThrowException()).Throws<ArgumentException>();
    await Assert.That(ex.ParamName).IsEqualTo("param");
    await Assert.That(ex.Message).Contains("error occurred");
}
```

Or use a different pattern that allows property access on the captured exception.

## Actual Behavior

The code is converted to:
```csharp
[Test]
public async Task Test_Exception_Message()
{
    var ex = awaitAssert.ThrowsAsync<ArgumentException>(() => ThrowException());
    await Assert.That(ex.ParamName).IsEqualTo("param");  // ex is ThrowsAssertion, not ArgumentException
    await Assert.That(ex.Message).Contains("error occurred");
}
```

This causes multiple errors:
```
error CS1061: 'ThrowsAssertion<ArgumentException>' does not contain a definition for 'ParamName'
error CS1061: 'ThrowsAssertion<ArgumentException>' does not contain a definition for 'Message'
```

## Properties Affected

The following exception properties commonly accessed after `Assert.Throws<T>()` are not properly handled:
- `.Message`
- `.ParamName` (for ArgumentException)
- `.InnerException`
- `.StackTrace`
- Custom exception properties (like `.IsVerificationError` in Moq's MockException)
- `.Reasons` (custom properties)

## Impact

This is a compound issue with the `awaitAssert` problem, but even if `awaitAssert` were defined correctly, the pattern of capturing an exception and accessing its properties would need different handling.

[...]

Issue #4332, opened 2026-01-12
xUnit Migration: Record.Exception() pattern not converted


thomhurst:
## Description

When using the TUnit xUnit migration code fixer (TUXU0001), the fixer does not convert xUnit's `Record.Exception()` pattern, leaving it as-is. Since TUnit doesn't have a `Record` class, this results in CS0103 compilation errors.

## Steps to Reproduce

1. Create an xUnit test using `Record.Exception()`:
```csharp
using Xunit;

[Fact]
public void Test_Exception_Recording()
{
    var ex = Record.Exception(() => SomeMethod());
    Assert.NotNull(ex);
    Assert.Contains("expected message", ex.Message);
}
```

2. Install TUnit package and run `dotnet format analyzers --severity info --diagnostics TUXU0001`

## Expected Behavior

The code should be converted to a valid TUnit pattern, such as:
```csharp
[Test]
public async Task Test_Exception_Recording()
{
    var assertion = await Assert.That(() => SomeMethod()).ThrowsException();
    await Assert.That(assertion.Message).Contains("expected message");
}
```

Or provide an alternative TUnit pattern for capturing and inspecting exceptions.

## Actual Behavior

The `Record.Exception()` call is left unchanged:
```csharp
[Test]
public async Task Test_Exception_Recording()
{
    var ex = Record.Exception(() => SomeMethod());
    await Assert.That(ex).IsNotNull();
    await Assert.That(ex.Message).Contains("expected message");
}
```

This causes:
```
error CS0103: The name 'Record' does not exist in the current context
```

## Example from Moq Test Suite

Original:
```csharp
[Fact]
public void Built_in_matcher_renders_nicely_when_using_delegate_based_setup_expression()
{
    var mock = new Mock<IOrderRepository>();

    var ex = Record.Exception(() => mock.VerifySet(repo => repo.Value = It.IsAny<int>()));
    Assert.Contains("repo => repo.Value = It.IsAny<int>()", ex.Message);
}
```

After migration (still broken):
```csharp
[Test]
public async Task Built_in_matcher_renders_nicely_when_using_delegate_based_setup_expression()
{
    var mock = new Mock<IOrderRepository>();

    var ex = Record.Exception(() => mock.VerifySet(repo => repo.Value = It.IsAny<int>()));
    await Assert.That(ex.Message).Contains("repo => repo.Value = It.IsAny<int>()");
}
```

## Impact

[...]

PR #4367, opened 2026-01-13
feat: migration code fixer improvements for xUnit and NUnit

thomhurst:
## Summary

This PR addresses several migration code fixer issues:

- **#4332**: Convert xUnit `Record.Exception()` pattern to try-catch blocks
- **#4334**: Verify and test xUnit `Assert.Throws<T>()` exception property access (already working)
- **#4339**: Convert NUnit `[Platform]` attribute to TUnit `[RunOn]`/`[ExcludeOn]`

### xUnit Record.Exception() Conversion (#4332)
Converts the xUnit pattern:
```csharp
var ex = Record.Exception(() => ThrowException());
```
To:
```csharp
Exception? ex = null;
try { ThrowException(); }
catch (Exception e) { ex = e; }
```

### xUnit Exception Property Access (#4334)
Added tests verifying that `Assert.Throws<T>()` with property access works correctly:
- Capturing exception and accessing `.ParamName`
- Using `Assert.Contains` with `ex.Message`

### NUnit [Platform] Conversion (#4339)
Instead of removing the NUnit `[Platform]` attribute, converts to TUnit equivalents:
- `[Platform(Include = "Win")]` → `[RunOn(OS.Windows)]`
- `[Platform(Exclude = "Linux")]` → `[ExcludeOn(OS.Linux)]`
- `[Platform(Include = "Win,Linux")]` → `[RunOn(OS.Windows | OS.Linux)]`

Supported platform mappings:
| NUnit | TUnit |
|-------|-------|
| Win, Win32, Windows | OS.Windows |
| Linux, Unix | OS.Linux |
| MacOsX, MacOS, OSX, Mac | OS.MacOs |

## Test plan

- [x] All new tests pass for Record.Exception conversion
- [x] All new tests pass for exception property access
- [x] All new tests pass for Platform attribute conversion
- [x] Existing migration tests continue to pass

### DIFF
--- a/TUnit.Analyzers.CodeFixers/NUnitMigrationCodeFixProvider.cs
+++ b/TUnit.Analyzers.CodeFixers/NUnitMigrationCodeFixProvider.cs
@@ -138,6 +138,13 @@ public class NUnitAttributeRewriter : AttributeRewriter
             return null;
         }
 
+        // [Platform(Include = "Win")] -> [RunOn(OS.Windows)]
+        // [Platform(Exclude = "Linux")] -> [ExcludeOn(OS.Linux)]
+        if (attributeName == "Platform")
+        {
+            return ConvertPlatformAttribute(attribute);
+        }
+
         return base.ConvertAttribute(attribute);
     }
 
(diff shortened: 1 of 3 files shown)


thomhurst:
## Summary
This PR adds NUnit Platform attribute conversion to RunOn/ExcludeOn and XUnit Record.Exception conversion to try-catch blocks for the migration code fixers.

## Critical Issues
None found ✅

## Suggestions

### 1. NUnit Platform Attribute - Positional Argument Bug
In `NUnitMigrationCodeFixProvider.cs:ConvertPlatformAttribute` (line 222-228), the positional argument handling has a logic issue:

```csharp
if (argName == "Include" || argName == null)
{
    // Named argument Include= or positional argument (which is Include)
    includeValue = value;
}
```

**Issue**: If there are multiple positional arguments (e.g., `[Platform("Win", "Linux")]`), the code will overwrite `includeValue` for each one, only keeping the last value. NUnit's Platform attribute doesn't actually support multiple positional arguments in this way, but the code should be more defensive.

**Suggestion**: Consider only processing the first positional argument or documenting this behavior.

### 2. Record.Exception - Missing Block Body Handling
In `XUnitMigrationCodeFixProvider.cs:ConvertLambdaBodyToStatement` (line 257-265), when a lambda has a block body with multiple statements:

```csharp
if (body is BlockSyntax block)
{
    // If it's a single statement block, extract the statement
    if (block.Statements.Count == 1)
    {
        return block.Statements[0];
    }
    // Otherwise return the block as-is (will need to be a compound statement)
    return block;
}
```

**Issue**: When returning the block directly for multi-statement lambdas, it's placed inside the try block as a nested block:
```csharp
try
{
    {  // Extra braces here
        statement1;
        statement2;
    }
}
```

**Suggestion**: When `actionBody` is already a `BlockSyntax` with multiple statements in `CreateTryCatchStatements`, extract its statements instead of nesting it:
```csharp
if (actionBody is BlockSyntax blockBody && blockBody.Statements.Count > 1)
{
    // Use block's statements directly
    tryBodyStatement = SyntaxFactory.Block(blockBody.Statements)...
}
```

### 3. Record.ExceptionAsync - Missing Async/Await Verification
In `XUnitMigrationCodeFixProvider.cs:CreateTryCatchStatements` (line 286-298), the async handling only checks the invocation name but doesn't verify if the action body actually returns a Task:

```csharp
// If async, we may need to await the action
if (isAsync && tryBodyStatement is ExpressionStatementSyntax exprStmt)
{
    var awaitExpr = SyntaxFactory.AwaitExpression(...);
```

**Issue**: This blindly adds `await` for any `Record.ExceptionAsync` usage, even if the lambda body is synchronous (e.g., `() => throw new Exception()`). While XUnit's `Record.ExceptionAsync` does expect async delegates, users might have incorrect code that still compiles.

**Suggestion**: Consider using semantic analysis to check if the expression is actually awaitable, or document that this migration assumes correct XUnit usage.

### 4. Platform Attribute - Case Sensitivity
In `NUnitMigrationCodeFixProvider.cs:MapNUnitPlatformToTUnitOS` (line 274), the mapping uses `ToLowerInvariant()`:

```csharp
return nunitPlatform.ToLowerInvariant() switch
{
    "win" or "win32" or ... => "Windows",
```

**Minor**: NUnit's platform matching is case-insensitive, so this is correct. However, consider adding a comment explaining this matches NUnit's behavior, as it's not immediately obvious why we're lower-casing.

### 5. Test Coverage - Missing Edge Cases
The tests cover basic scenarios well, but consider adding tests for:
- `[Platform(Include = "UnknownPlatform")]` - what happens with unknown platforms?
- `[Platform(Include = "Win", Exclude = "Linux")]` - both Include and Exclude (currently only Include is used)
- `Record.Exception` with method groups: `Record.Exception(ThrowingMethod)` instead of lambdas
- `Record.ExceptionAsync` with async lambdas: `async () => await SomeAsyncMethod()`

## Verdict

The implementation is solid and follows TUnit conventions well. The suggestions above are optional improvements for edge cases and code robustness, but not blocking issues.

Issue #4344, opened 2026-01-12
NUnit Migration: `dotnet format` crashes on multi-targeted projects


thomhurst:
## Description

When running the TUnit NUnit migration code fixer on projects that target multiple frameworks (e.g., `<TargetFrameworks>net8.0;net9.0;net10.0;net462</TargetFrameworks>`), the `dotnet format analyzers` command crashes with an exception related to linked file merging.

## Steps to Reproduce

1. Create a project with NUnit tests that targets multiple frameworks:
   ```xml
   <TargetFrameworks>net8.0;net9.0;net10.0;net462</TargetFrameworks>
   ```
2. Add TUnit packages
3. Run `dotnet format analyzers --severity info --diagnostics TUNU0001`

## Expected Behavior

The migration should complete successfully for all target frameworks.

## Actual Behavior

The command crashes with the following error:
```
Unhandled exception: System.AggregateException: One or more errors occurred. 
(Changes must be within bounds of SourceText (Parameter 'changes'))
 ---> System.ArgumentException: Changes must be within bounds of SourceText (Parameter 'changes')
   at Microsoft.CodeAnalysis.Text.SourceText.WithChanges(IEnumerable`1 changes)
   at Microsoft.CodeAnalysis.LinkedFileMergeConflictCommentAdditionService.GetCommentChangesForDocument(...)
   at Microsoft.CodeAnalysis.LinkedFileDiffMergingSession.MergeLinkedDocumentGroupAsync(...)
   ...
```

The migration process appears to identify files to format and attempts changes, but crashes when trying to merge the changes across multiple target frameworks.

## Workaround

Temporarily change the project to a single target framework:
```xml
<TargetFramework>net8.0</TargetFramework>
<!-- <TargetFrameworks>net8.0;net9.0;net10.0;net462</TargetFrameworks> -->
```

Then run the migration, and restore multi-targeting afterward.

## Notes

This may be a `dotnet format` issue rather than a TUnit code fixer issue, but it significantly impacts the migration experience for projects with multiple target frameworks. 

The error occurs in `Microsoft.CodeAnalysis.LinkedFileMergeConflictCommentAdditionService`, suggesting the issue is with how changes are being merged across the same files compiled for different target frameworks.

## Possible Solutions

1. Document the workaround in migration guide
2. Investigate if the TUnit code fixer can make changes more compatible with linked file scenarios
3. Report to dotnet/format if this is an upstream issue

## Environment

- TUnit version: 1.9.91
[...]

PR #4369, opened 2026-01-13
fix: add defensive syntax fallbacks to migration code fixers

thomhurst:
## Summary

- Adds deterministic syntax-based fallbacks to migration code fixers to prevent crashes during `dotnet format` on multi-targeted projects
- When semantic analysis fails or produces different results across TFMs, the code fixers now fall back to syntax-based detection ensuring consistent transformations

### `IsLikelyComparerArgument` (AssertionRewriter.cs)
- Tries semantic analysis first for type resolution
- Falls back to case-insensitive syntax-based detection (checks for names containing "comparer")
- Changed return type from `bool?` to `bool` for consistency

### `IsFrameworkAssertion` (AssertionRewriter.cs)  
- Tries semantic analysis first
- Falls back to new `IsKnownAssertionTypeBySyntax` abstract method

### `IsKnownAssertionTypeBySyntax` (new abstract method)
Added to each framework's assertion rewriter with framework-specific patterns:
- **NUnit**: `Assert`, `CollectionAssert`, `StringAssert`, `FileAssert`, `DirectoryAssert`
- **MSTest**: `Assert`, `CollectionAssert`, `StringAssert`, `FileAssert`, `DirectoryAssert`  
- **XUnit**: `Assert`

Note: `ClassicAssert` is explicitly excluded from NUnit's syntax-based detection because it's in `NUnit.Framework.Legacy` which should not be auto-converted.

## Problem Addressed

Related to #4344 - `dotnet format` crashes on multi-targeted projects with error "Changes must be within bounds of SourceText". 

The root cause is that SemanticModel can produce different results for different TFMs, leading to divergent code transformations that can't be merged during linked file processing.

## Test plan

- [x] All 520 analyzer tests pass
- [x] Build succeeds with no new errors

### DIFF
--- a/TUnit.Analyzers.CodeFixers/XUnitMigrationCodeFixProvider.cs
+++ b/TUnit.Analyzers.CodeFixers/XUnitMigrationCodeFixProvider.cs
@@ -385,7 +385,7 @@ public class XUnitMigrationCodeFixProvider : BaseMigrationCodeFixProvider
             var type = objectCreationExpressionSyntax switch
             {
                 ObjectCreationExpressionSyntax explicitObjectCreationExpressionSyntax => explicitObjectCreationExpressionSyntax.Type,
-                ImplicitObjectCreationExpressionSyntax implicitObjectCreationExpressionSyntax => SyntaxFactory.ParseTypeName(compilation.GetSemanticModel(implicitObjectCreationExpressionSyntax.SyntaxTree).GetTypeInfo(implicitObjectCreationExpressionSyntax).Type!.ToDisplayString()),
+                ImplicitObjectCreationExpressionSyntax implicitObjectCreationExpressionSyntax => GetTypeFromImplicitCreation(compilation, implicitObjectCreationExpressionSyntax),
                 _ => null
             };
 
@@ -454,6 +454,31 @@ public class XUnitMigrationCodeFixProvider : BaseMigrationCodeFixProvider
         }
 
         return currentRoot;
+    }
+
+    /// <summary>
+    /// Safely gets the type from an implicit object creation expression using semantic analysis.
+    /// Returns null if semantic analysis fails (defensive for multi-TFM scenarios).
+    /// </summary>
+    private static TypeSyntax? GetTypeFromImplicitCreation(Compilation compilation, ImplicitObjectCreationExpressionSyntax implicitCreation)
+    {
+        try
+        {
+            var semanticModel = compilation.GetSemanticModel(implicitCreation.SyntaxTree);
+            var typeInfo = semanticModel.GetTypeInfo(implicitCreation);
+
+            if (typeInfo.Type is null || typeInfo.Type.TypeKind == TypeKind.Error)
+            {
+                return null;
+            }
+
+            return SyntaxFactory.ParseTypeName(typeInfo.Type.ToDisplayString());
+        }
+        catch (InvalidOperationException)
+        {
+            // Semantic analysis failed due to invalid compilation state
+            return null;
+        }
     }
 
     private static SyntaxNode UpdateInitializeDispose(Compilation compilation, SyntaxNode root)


thomhurst:
## Summary
This PR adds defensive syntax-based fallbacks to migration code fixers to prevent crashes during dotnet format on multi-targeted projects when semantic analysis fails across TFMs.

## Critical Issues

### 1. Return Type Change Needs Verification
The IsLikelyComparerArgument method changed from returning bool? to bool. The call sites in NUnitMigrationCodeFixProvider.cs at lines 1372 and 1393 still use IsLikelyComparerArgument(arguments[2]) == true, which suggests they were expecting nullable behavior. While this still works with the new bool return, the intent has changed:

Before: null meant unknown/ambiguous resulting in conservative default behavior
After: Always returns true or false where syntax fallback makes decision

The change to non-nullable is correct IF the syntax fallback is reliable. However, verify that all call sites handle the new behavior correctly, especially when semantic analysis previously returned null.

### 2. Syntax-Based Comparer Detection May Be Too Broad
In IsLikelyComparerArgumentBySyntax (AssertionRewriter.cs line 376 in diff), the check lowerExpressionText.EndsWith("comparer") could match unintended patterns like myStringComparerFactory or isComparer (boolean variable). Consider being more specific by checking for complete tokens rather than substrings. This is not blocking if tests pass, but worth considering for robustness.

### 3. Record.Exception Try-Catch Indentation Logic Is Complex
The new XUnit ConvertRecordException method generates try-catch blocks with manual indentation handling. This is fragile with mixed tabs/spaces, different indentation widths, and editorconfig settings. Have you tested this with various indentation styles (2-space, 4-space, tabs, mixed)?

## Suggestions

### 4. Platform Attribute Mapping Incompleteness
The MapNUnitPlatformToTUnitOS method maps platform names, but NUnit supports many more platforms (net variants, mono). Currently unmapped platforms are silently ignored (returns null), which removes the attribute. Consider adding a TODO comment when an unknown platform is encountered.

### 5. Missing Abstract Method Documentation
The new IsKnownAssertionTypeBySyntax abstract method needs XML documentation explaining when derived classes should return true vs false, with examples showing NUnit returning true for Assert, CollectionAssert, StringAssert but NOT ClassicAssert.

## Previous Review Status
No previous comments found.

## Verdict
⚠️ REQUEST CHANGES - Critical issue 1 (return type change) needs verification that all call sites handle the new behavior correctly. Issues 2 and 3 should be addressed or explicitly acknowledged as acceptable trade-offs.

Testing Note: The PR states All 520 analyzer tests pass which is excellent. However, please verify: (1) Tests with multi-targeted projects (the original issue scenario), (2) Edge cases for syntax-based fallbacks (unusual variable names, nested expressions), (3) Various indentation styles for Record.Exception conversion.

thomhurst:
## Summary
Adds defensive syntax-based fallbacks to migration code fixers to prevent crashes during `dotnet format` on multi-targeted projects.

## Critical Issues

**None found** ✅

The previous review raised valid concerns about the return type change from `bool?` to `bool` for `IsLikelyComparerArgument`. Upon inspection, **all call sites were properly updated** to remove the `== true` checks:

- TUnit.Analyzers.CodeFixers/NUnitMigrationCodeFixProvider.cs:1378 ✅
- TUnit.Analyzers.CodeFixers/NUnitMigrationCodeFixProvider.cs:1399 ✅  
- TUnit.Analyzers.CodeFixers/XUnitMigrationCodeFixProvider.cs:786 ✅
- TUnit.Analyzers.CodeFixers/XUnitMigrationCodeFixProvider.cs:790 ✅

The semantic change (null → false with syntax fallback) is acceptable because the fallback provides deterministic behavior when semantic analysis fails across TFMs, which is the core issue being addressed.

## Suggestions

### 1. Syntax-Based Comparer Detection Could Match Unintended Patterns

In `IsLikelyComparerArgumentBySyntax` (AssertionRewriter.cs:376 in diff), the check:

```csharp
if (lowerExpressionText.EndsWith("comparer") || ...)
```

Could match edge cases like:
- `myStringComparerFactory` (factory, not comparer instance)
- `isComparer` (boolean variable)
- String literals containing "comparer"

However, since:
1. All 520 analyzer tests pass
2. The fallback is only used when semantic analysis fails
3. The conservative default returns false for ambiguous cases
4. The primary check (`string literals → false`) handles the most common false positive

**This is acceptable.** If false positives occur in practice, they can be refined incrementally.

### 2. Record.Exception Indentation Handling

The `ConvertRecordException` method manually constructs try-catch blocks with string-based indentation. While this works, it's fragile with:
- Mixed tabs/spaces
- Non-standard indentation widths
- Editorconfig conflicts

**Recommendation:** Consider using Roslyn's `NormalizeWhitespace()` or `WithAdditionalAnnotations(Formatter.Annotation)` instead of manual trivia construction. Not blocking since tests pass, but would improve robustness.

### 3. Platform Attribute Mapping Silent Failures

`MapNUnitPlatformToTUnitOS` returns `null` for unknown platforms (e.g., "net", "mono"), causing the attribute to be silently removed. Consider adding a TODO comment when an unmapped platform is encountered:

```csharp
_ => null // TODO: TUnit migration - Unknown platform '{nunitPlatform}' could not be converted
```

### 4. Missing XML Documentation

The new `IsKnownAssertionTypeBySyntax` abstract method lacks documentation explaining:
- When to return true vs false
- Why `ClassicAssert` is NOT included for NUnit
- Examples for each framework

Minor: Add XML doc comments for maintainability.

## Previous Review Status

The previous review flagged:
1. ✅ **Return type change** - Addressed (all call sites updated)
2. ⚠️ **Syntax detection too broad** - Acceptable (tests pass, conservative fallback)
3. ⚠️ **Indentation complexity** - Acceptable (tests pass, could be improved)
4. 💡 **Platform mapping** - Still applicable (suggestion only)
5. 💡 **Missing docs** - Still applicable (suggestion only)

## TUnit Rules Compliance

✅ **Dual-Mode**: N/A (analyzers only, not core engine)  
✅ **Snapshot Testing**: N/A (no source generator/public API changes)  
✅ **No VSTest**: No violations detected  
✅ **Performance**: Code fixers run on-demand, not hot path  
✅ **AOT Compatibility**: N/A (analyzers run at design-time)  
✅ **Security**: No injection vulnerabilities (string parsing is for syntax analysis only, not execution)

## Verdict

✅ **APPROVE** - No critical issues. The previous concerns were addressed (return type) or are acceptable trade-offs (syntax detection robustness vs. determinism). The PR successfully solves the multi-TFM crash issue with 520 passing tests.

PR #4375, opened 2026-01-13
fix(assertions): IsEquivalentTo now uses structural comparison for all value types

### LINKED ISSUE #4358: [Bug]: IsEquivalentTo broken for value types
### Description

Value Types like `ValueTuple` are compared using `Equals` during `IsEquivalentTo`, which is incorrect, as these can contain objects that contain references.

### Expected Behavior

Tuples that contain structurally equal objects are asserted as equal, even when they contain non-record reference types.

### Actual Behavior

Tuples that contain structurally equal objects are asserted as non-equal

### Steps to Reproduce

See reproducer at https://github.com/davidmehren/TUnitIgnoreTypeRepro/blob/broken-equivalence/Tests.cs

`Thing.Numbers` is a non-record reference type inside a record and needs to be compared structurally.

### TUnit Version

1.10.0

### .NET Version

.NET 10.0.1

### Operating System

Linux

### IDE / Test Runner

```shell
████████╗██╗   ██╗███╗   ██╗██╗████████╗
╚══██╔══╝██║   ██║████╗  ██║██║╚══██╔══╝
   ██║   ██║   ██║██╔██╗ ██║██║   ██║
   ██║   ██║   ██║██║╚██╗██║██║   ██║
   ██║   ╚██████╔╝██║ ╚████║██║   ██║
   ╚═╝    ╚═════╝ ╚═╝  ╚═══╝╚═╝   ╚═╝

   TUnit v1.10.0.0 | 64-bit | Fedora Linux 43 (Workstation Edition) | linux-x64 | .NET 10.0.1 | Microsoft Testing Platform v2.0.2

   Engine Mode: SourceGenerated

fehlerhaft Fails (24ms)
  TUnit.Engine.Exceptions.TestFailedException: AssertionException: Expected (foo, bar to be equivalent to (bar, foo)
  but Property  did not match
  Expected: (Thing { Name = Foo, Numbers = System.Int32[] }, Thing { Name = Foo, Numbers = System.Int32[] })
  Received: (Thing { Name = Foo, Numbers = System.Int32[] }, Thing { Name = Foo, Numbers = System.Int32[] })
  
  at Assert.That((foo, bar)).IsEquivalentTo((bar, foo))

Testlaufzusammenfassung: Fehler! - bin/Debug/net10.0/TUnitIgnoreTypeRepro.dll (net10.0|x64)
  gesamt: 1
  fehlerhaft: 1
  erfolgreich: 0
  übersprungen: 0
  Dauer: 186ms
```

### Additional Context

_No response_

### IDE-Specific Issue?

thomhurst:
## Summary

- Fixes `IsEquivalentTo` for ValueTuples and structs containing reference type properties
- Removes the `IEquatable<T>` optimization that caused value types to use `Equals()` instead of structural comparison
- Adds 46 comprehensive tests covering various scenarios

## Problem

Previously, `TypeHelper.IsPrimitiveOrWellKnownType()` treated any value type implementing `IEquatable<T>` as a "primitive" type, which meant it used `Equals()` instead of structural comparison. This caused issues with:

- `ValueTuple<T1, T2, ...>` containing records with array properties
- Any struct with reference type fields that need deep comparison

For example, this test would fail:
```csharp
public record Thing(string Name, int[] Numbers);

var foo = new Thing("Foo", [1, 2, 3]);
var bar = new Thing("Foo", [1, 2, 3]);

// Would FAIL because ValueTuple.Equals() delegates to record.Equals()
// which uses reference equality for arrays
await Assert.That((foo, bar)).IsEquivalentTo((bar, foo));
```

## Solution

Removed the `IEquatable<T>` check for value types. `IsEquivalentTo` should always perform structural comparison:

- **For actual primitives** (int, string, DateTime, etc.) - handled explicitly as primitives
- **For composite types** (structs) - compare their fields/properties structurally
- **For types like Vector2** - structural comparison of X/Y yields the same result as `Equals()` anyway

Users who want `Equals()` behavior should use `IsEqualTo()` instead.

## Test plan

- [x] All 46 new tests pass covering:
  - ValueTuple with various arities (2-7 elements)
  - Nested ValueTuples (multiple levels deep)
  - Structs with reference properties
  - Structs implementing `IEquatable<T>` (verifies structural comparison is used, not `Equals()`)
  - Nullable references in structs
  - Empty collections in tuples/structs
  - Complex mixed scenarios (lists of tuples, arrays of structs, dictionaries)
- [x] All existing equivalency tests pass (41 tests)

Fixes #4358

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/TUnit.Assertions.Tests/Bugs/Tests4358.cs
+++ b/TUnit.Assertions.Tests/Bugs/Tests4358.cs
@@ -1,0 +1,788 @@
+namespace TUnit.Assertions.Tests.Bugs;
+
+/// <summary>
+/// Tests for issue #4358: IsEquivalentTo broken for value types
+/// ValueTuples and structs containing reference type properties were incorrectly
+/// compared using Equals() instead of structural comparison.
+/// </summary>
+public class Tests4358
+{
+    #region Test Types
+
+    public record Thing(string Name, int[] Numbers);
+
+    public record Person(string Name, int Age, List<string> Tags);
+
+    public class NonRecordClass
+    {
+        public string Value { get; set; } = string.Empty;
+        public int[] Data { get; set; } = [];
+    }
+
+    public struct StructWithReferenceProperties
+    {
+        public string Name { get; set; }
+        public int[] Numbers { get; set; }
+        public List<string> Tags { get; set; }
+    }
+
+    public struct StructWithEquatable : IEquatable<StructWithEquatable>
+    {
+        public string Id { get; set; }
+        public int[] Data { get; set; }
+
+        // This Equals only compares Id, NOT Data - demonstrating why structural comparison is needed
+        public bool Equals(StructWithEquatable other) => Id == other.Id;
+        public override bool Equals(object? obj) => obj is StructWithEquatable other && Equals(other);
+        public override int GetHashCode() => Id?.GetHashCode() ?? 0;
+    }
+
+    public struct NestedStruct
+    {
+        public string Name { get; set; }
+        public StructWithReferenceProperties Inner { get; set; }
+    }
+
+    public struct StructWithNullableReference
+    {
+        public string? Name { get; set; }
+        public int[]? Numbers { get; set; }
+    }
+
+    #endregion
+
+    #region Original Bug Reproduction Tests
+
+    [Test]
+    public async Task IsEquivalentTo_ValueTuple_WithEquivalentRecords_ShouldSucceed()
+    {
+        // Arrange - Two structurally equivalent records (different instances, same content)
+        var foo = new Thing("Foo", [1, 2, 3]);
+        var bar = new Thing("Foo", [1, 2, 3]);
+
+        // Act & Assert - Should pass because records have equivalent content
+        // This test verifies that ValueTuples use structural comparison, not Equals()
+        await Assert.That((foo, bar)).IsEquivalentTo((bar, foo));
+    }
+
+    [Test]
+    public async Task IsEquivalentTo_Record_WithArrayProperty_ShouldSucceed()
+    {
+        // Arrange - Two records with identical array content but different array instances
+        var foo = new Thing("Foo", [1, 2, 3]);
+        var bar = new Thing("Foo", [1, 2, 3]);
+
+        // Act & Assert - Records with equivalent content should be considered equivalent
+        await Assert.That(foo).IsEquivalentTo(bar);
+    }
+
+    [Test]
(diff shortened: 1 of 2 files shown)


thomhurst:
## Summary
Fixes IsEquivalentTo to use structural comparison for all value types, resolving issues with ValueTuples and structs containing reference properties.

## Critical Issues
None found ✅

## Analysis

This PR correctly addresses issue #4358 by removing the IEquatable<T> optimization for value types in the TypeHelper.IsPrimitiveOrWellKnownType() method.

### What Changed
The PR removes lines 90-98 from TypeHelper.cs that treated any value type implementing IEquatable<T> as a "primitive" type. This was causing ValueTuples and structs with reference type fields to use Equals() instead of structural comparison.

### Why This Is Correct
1. **Root Cause**: ValueTuple<T1, T2> implements IEquatable<ValueTuple<T1, T2>>, which triggered the old optimization. When the tuple contained records with array properties, ValueTuple.Equals() delegated to the record's Equals(), which uses reference equality for arrays.

2. **Solution**: By removing this optimization, IsEquivalentTo now performs structural comparison for ALL value types (except explicitly listed primitives), which correctly handles:
   - ValueTuples with any content (records, arrays, nested structures)
   - Custom structs with reference type fields
   - Nested value types

3. **Performance Impact**: Minimal. For actual primitives like Vector2, structural comparison of X/Y properties yields identical results to Equals() anyway. The TypeHelper method is only called during assertion execution (not a hot path during test discovery).

### Test Coverage
Excellent - 46 comprehensive tests covering:
- All ValueTuple arities (2-7 elements)
- Nested tuples (multiple levels)
- Structs with reference properties
- Structs implementing IEquatable<T> (verifying structural comparison wins)
- Empty collections, nullable references
- Complex mixed scenarios

### TUnit Rules Compliance
✅ **Dual-Mode**: Not applicable (assertions are post-discovery)
✅ **Snapshot Testing**: Not applicable (no generator or public API changes)
✅ **No VSTest**: No VSTest dependencies added
✅ **Performance**: Change affects assertion execution only, not hot paths
✅ **AOT Compatible**: Existing [DynamicallyAccessedMembers] annotations remain intact

### Code Quality
- Clear inline comments explaining the design decision (TUnit.Assertions/Conditions/Helpers/TypeHelper.cs:90-96 in the diff)
- Users wanting Equals() behavior can use IsEqualTo() instead
- ImplementsSelfEquatable() method removed entirely (dead code cleanup)
- Tests are well-organized with clear section comments

## Verdict
✅ **APPROVE** - This is a correct bug fix with excellent test coverage that follows all TUnit development principles.

PR #4393, opened 2026-01-13
perf: avoid creating closure

TimothyMakkison:
Prevent closure creation for `AssmeblyMetada` and `ClassMetadata`, using `GetOrAdd<TArg>(TKey, Func<TKey,TArg,TValue>, TArg)`
- Also removed an extra line 

Not a massive saving probably around 100KB on `TUnit.TestProject`


thomhurst:
## Summary
Eliminates closure allocations in metadata caching hot paths by using the state-passing overload of `ConcurrentDictionary.GetOrAdd`.

## Critical Issues
None found ✅

## Suggestions
None - this is a clean performance optimization that aligns perfectly with TUnit's "Performance First" principle. The use of `static` lambdas with the state-passing `GetOrAdd` overload is exactly the right approach for hot path allocation reduction.

## Verdict

Great optimization! This is precisely the kind of hot path improvement that benefits test discovery performance across all TUnit projects.

PR #4405, opened 2026-01-14
perf: make methods synchronous in `ObjectLifecycleService`

TimothyMakkison:
- Remove `async` keyword making methods synchronous
- Use `ValueListBuilder` in `RegisterArgumentsAsync`, using the newer `Task.WhenAll(ReadOnlySpan<Task>)` overload where possible.
  - Older versions of .NET either return a single `Task` or create an array.

### Before
image

### After
image

### DIFF
--- a/TUnit.Engine/Services/ObjectLifecycleService.cs
+++ b/TUnit.Engine/Services/ObjectLifecycleService.cs
@@ -4,7 +4,6 @@ using TUnit.Core;
 using TUnit.Core.Helpers;
 using TUnit.Core.Interfaces;
 using TUnit.Core.PropertyInjection;
-using TUnit.Core.PropertyInjection.Initialization;
 using TUnit.Core.Tracking;
 
 namespace TUnit.Engine.Services;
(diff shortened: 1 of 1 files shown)


thomhurst:
## Summary
This PR optimizes ObjectLifecycleService by removing unnecessary async keywords and using ValueListBuilder for better performance in RegisterArgumentsAsync.

## Critical Issues
None found ✅

## Suggestions

### Consider Task.WhenAll behavior on older .NET versions
In RegisterArgumentsAsync, the fallback for pre-NET9 targets allocates a new array via .ToArray(), which reduces the performance benefit of using ValueListBuilder. However, this is only for older .NET targets where the span-based Task.WhenAll overload does not exist, so it is a reasonable trade-off. The NET9+ path gets the full benefit.

### Performance validation
The PR description shows allocation improvements in a screenshot. Since this touches a hot path (object registration), the performance improvement is welcome and aligns with TUnits Performance First principle.

## Previous Review Status
No previous comments found.

## Verdict
✅ APPROVE - No critical issues. The changes are clean, correct, and improve performance:
- Methods that were unnecessarily async are now synchronous (eliminating state machine overhead)
- ValueListBuilder reduces allocations on the happy path
- ValueTask.AsTask() is correctly used for the cleanup method
- Unused using directive correctly removed
- All changes preserve existing behavior while improving efficiency

PR #4407, opened 2026-01-14
feat: only emit `Properties` setter when it contains elements

TimothyMakkison:
Only emit `ClassMetadata` reference setter when there are properties cutting down on unneeded generated code. Might have a tiny performance benefit 🤷 the JIT is hard to understand.

I did think that the loop could be a dedicated helper method inside `TUnit.Core`

### DIFF
--- a/TUnit.Core.SourceGenerator/Utilities/MetadataGenerationHelper.cs
+++ b/TUnit.Core.SourceGenerator/Utilities/MetadataGenerationHelper.cs
@@ -576,7 +576,7 @@ internal static class MetadataGenerationHelper
 
         propertyCount = properties.Count;
 
-        if (!properties.Any())
+        if (properties.Count == 0)
         {
             writer.Append("global::System.Array.Empty<global::TUnit.Core.PropertyMetadata>()");
             return;
@@ -618,7 +618,7 @@ internal static class MetadataGenerationHelper
 
         propertyCount = properties.Count;
 
-        if (!properties.Any())
+        if (properties.Count == 0)
         {
             return "global::System.Array.Empty<global::TUnit.Core.PropertyMetadata>()";
         }


TimothyMakkison:
While I can run some tests, anything that relies on `Sourcy` won't build. Would it be possible if you update the snapshot tests?

TUnit.Core.SourceGenerator/Utilities/MetadataGenerationHelper.cs
```diff
@@ -601,13 +609,15 @@ private static void WritePropertyMetadataArray(ICodeWriter writer, INamedTypeSym
...
+    private static string GeneratePropertyMetadataArray(INamedTypeSymbol typeSymbol, int currentIndentLevel, out int propertyCount)
     {
         var properties = typeSymbol.GetMembers()
             .OfType<IPropertySymbol>()
             .Where(p => p.DeclaredAccessibility == Accessibility.Public && p.Name != "EqualityContract")
             .ToList();
 
+        propertyCount = properties.Count;
+
         if (!properties.Any())
```
thomhurst: Probably better to use `propertyCount == 0` here to avoid LINQ too?

TUnit.Core.SourceGenerator/Utilities/MetadataGenerationHelper.cs
```diff
@@ -561,13 +567,15 @@ private static string GenerateParameterMetadataArrayForConstructor(IMethodSymbol
...
+    private static void WritePropertyMetadataArray(ICodeWriter writer, INamedTypeSymbol typeSymbol, out int propertyCount)
     {
         var properties = typeSymbol.GetMembers()
             .OfType<IPropertySymbol>()
             .Where(p => p.DeclaredAccessibility == Accessibility.Public && p.Name != "EqualityContract")
             .ToList();
 
+        propertyCount = properties.Count;
+
         if (!properties.Any())
```
thomhurst:
```suggestion
        if (propertyCount == 0)
```

TUnit.Core.SourceGenerator/Utilities/MetadataGenerationHelper.cs
thomhurst:
```suggestion
        if (propertyCount == 0)
```

thomhurst:
> While I can run some tests, anything that relies on Sourcy won't build. Would it be possible if you update the snapshot tests?

This is really annoying :(

Does the compiler give any sort of error or anything?

TUnit.Core.SourceGenerator/Utilities/MetadataGenerationHelper.cs
TimothyMakkison: `List.Count` is a property and returns an already known integer, this is an `O(1)` operation. 

Calling `someEnumerable.Count()` will use `Linq` if the underlying `IEnumerable` isn't backed by a constant length collection like an array or `List`

TimothyMakkison:
>         if (propertyCount == 0)

👍 added.

Was thinking of looking at the source generators performance as I assume they contribute to build times. Can't imagine it's possible to do any more incremental generator caching due to the nature of `TUnit`

> Does the compiler give any sort of error or anything?

`Sourcy does not exist in the current context`. This issue persists in terminal, rider and VS2026.I wonder if I need the exact same version of .NET for it to work

>     Would it be possible if you update the snapshot tests?

Just checking; you do have an extension which lets you mass approve verify changes right? I don't want to waste your evening 😅

thomhurst:
I wonder if it's failing to run on your machine for some reason (it's a source generator)

thomhurst:
> Just checking; you do have an extension which lets you mass approve verify changes right? I don't want to waste your evening

Yep!

TUnit.Core.SourceGenerator/Utilities/MetadataGenerationHelper.cs
TimothyMakkison: lmao just understood why you said that 😅  mb

thomhurst:
## Summary
This PR optimizes source generator output by only emitting the `foreach` loop that sets `ClassMetadata` and `ContainingTypeMetadata` references on properties when the class actually has properties, avoiding unnecessary code generation for classes with empty property arrays.

## Critical Issues
None found ✅

## Suggestions

### Minor: Consider extracting loop body to helper method
The PR author mentioned in the description: "I did think that the loop could be a dedicated helper method inside `TUnit.Core`". This is worth considering as it would:
- Further reduce generated code size
- Centralize the property metadata initialization logic
- Make future changes to property initialization easier to maintain

However, this would require:
- A public/internal helper method in TUnit.Core
- Passing the properties array as a parameter
- Verifying AOT compatibility

This is not blocking and can be done in a follow-up if desired.

### Observation: Consistent optimization patterns
The change is applied to both overloads of the property array generation methods (`WritePropertyMetadataArray` and `GeneratePropertyMetadataArray`), which is correct and maintains consistency.

## Verification Checklist

✅ **Snapshot tests updated**: All 100+ `.verified.txt` files updated, no `.received.txt` files committed  
✅ **Dual-mode compatibility**: Not applicable - this only affects source generator output, not reflection mode  
✅ **AOT compatibility**: No changes to `[DynamicallyAccessedMembers]` annotations needed  
✅ **Performance**: Positive impact - eliminates unnecessary foreach loops in generated code  
✅ **No VSTest references**: Not applicable to this change  

## Verdict
✅ **APPROVE** - Clean optimization that reduces generated code size without affecting functionality.

PR #4410, opened 2026-01-14
perf: use `VLB` and make methods sync in `EventReceiverOrchestrator`

TimothyMakkison:
- Make methods synchronous
- Use `ValueListBuilder` in `RegisterReceivers`

### Before

image

### After
image

Fun fact: `Attribute.Equals` and `Attribute.GetHashCode` allocate a `FieldInfo[]` each time they are called.

### DIFF
--- a/TUnit.Engine/Building/TestBuilder.cs
+++ b/TUnit.Engine/Building/TestBuilder.cs
@@ -1038,15 +1038,13 @@ internal sealed class TestBuilder : ITestBuilder
 #if NET6_0_OR_GREATER
     [System.Diagnostics.CodeAnalysis.RequiresUnreferencedCode("Scoped attribute filtering uses Type.GetInterfaces and reflection")]
 #endif
-    private async Task InvokeDiscoveryEventReceiversAsync(TestContext context)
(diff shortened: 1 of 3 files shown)


thomhurst:
## Summary
Performance optimization using ValueListBuilder for allocation reduction and making methods synchronous.

## Critical Issues

Missing try-finally in RegisterReceivers (line 41-77): The ValueListBuilder is only disposed if no exception occurs. This could leak array pool buffers.

## Verdict
REQUEST CHANGES - Need try-finally block around VLB usage.

thomhurst:
### Detailed Fix

In EventReceiverOrchestrator.RegisterReceivers, the ValueListBuilder needs try-finally:

Before: vlb.Dispose() only runs if no exception occurs
After: Wrap the usage in try-finally to guarantee disposal

This prevents leaking array pool buffers when exceptions occur during event object enumeration or registration.

### Performance Notes

The changes are excellent for performance:
- ValueListBuilder reduces allocations vs List
- Synchronous methods avoid async state machine overhead
- Early exit checks prevent allocations when no receivers exist

The only minor point is the duplicated early-exit logic in InvokeTestEndEventReceiversAsync, but the performance benefit likely justifies it.

PR #4427, opened 2026-01-15
fix: handle [Test(Description=...)] in NUnit migration code fixer

### LINKED ISSUE #4426: TUNU0001 code fixer: NUnit [Test(Description=...)] produces uncompilable code
## Description

The TUNU0001 code fixer keeps the `Description` property on `[Test]` attributes when converting from NUnit, but TUnit's `[Test]` attribute doesn't have a `Description` property, resulting in compile errors.

## Steps to Reproduce

1. Have an NUnit test with Description property:
```csharp
using NUnit.Framework;

[TestFixture]
public class MyTests
{
    [Test(Description = "This test verifies that the calculation is correct")]
    public void MyCalculationTest()
    {
        Assert.AreEqual(4, 2 + 2);
    }
}
```

2. Run the code fixer:
```bash
dotnet format analyzers --severity info --diagnostics TUNU0001
```

## Expected Behavior

The code fixer should either:
1. Remove the `Description` property entirely:
```csharp
public class MyTests
{
    [Test]
    public async Task MyCalculationTest()
    {
        await Assert.That(2 + 2).IsEqualTo(4);
    }
}
```

2. Or add a comment preserving the description:
```csharp
public class MyTests
{
    // Description: This test verifies that the calculation is correct
    [Test]
    public async Task MyCalculationTest()
    {
        await Assert.That(2 + 2).IsEqualTo(4);
    }
}
```

3. Or use TUnit's equivalent attribute if one exists (e.g., a separate description attribute)

## Actual Behavior

The code fixer keeps the `Description` property:
```csharp
public class MyTests
{
    [Test(Description = "This test verifies that the calculation is correct")]  // <-- ERROR!
    public async Task MyCalculationTest()
    {
        await Assert.That(2 + 2).IsEqualTo(4);
    }
}
```

Build error:
```
error CS0246: The type or namespace name 'Description' could not be found (are you missing a using directive or an assembly reference?)
```

## Real-World Example

From Castle.Core.Tests migration:
```csharp
[Test(Description =
    "There is a strange CLR bug resulting from our loading the tokens of methods in generic types. " +
    "This test ensures we do not trigger it.")]
public async Task TypeGetMethodsIsStable()
{
    // ...
}
```

## Environment

- TUnit Version: 1.11.*
- .NET SDK: 10.0.102
- OS: Windows

## Suggested Fix

1. Strip the `Description` property from `[Test]` attributes during conversion
2. Optionally preserve the description as an XML doc comment:
```csharp
/// <summary>
/// This test verifies that the calculation is correct
/// </summary>
[Test]
public async Task MyCalculationTest()
```

## Tests

Please add tests for:
- Converting `[Test(Description = "...")]` to `[Test]` (with description removed)
- Handling multi-line description strings
- Optionally verifying description is preserved as a comment if that approach is taken

thomhurst:
## Summary

- Fixes #4426 - NUnit `[Test(Description="...")]` produces uncompilable code

NUnit's `[Test]` attribute supports properties like `Description` and `Author` that TUnit's `[Test]` attribute does not support. This change:

- Extends `NUnitAttributeRewriter` to strip named properties from `[Test]` and `[Theory]` attributes via `ConvertTestArguments()`
- Extends `NUnitTestCasePropertyRewriter` to extract `Description` and `Author` from `[Test]` attributes and convert them to `[Property]` attributes
- Adds tests verifying using directive removal for `NUnit.Framework` and subnamespaces

## Test plan

- [x] All 500 NUnit migration tests pass
- [x] New tests verify `[Test(Description="...")]` is correctly converted

### DIFF
--- a/TUnit.Analyzers.CodeFixers/NUnitMigrationCodeFixProvider.cs
+++ b/TUnit.Analyzers.CodeFixers/NUnitMigrationCodeFixProvider.cs
@@ -437,11 +437,39 @@ public class NUnitAttributeRewriter : AttributeRewriter
     {
         return attributeName switch
         {
+            "Test" or "Theory" => ConvertTestArguments(argumentList),
             "TestCase" => ConvertTestCaseArguments(argumentList),
             "TestCaseSource" => ConvertTestCaseSourceArguments(argumentList),
             "Category" => ConvertCategoryArguments(argumentList),
             _ => argumentList
         };
+    }
+
+    private AttributeArgumentListSyntax? ConvertTestArguments(AttributeArgumentListSyntax argumentList)
+    {
+        // NUnit's [Test] attribute supports Description, Author, ExpectedResult, TestOf, etc.
+        // TUnit's [Test] attribute does not support these named properties.
+        // Strip all named arguments - they will be converted to [Property] attributes
+        // by NUnitTestCasePropertyRewriter (for Description and Author).
+        // ExpectedResult is not applicable for [Test] without parameters.
+
+        // If all arguments are named properties to strip, return null to remove argument list entirely
+        if (argumentList.Arguments.All(arg => arg.NameEquals != null))
+        {
+            return null;
+        }
+
+        // Keep only positional arguments (shouldn't normally exist for [Test])
+        var positionalArgs = argumentList.Arguments
+            .Where(arg => arg.NameEquals == null)
+            .ToList();
+
+        if (positionalArgs.Count == 0)
+        {
+            return null;
+        }
+
+        return SyntaxFactory.AttributeArgumentList(SyntaxFactory.SeparatedList(positionalArgs));
     }
 
     private AttributeArgumentListSyntax ConvertTestCaseArguments(AttributeArgumentListSyntax argumentList)
--- a/TUnit.Analyzers.CodeFixers/NUnitTestCasePropertyRewriter.cs
+++ b/TUnit.Analyzers.CodeFixers/NUnitTestCasePropertyRewriter.cs
@@ -6,25 +6,30 @@ using Microsoft.CodeAnalysis.CSharp.Syntax;
 namespace TUnit.Analyzers.CodeFixers;
 
 /// <summary>
-/// Extracts NUnit TestCase properties and converts them to TUnit attributes.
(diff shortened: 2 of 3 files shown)


thomhurst:
## Summary
Fixes #4426 by extending NUnit migration to handle `[Test(Description="...")]` and other named properties on `[Test]` and `[Theory]` attributes.

## Critical Issues
None found ✅

## Suggestions

### 1. Consider handling `[Theory]` attributes in `GetTestAttributes`
The `GetTestAttributes` method in `NUnitTestCasePropertyRewriter.cs:64-82` correctly identifies both `Test` and `Theory` attributes, which is good since `ConvertTestArguments` strips properties from both. However, I notice that NUnit's `[Theory]` attribute doesn't actually support `Description` or `Author` properties (only `[Test]` does). While this doesn't cause incorrect behavior (it just extracts nothing), you might want to document this or narrow the scope to only `[Test]` in `ExtractTestProperties`.

### 2. Minor: Duplicate property merging could be more efficient
In `MergeProperties` at TUnit.Analyzers.CodeFixers/NUnitTestCasePropertyRewriter.cs:210-228, the method adds all source properties to target HashSets. Since these are HashSets, duplicates are handled automatically. However, if the goal is to preserve only unique values across Test and TestCase attributes, consider documenting this behavior - it's correct but subtle.

### 3. Test coverage: Consider edge case with `ExpectedResult`
The comment at TUnit.Analyzers.CodeFixers/NUnitMigrationCodeFixProvider.cs:453 mentions "ExpectedResult is not applicable for [Test] without parameters" - this is correct. However, NUnit allows `[Test(ExpectedResult = value)]` with a return type. While your code silently ignores it (reasonable), consider adding a test case to verify this edge case is handled gracefully (i.e., the ExpectedResult is stripped without errors).

## Verdict

The implementation correctly:
- Strips named properties from `[Test]` and `[Theory]` attributes that TUnit doesn't support
- Extracts `Description` and `Author` properties and converts them to `[Property]` attributes  
- Handles the dual-path conversion (both `ConvertTestArguments` stripping AND `ExtractTestProperties` extraction)
- Includes comprehensive test coverage for the new functionality
- Properly removes argument lists when all arguments are named properties

The code is well-structured and the test coverage verifies the core scenarios. Nice work! 🤖

PR #4447, opened 2026-01-16
fix: improve error messages for unresolved generic type arguments in test generation

thomhurst:
This pull request improves generic test support and error messaging in the test engine and source generator


thomhurst:
## Summary
Improves error messages for generic tests and fixes detection of class-level data sources on generic types.

## Critical Issues
None found ✅

## Suggestions

### 1. Source Generator Change May Need Dual-Mode Verification

The source generator now detects class-level data sources on generic types via the `hasClassDataSources` check:

```csharp
var hasClassDataSources = testMethod.IsGenericType && testMethod.TypeSymbol.GetAttributesIncludingBaseTypes()
    .Any(a => DataSourceAttributeHelper.IsDataSourceAttribute(a.AttributeClass));
```

**Question**: Does the reflection-based test discovery in `TUnit.Engine` handle this scenario correctly? According to TUnit's dual-mode rule, changes to test discovery logic should work identically in both source-gen and reflection modes.

The change appears to fix a bug where generic classes with class-level data sources weren't being detected properly. If this is a bug fix, the reflection path may need the same fix, or we should verify it already handles this case.

Reference: `.claude/docs/mandatory-rules.md` - Rule 1 (Dual-Mode Implementation)

### 2. Snapshot Testing

Since the source generator output has changed (new code path for `hasClassDataSources`), snapshot tests should be run:

```bash
dotnet test TUnit.Core.SourceGenerator.Tests
```

If snapshots changed, commit the `.verified.txt` files (never `.received.txt`).

Reference: `.claude/docs/mandatory-rules.md` - Rule 2 (Snapshot Testing)

### 3. Minor: InterceptorsNamespaces Change

The change to `TUnit.Example.Asp.Net.TestProject.csproj` adds an InterceptorsNamespaces property.

This seems unrelated to the PR's stated purpose (improving generic test error messages). Is this an incidental fix for a build issue, or should it be in a separate commit?

## Verdict
💬 **COMMENT** - Excellent error message improvements! Just want to verify dual-mode compatibility and snapshot testing.

PR #4455, opened 2026-01-17
perf: eliminate `Func` and closure allocation in `DedicatedThreadExecutor`

TimothyMakkison:
Pass `action` into `StartNew` eliminating allocation of `Func<TResult>` and associated closure

### Before
image

### After
image


thomhurst:
## Summary
Eliminates closure and Func allocation in DedicatedThreadExecutor by passing action as a parameter to StartNew instead of capturing it.

## Critical Issues
None found ✅

## Suggestions
None - this is a clean performance optimization with measurable benefit in a hot path.

## Verdict

The static lambda + parameter passing approach correctly eliminates the closure allocation while maintaining the same behavior. The cast back to `Func<ValueTask>` is safe since we're passing the original typed parameter through StartNew's object state parameter.

PR #4460, opened 2026-01-17
perf: eliminate `Func` creation in `ExecuteAsync`

TimothyMakkison:
Small change avoid `Action` creation by using a static lambda and manually boxing state.]

### Before
image

### After
image

### DIFF
--- a/TUnit.Core/Executors/DedicatedThreadExecutor.cs
+++ b/TUnit.Core/Executors/DedicatedThreadExecutor.cs
@@ -18,17 +18,18 @@ public class DedicatedThreadExecutor : GenericAbstractExecutor, ITestRegisteredE
 
         var tcs = new TaskCompletionSource<object?>();
 
-        var thread = new Thread(() =>
+        var thread = new Thread(static state =>
         {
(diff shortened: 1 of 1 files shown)


thomhurst:
## Summary
Eliminates delegate allocation in `DedicatedThreadExecutor.ExecuteAsync` by using a static lambda with manually boxed state.

## Critical Issues
None found ✅

## Suggestions
None - this is a clean performance optimization that aligns with TUnit's "Performance First" principle.

## Analysis
The change converts a capturing lambda (which allocates a closure object) to a static lambda with explicit state passing:
- All captured variables (`this`, `action`, `tcs`) are correctly passed through the tuple
- Thread-safety is maintained - the tuple is created before `thread.Start(state)`
- The optimization is meaningful since `DedicatedThreadExecutor` is used for STAThread tests and dedicated thread execution, making this a hot path
- AOT-compatible and uses modern C# patterns
- Bonus: Fixes indentation on line 412

The benchmark screenshots show measurable improvement in allocations, which directly supports TUnit's performance-first philosophy.

## Verdict

PR #4508, opened 2026-01-19
perf: defer `Queue` creation in `DedicatedThreadExecutor`

TimothyMakkison:
Defer initialisation of `Queue` in `DedicatedThreadExecutor`

I tried this with `List<Task> _taskQueue` and didn't see an improvement, presumably it will always contain at least one object so deferring is useless.

### Before
image

### After
image

### DIFF
--- a/TUnit.Core/Executors/DedicatedThreadExecutor.cs
+++ b/TUnit.Core/Executors/DedicatedThreadExecutor.cs
@@ -205,10 +205,10 @@ public class DedicatedThreadExecutor : GenericAbstractExecutor, ITestRegisteredE
     {
         private readonly Thread _dedicatedThread;
         private readonly ManualResetEventSlim? _workAvailableEvent;
+        private readonly Lock _queueLock = new();
         private readonly List<Task> _taskQueue =
         [
         ];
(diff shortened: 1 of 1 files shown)


thomhurst:
## Summary
Defers Queue allocation in DedicatedThreadSynchronizationContext to reduce memory allocations when the work queue is never used.

## Critical Issues
None found ✅

## Suggestions
None - this is a solid performance optimization that aligns with TUnit's "Performance First" principle. The lazy initialization pattern is correctly implemented with proper null checks and thread-safe initialization under the existing lock.

## Verdict

The change demonstrates:
- Reduced allocations in hot path (aligns with CRITICAL RULE #4)
- Thread-safe lazy initialization
- Proper null handling in ProcessPendingWork
- Measurable performance improvement shown in benchmarks

TimothyMakkison:
Is the modular pipelines check broken? #4508 is also failing because of a weak refernce check failiure in NET framework

thomhurst:
> Is the modular pipelines check broken? #4508 is also failing because of a weak refernce check failiure in NET framework

I think it's a flaky test that I need to look into

PR #4510, opened 2026-01-19
perf: store `_taskQueue` as a single type or `List`

TimothyMakkison:
Treat `_taskQueue ` as a pseudo enum, where it can be empty, a single value or `List<Task>`. Fairly compiclated way to save 3MB WDYT.

### Before
image

### After
image


thomhurst:
Ehhh honestly I don't think it's worth it.
I'm all for the improvements if they're quick wins and easy swap outs, but creating code paths that are harder to maintain I don't think is worth it

TimothyMakkison:
> Ehhh honestly I don't think it's worth it. I'm all for the improvements if they're quick wins and easy swap outs, but creating code paths that are harder to maintain I don't think is worth it

Works for me, this could be revisited in the future when discriminated unions are added to the language

PR #4512, opened 2026-01-19
feat: add line-based diff output for multiline string comparisons

### LINKED ISSUE #4511: [Feature]: Handle multiline string diffs more cleanly
### Problem Statement

The current string diff is great for single-line strings, but when the difference is near newlines, it breaks really bad. For my current project which involves multiline strings as code, it's very hard to make much out of most of the assertion failure text.

### Proposed Solution

The best way to represent that would be to show a line-based diff. For example, consider [Roslyn's diff implementation](https://github.com/dotnet/roslyn/blob/main/src/Compilers/Test/Core/Assert/DiffUtil.cs), which cleanly shows the differences in the strings, since almost always those are multiline.

My proposal is to show the surrounding 2 lines up and down of each changed line segment, and replace the current "expected ... but found ..." text which is completely useless once the diff is shown.

### Alternatives Considered

_No response_

### Feature Category

Assertions

Important - significantly impacts my workflow

### Additional Context

TUnit v1.9.64 and v1.12.3

### Contribution

thomhurst:
## Summary

- Adds `LineDiffHelper` class that generates line-based diffs for multiline string comparisons
- Integrates with `StringEqualsAssertion` to automatically use line diff when strings contain newlines
- Shows unified-diff-style output with line numbers and context instead of character-based pointer

**Before:**
```
Expected to be equal to "line 1\nline 2\nline 3"
but found "line 1\nchanged\nline 3" which differs at index 7:
       ↓
   "line 1\nchanged..."
   "line 1\nline 2..."
       ↑
```

**After:**
```
Expected to be equal to "line 1
line 2
line 3"
but found multiline string with differences starting at line 2:
  1: line 1
- 2: line 2
+ 2: changed
  3: line 3
```

## Features

- Automatic detection of multiline strings (no opt-in required)
- Configurable context lines (default: 2)
- Limits shown differences (default: 3, then "... and N more")
- Handles different line endings (CRLF, LF, CR)
- Supports StringComparison options (e.g., case-insensitive)
- Truncates long lines with ellipsis
- Provides summary for very large inputs (>1000 lines)

Closes #4511

## Test plan

- [x] Run LineDiffHelper unit tests (23 tests)
- [x] Run StringDifference integration tests (10 tests)
- [x] Run StringEqualsAssertion tests (13 tests)
- [ ] Verify output format manually with sample test case

### DIFF
--- a/TUnit.Assertions.Tests/Helpers/StringDifferenceTests.cs
+++ b/TUnit.Assertions.Tests/Helpers/StringDifferenceTests.cs
@@ -77,4 +77,87 @@ public class StringDifferenceTests
         var exception = await Assert.That(sut).ThrowsException();
         await Assert.That(exception.Message.NormalizeLineEndings()).IsEqualTo(expectedMessage);
     }
+
+    [Test]
+    public async Task Multiline_String_Difference_Shows_Line_Diff()
+    {
+        var expected = "line 1\nline 2\nline 3";
+        var actual = "line 1\nchanged\nline 3";
+
+        var sut = async () => await Assert.That(actual).IsEqualTo(expected);
+
+        var exception = await Assert.That(sut).ThrowsException();
+        await Assert.That(exception.Message.NormalizeLineEndings())
+            .Contains("differences starting at line 2")
+            .And.Contains("- 2: line 2")
+            .And.Contains("+ 2: changed");
+    }
+
+    [Test]
+    public async Task Multiline_String_Shows_Context_Lines()
+    {
+        var expected = "line 1\nline 2\nline 3\nline 4\nline 5";
+        var actual = "line 1\nline 2\nchanged\nline 4\nline 5";
+
+        var sut = async () => await Assert.That(actual).IsEqualTo(expected);
+
+        var exception = await Assert.That(sut).ThrowsException();
+        // Should show context lines around the difference
+        await Assert.That(exception.Message.NormalizeLineEndings())
+            .Contains("1: line 1")
+            .And.Contains("2: line 2")
+            .And.Contains("- 3: line 3")
+            .And.Contains("+ 3: changed")
+            .And.Contains("4: line 4")
+            .And.Contains("5: line 5");
+    }
+
+    [Test]
+    public async Task Multiline_String_Handles_Added_Lines()
+    {
+        var expected = "line 1\nline 2";
+        var actual = "line 1\nline 2\nline 3";
+
+        var sut = async () => await Assert.That(actual).IsEqualTo(expected);
+
+        var exception = await Assert.That(sut).ThrowsException();
+        await Assert.That(exception.Message.NormalizeLineEndings())
+            .Contains("+ 3: line 3");
+    }
+
+    [Test]
+    public async Task Multiline_String_Handles_Removed_Lines()
+    {
+        var expected = "line 1\nline 2\nline 3";
+        var actual = "line 1\nline 2";
+
+        var sut = async () => await Assert.That(actual).IsEqualTo(expected);
+
+        var exception = await Assert.That(sut).ThrowsException();
+        await Assert.That(exception.Message.NormalizeLineEndings())
+            .Contains("- 3: line 3");
+    }
+
+    [Test]
+    public async Task Multiline_String_With_CRLF_Line_Endings()
+    {
+        var expected = "line 1\r\nline 2\r\nline 3";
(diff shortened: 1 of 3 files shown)


thomhurst:
## Summary
Adds line-based diff output for multiline string comparisons with context lines, unified diff-style formatting, and configurable limits.

## Critical Issues
None found ✅

## Suggestions

### 1. Public API Snapshot Testing (Minor)
Since `LineDiffHelper` is a new public class in `TUnit.Assertions.Conditions.Helpers` namespace, you may need to verify that public API snapshot tests pass. Per TUnit Rule 2, run:
```bash
dotnet test TUnit.PublicAPI
```
If snapshots changed, commit the `.verified.txt` files (never `.received.txt`).

### 2. Performance Considerations (Optional)
The `LineDiffHelper.GenerateDiff` method is called in `StringEqualsAssertion.BuildStringDifferenceMessage` which is in the assertion failure path. While this isn't a "hot path" in TUnit's sense (discovery/execution), consider:

**TUnit.Assertions/Conditions/Helpers/LineDiffHelper.cs:132-134:**
- `SplitIntoLines` creates intermediate string allocations with `Replace().Replace().Replace().Split()`
- For very large strings, this could be memory-intensive
- Current mitigation: `MaxLines: 1000` default limit is reasonable

**TUnit.Assertions/Conditions/Helpers/LineDiffHelper.cs:143-160:**
- `FindChanges` allocates a `List<LineChange>` and iterates all lines
- Consider: Could short-circuit after finding MaxDifferences to avoid processing large identical files

These are **not blocking** since assertion failure paths don't need micro-optimization, but worth considering for exceptionally large strings.

### 3. String.Concat vs StringBuilder (Very Minor)
**TUnit.Assertions/Conditions/Helpers/LineDiffHelper.cs:330:**
```csharp
return line.Substring(0, maxLength) + "…";
```
This is fine as-is, but for consistency with TUnit's performance-first principle, could use:
```csharp
return string.Concat(line.AsSpan(0, maxLength), "…");
```
(Though this only applies if targeting NET8_0_OR_GREATER, and the current code already has good NET8 optimizations elsewhere.)

## Previous Review Status
No previous comments found.

## Verdict

This is a nice enhancement that significantly improves the UX for multiline string assertion failures. The implementation is well-tested (23 LineDiffHelper tests + 10 StringDifference integration tests) and the automatic fallback to line-based diff is a good design choice. Just verify the public API snapshots pass.

thomhurst:
## Summary
Adds line-based diff output for multiline string comparisons, automatically switching from character-pointer to unified-diff-style output when newlines are detected.

## Critical Issues
None found ✅

## Suggestions

### 1. Test Count Discrepancy
**TUnit.Assertions.Tests/Helpers/StringDifferenceTests.cs**: The PR description mentions "23 LineDiffHelper unit tests" but I only count 6 new integration tests in this file:
- `Multiline_String_Difference_Shows_Line_Diff`
- `Multiline_String_Shows_Context_Lines`
- `Multiline_String_Handles_Added_Lines`
- `Multiline_String_Handles_Removed_Lines`
- `Multiline_String_With_CRLF_Line_Endings`
- `Multiline_String_With_Case_Insensitive_Comparison`

These are integration tests through `StringEqualsAssertion`, not direct unit tests of `LineDiffHelper`. Consider updating the PR description or adding dedicated unit tests for `LineDiffHelper` methods if the 23 tests exist elsewhere.

### 2. Allocation Optimization (Optional)
**TUnit.Assertions/Conditions/Helpers/LineDiffHelper.cs:332**: The `TruncateLine` method uses substring concatenation:
```csharp
return line.Substring(0, maxLength) + "…";
```

For consistency with TUnit's existing code (see `StringEqualsAssertion.cs:218`), consider using NET8_0_OR_GREATER optimization:
```csharp
#if NET8_0_OR_GREATER
    return string.Concat(line.AsSpan(0, maxLength), "…");
#else
    return line.Substring(0, maxLength) + "…";
#endif
```

This is very minor since this code is only in the assertion failure path (not a hot path).

## Previous Review Status
The previous review by thomhurst mentioned public API snapshot testing, but since `LineDiffHelper` is `internal static class`, there are no public API changes and snapshot tests are not required.

## Verdict

The implementation follows TUnit principles: modern C# features, proper StringComparison handling, and good UX improvements. The automatic detection of multiline strings is well-designed. The integration tests adequately cover the feature's behavior through the public API.

PR #4517, opened 2026-01-20
perf: remove unused value and early return in `DedThreadExe`

TimothyMakkison:
- Remove unsued `DedicatedThreadTaskScheduler _taskScheduler` field
- Make `ExecuteAsync` sync
- Make `Task.Factory.StartNew` sync

### Before
image

### After
image


thomhurst:
Looks like ya posted the same image for before and after

TimothyMakkison:
> Looks like ya posted the same image for before and after

Yeah, I've fixed it, but semi abadoned this PR when I realised the savings weren't adding up and I had no idea how to prove that line 82 did anything.

TimothyMakkison:
This should help a little, don't know how much though

PR #4524, opened 2026-01-21
perf: use shared `Task.FromResult(Passed)`

TimothyMakkison:
Use return a shared static `Task.FromResult(Passed)` for successful assertions. This should be safe because `AssertionResult` is a `readonly struct`

### Before
image

### After
image

### DIFF
--- a/TUnit.Assertions/Core/AssertionResult.cs
+++ b/TUnit.Assertions/Core/AssertionResult.cs
@@ -5,6 +5,8 @@ namespace TUnit.Assertions.Core;
 /// </summary>
 public readonly struct AssertionResult
 {
+    internal static readonly Task<AssertionResult> _passedTask = Task.FromResult(Passed);
+
     public bool IsPassed { get; }
     public string Message { get; }
 
--- a/TUnit.Assertions/Assertions/Enums/EnumAssertions.cs
+++ b/TUnit.Assertions/Assertions/Enums/EnumAssertions.cs
@@ -35,7 +35,7 @@ public class HasFlagAssertion<TEnum> : Assertion<TEnum>
 
         if (enumValue.HasFlag(enumFlag))
         {
-            return Task.FromResult(AssertionResult.Passed);
+            return AssertionResult._passedTask;
         }
 
         return Task.FromResult(AssertionResult.Failed($"value {value} does not have flag {_expectedFlag}"));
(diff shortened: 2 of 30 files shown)


TUnit.Assertions/Core/AssertionResult.cs
```diff
@@ -5,6 +5,8 @@ namespace TUnit.Assertions.Core;
 /// </summary>
 public readonly struct AssertionResult
 {
+    internal static Task<AssertionResult> _passedTask = Task.FromResult(Passed);
```
thomhurst: Does this need the read-only keyword, or does it inherit it from being a read-only struct?

thomhurst:
## Summary
This PR introduces a shared static cached `Task<AssertionResult>` for successful assertions to reduce allocations in the hot path.

## Critical Issues

**CRITICAL: Static field initialization bug** ⚠️

The new field in `AssertionResult.cs:8`:
```csharp
internal static Task<AssertionResult> _passedTask = Task.FromResult(Passed);
```

This creates a **static initialization order issue**. The `_passedTask` field references the `Passed` property, but C# doesn't guarantee the order of static field initializers. Since `Passed` is a property (not a field), this should work, but it's fragile.

**More importantly**: The `Passed` property creates a **new instance** on every access:
```csharp
public static AssertionResult Passed => new(true, string.Empty);
```

This means `_passedTask` is initialized with a `Task<AssertionResult>` containing one instance, but later direct uses of `AssertionResult.Passed` create different instances. While this works because `AssertionResult` is a readonly struct with value semantics, it's confusing.

**Recommendation**: Make the initialization explicit and safe:
```csharp
internal static readonly Task<AssertionResult> _passedTask = Task.FromResult(new AssertionResult(true, string.Empty));
```

And add `readonly` modifier for thread safety (static fields should be readonly when immutable).

---

**Minor: Field naming convention**

The field `_passedTask` uses underscore prefix which typically indicates instance fields in C#. For static fields, consider `s_passedTask` or `PassedTask` (if keeping it internal).

## Suggestions

1. **Consider `readonly` modifier**: Since this is a shared static field accessed from multiple threads (assertions run concurrently), marking it `readonly` makes the immutability contract explicit and prevents accidental reassignment.

2. **Performance validation**: The PR shows impressive allocation improvements in the screenshots. Consider adding a comment in the code explaining this optimization for future maintainers.

## Verdict
⚠️ **REQUEST CHANGES** - Fix static initialization to avoid fragile dependency on property access order

TUnit.Assertions/Core/AssertionResult.cs
TimothyMakkison: Good catch, fixed. 

> does it inherit it from being a read-only struct?


Don't think so, `Task` is still a reference type.

TimothyMakkison:
> CRITICAL: Static field initialization bug ⚠️

`Since Passed is a property (not a field), this should work, but it's fragile.`

> Minor: Field naming convention

`Name 'PassedTask' does not match rule 'camel_case_for_private_internal_fields'. Suggested name is '_passedTask'.`

PR #4526, opened 2026-01-21
perf: convert `AfterHookExecutionFlags` to `struct`

TimothyMakkison:
Convert `AfterHookExecutionFlags` to `struct`,

### Before
image

### After
image

### DIFF
--- a/TUnit.Engine/Services/TestLifecycleCoordinator.cs
+++ b/TUnit.Engine/Services/TestLifecycleCoordinator.cs
@@ -92,7 +92,7 @@ internal sealed class TestLifecycleCoordinator
 /// <summary>
 /// Flags indicating which After hooks should be executed based on test completion counts.
 /// </summary>
-internal sealed class AfterHookExecutionFlags
+internal struct AfterHookExecutionFlags
 {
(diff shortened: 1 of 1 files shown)


thomhurst:
## Summary
Converts `AfterHookExecutionFlags` from a reference type (class) to a value type (struct) for performance optimization.

## Critical Issues
None found ✅

## Suggestions

**Performance consideration**: The struct is instantiated in `DecrementAndCheckAfterHooks` (line 58) which is called for every test completion - a hot path. The conversion to struct is good, but consider whether the three bool properties could be represented more efficiently:

- Current: 3 bools = 3 bytes (likely padded to 4+ bytes)
- Alternative: Single `byte` with flags = 1 byte

Example:
```csharp
[Flags]
internal enum AfterHookExecutionFlags : byte
{
    None = 0,
    ShouldExecuteAfterClass = 1 << 0,
    ShouldExecuteAfterAssembly = 1 << 1,
    ShouldExecuteAfterTestSession = 1 << 2
}
```

This would reduce allocation size by ~75% in a hot path. However, if the current approach is clearer and the perf difference is negligible in benchmarks, the readability benefit may outweigh the micro-optimization.

## Verdict

PR #4527, opened 2026-01-21
perf: `IReadOnlyList<Artifact>` call `ToArray`

TimothyMakkison:
Call `ToArray` instead of `ToList`, most of the time the underlying collection contains zero items so it will return an empty array.

### Before
image

### After
image
image

### DIFF
--- a/TUnit.Core/TestContext.Output.cs
+++ b/TUnit.Core/TestContext.Output.cs
@@ -14,7 +14,7 @@ public partial class TestContext
     internal ConcurrentBag<Timing> Timings { get; } = [];
     private readonly ConcurrentBag<Artifact> _artifactsBag = new();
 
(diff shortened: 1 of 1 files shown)


thomhurst:
## Summary
Changes Artifacts property from ToList() to ToArray() to reduce allocations when the collection is empty.

## Critical Issues
None found ✅

## Suggestions

Consider caching to eliminate repeated allocations:

The Artifacts property is accessed multiple times in hot paths (TUnit.Engine/Extensions/TestExtensions.cs:144,146 calls it twice). Each call allocates a new array/list from the ConcurrentBag.

While ToArray() is better for empty collections (returns cached Array.Empty), both ToArray() and ToList() allocate for non-empty collections on every access.

Consider caching if the bag is immutable after test execution starts. This would eliminate all repeated allocations.

## Verdict
✅ APPROVE - The change improves performance for the common case (empty collections).

PR #4528, opened 2026-01-21
perf: fix `_parallelConstraints` deferred initialisation

TimothyMakkison:
`[]` will usually default to a `List<T>` when it is ambiguous, this ensures that an empty array is used when possible.

### Before
image

### After
image

### DIFF
--- a/TUnit.Core/TestContext.Parallelization.cs
+++ b/TUnit.Core/TestContext.Parallelization.cs
@@ -5,7 +5,10 @@ namespace TUnit.Core;
 
 public partial class TestContext
 {
-    internal IReadOnlyList<IParallelConstraint> ParallelConstraints => _parallelConstraints ?? [];
+    internal IReadOnlyList<IParallelConstraint> ParallelConstraints => _parallelConstraints == null
+        ? Array.Empty<IParallelConstraint>()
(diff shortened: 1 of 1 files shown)


TUnit.Core/TestContext.Parallelization.cs
```diff
@@ -5,7 +5,10 @@ namespace TUnit.Core;
 
 public partial class TestContext
 {
-    internal IReadOnlyList<IParallelConstraint> ParallelConstraints => _parallelConstraints ?? [];
+    internal IReadOnlyList<IParallelConstraint> ParallelConstraints => _parallelConstraints == null
+        ? Array.Empty<IParallelConstraint>()
+        : _parallelConstraints;
+
```
thomhurst, line 11: Can we keep it with a null coalescing operator?

```suggestion
    internal IReadOnlyList<IParallelConstraint> ParallelConstraints => _parallelConstraints ?? Array.Empty<IParallelConstraint>();
```

thomhurst:
This feels like the compiler should have logic to use empty arrays in these scenarios! I was under the impression `[]` abstracted a lot of the stuff away and would choose the best type for you

PR #4532, opened 2026-01-21
perf: `ThreadSafeDict` extract slow logic

TimothyMakkison:
C# compiler like to eagerly initialise closures at the start of the method, so even on fast paths the display class for `valueFactory` was being created. Extracting the slow code to a new method solves this issue. 

I have no idea why the .NET compiler does this, I assume it's a correctness thing (sometimes due `if` `else` usage). I really wish it didn't do this.

### Before
image

### After
image

### DIFF
--- a/TUnit.Core/Data/ThreadSafeDictionary.cs
+++ b/TUnit.Core/Data/ThreadSafeDictionary.cs
@@ -81,6 +81,11 @@ public class ThreadSafeDictionary<TKey,
             return existingLazy.Value;
         }
 
+        return GetOrAddSlow(key, func);
+    }
+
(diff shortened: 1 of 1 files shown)


thomhurst:
## Summary
Extracts slow path logic in ThreadSafeDictionary.GetOrAdd to a separate method to prevent eager closure allocation on the fast path.

## Critical Issues
None found ✅

## Suggestions
None - this is a clean micro-optimization that avoids unnecessary allocations on the hot path (key exists case) without changing behavior.

## Verdict

---

**Review Notes:**
- Performance optimization aligns with TUnit's "Performance First" principle
- No public API changes, so no snapshot updates needed
- Change is localized to a single method with clear intent
- The extracted method preserves the original slow-path logic exactly

PR #4538, opened 2026-01-22
perf: change `TrackedObjects` to `Dictionary`

TimothyMakkison:
Change `TrackedObjects` to `Dictionary` 

### Why I think this is safe
The only function that mutates `TrackedObjects` is `ObjectTracker.TrackObjects` which is in turn called by the following methods:

- `TestArgumentRegistrationService.RegisterTestArgumentsAsync` 
	- `TestBuilder.InvokeTestRegisteredEventReceiversAsync`
		- `TestBuilder.BuildTestAsync` * 4
	- `TestFilterService.RegisterTest`
		- `TestDiscoveryService.DiscoverTests` * 2
			- Both go to both branches of `TestRequestHandler.HandleRequestAsync`

Note that when a method only has 1 caller I skip it. To my knowledge `TrackedObjects` is only ever mutated by one thread at a time and usually only once.

### Before
image
image

### After
image
image

I'm pretty sure I can simplify a lot of logic in `ObjectGraphDiscoverer` as the function signatures are similar now

### DIFF
--- a/TUnit.Core/Discovery/ObjectGraphDiscoverer.cs
+++ b/TUnit.Core/Discovery/ObjectGraphDiscoverer.cs
@@ -110,7 +110,7 @@ internal sealed class ObjectGraphDiscoverer : IObjectGraphTracker
                 return false;
             }
 
-            AddToDepth(objectsByDepth, depth, obj);
+            TryAddToHashSet(objectsByDepth, depth, obj);
 
             return true;
         }
@@ -133,7 +133,7 @@ internal sealed class ObjectGraphDiscoverer : IObjectGraphTracker
 
         if (visitedObjects.Add(rootObject))
         {
-            AddToDepth(objectsByDepth, 0, rootObject);
+            TryAddToHashSet(objectsByDepth, 0, rootObject);
 
             DiscoverNestedObjects(rootObject, objectsByDepth, visitedObjects, currentDepth: 1, cancellationToken);
         }
@@ -148,7 +148,7 @@ internal sealed class ObjectGraphDiscoverer : IObjectGraphTracker
     /// <param name="testContext">The test context to discover objects from.</param>
     /// <param name="cancellationToken">Cancellation token for the operation.</param>
     /// <returns>The tracked objects dictionary (same as testContext.TrackedObjects).</returns>
-    public ConcurrentDictionary<int, HashSet<object>> DiscoverAndTrackObjects(TestContext testContext, CancellationToken cancellationToken = default)
+    public Dictionary<int, HashSet<object>> DiscoverAndTrackObjects(TestContext testContext, CancellationToken cancellationToken = default)
     {
         var visitedObjects = testContext.TrackedObjects;
 
@@ -188,7 +188,7 @@ internal sealed class ObjectGraphDiscoverer : IObjectGraphTracker
                 return false;
             }
 
-            AddToDepth(objectsByDepth, depth, value);
+            TryAddToHashSet(objectsByDepth, depth, value);
 
             return true;
         }
@@ -212,7 +212,7 @@ internal sealed class ObjectGraphDiscoverer : IObjectGraphTracker
     /// </summary>
     private void DiscoverNestedObjectsForTracking(
         object obj,
-        ConcurrentDictionary<int, HashSet<object>> visitedObjects,
+        Dictionary<int, HashSet<object>> visitedObjects,
         int currentDepth,
         CancellationToken cancellationToken)
     {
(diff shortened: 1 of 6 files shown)


thomhurst:
## Summary
Changes TrackedObjects from ConcurrentDictionary to Dictionary to improve performance.

## Critical Issues

BLOCKING: Thread-Safety Violation

The PR removes thread-safe access to TrackedObjects, but tests are built in parallel:

Evidence of parallel execution:
- TestBuilderPipeline.BuildTestsFromMetadataAsync:L205 uses .ProcessInParallel(Environment.ProcessorCount)
- Each parallel task calls RegisterTestArgumentsAsync -> TrackObjects -> mutates TrackedObjects

The race condition:

1. ObjectGraphDiscoverer.TryAddToHashSet (modified in PR):
   - BEFORE: Used ConcurrentDictionary with lock on HashSet
   - AFTER: Uses plain Dictionary with NO lock on HashSet.Add()

2. ObjectTracker.FlattenTrackedObjects (TUnit.Core/Tracking/ObjectTracker.cs:56-77) still expects thread-safety by locking kvp.Value

3. The bugs:
   - Multiple threads calling dict.GetOrAdd on plain Dictionary -> undefined behavior
   - HashSet.Add() without lock -> corruption (HashSet is not thread-safe)
   - Reading dictionary while being mutated -> possible exceptions or lost data

Why the PR author's analysis is incorrect:
The PR description traces the call chain but misses that BuildTestsFromMetadataAsync fans out to multiple threads at line 205. Each thread gets its own TestContext but if any two threads happen to add objects at the same depth, they will be calling GetOrAdd and HashSet.Add concurrently.

Required Fix:
Option 1: Revert to ConcurrentDictionary with locks (safest)
Option 2: If optimization is needed, ensure ALL access paths are single-threaded with proper documentation and locks

## Verdict
REQUEST CHANGES - Critical thread-safety violation will cause non-deterministic failures

TimothyMakkison:
> Evidence of parallel execution:

    TestBuilderPipeline.BuildTestsFromMetadataAsync:L205 uses .ProcessInParallel(Environment.ProcessorCount)
    Each parallel task calls RegisterTestArgumentsAsync -> TrackObjects -> mutates TrackedObjects

Right, tests are built in parallel however multiple threads don't build a given test. From a tests point of view, it's build logic is built sequentially using async. AFAIK `TrackedObjects ` is only mutated once or twice when a test is built and this is never done in parallel by multiple threads.

> `ObjectTracker.FlattenTrackedObjects`

Removed the lock

thomhurst:
## Summary
Changes TrackedObjects from ConcurrentDictionary to Dictionary to improve performance by removing unnecessary thread-safety overhead.

## Critical Issues
Thread-Safety Violation Risk

The change from ConcurrentDictionary to Dictionary appears unsafe based on code analysis:

1. Concurrent Access Pattern in ObjectLifecycleService.cs:219-257: The InitializeTrackedObjectsAsync method reads from trackedObjects (line 222) and iterates over keys (lines 227-231) without external synchronization. While the method uses locks when copying individual HashSets (lines 240-243), there is no lock protecting the Dictionary itself during the initial read operations (.Count, .Keys.CopyTo).

2. Potential Race Condition: If test discovery runs in parallel (which appears possible based on Task.WhenAll in TestBuilder), multiple threads could be calling ObjectTracker.TrackObjects which modifies TrackedObjects via GetOrAdd (ObjectGraphDiscoverer.cs:277-278 in the PR version), while another thread in InitializeTrackedObjectsAsync reads the dictionary.

3. Dictionary.GetOrAdd Extension Method: The PR changes line 277 from using ConcurrentDictionary.GetOrAdd (thread-safe) to using a Dictionary.GetOrAdd extension method. This extension method cannot provide thread-safety for the underlying Dictionary.

## Verification Needed

To determine if this change is safe, please verify:

1. Discovery Threading Model: Does test discovery ever run in parallel? If TestBuilder.BuildTestAsync can be called concurrently for different tests, then TrackedObjects may be written by multiple threads simultaneously.

2. Discovery vs Execution Separation: The PR description states "to my knowledge TrackedObjects is only ever mutated by one thread at a time." Can you confirm that test discovery completes entirely (single-threaded) before any test execution begins, OR each TestContext has isolated access to its TrackedObjects during both discovery and execution?

3. Thread-Safe Extension Method: If Dictionary.GetOrAdd in your codebase has special synchronization, please point to its implementation.

## Suggestions
If the threading concern is addressed, consider documenting the thread-safety assumptions in the code with a comment explaining why Dictionary is safe here.

## Verdict
REQUEST CHANGES - Need confirmation that the Dictionary access is actually single-threaded, or the change risks concurrent modification exceptions.

TimothyMakkison:
> Concurrent Access Pattern in ObjectLifecycleService.cs:219-257: The InitializeTrackedObjectsAsync method reads from trackedObjects (line 222) and iterates over keys (lines 227-231) without external synchronization. While the method uses locks when copying individual HashSets (lines 240-243), there is no lock protecting the Dictionary itself during the initial read operations (.Count, .Keys.CopyTo).

Nice, missed that I've removed the `Lock`, this should save another 400KB in `object[]`


>  If test discovery runs in parallel (which appears possible based on Task.WhenAll in TestBuilder), multiple threads could be calling ObjectTracker.TrackObjects 

See above comment, tests are built in parallel but are individually sequential. All calls that mutate or may mutate are awaited so they shouldn't intefere with one another.


> Dictionary.GetOrAdd Extension Method: 

`TUnit` uses an extension method for better semantics, this is basically syntax sugar

PR #4542, opened 2026-01-23
fix: prevent IndexOutOfRangeException in MethodDataSource analyzer

### LINKED ISSUE #4540: [Bug]: MethodDataSource is not functioning correctly
### Description

Hello, I am leaving screenshots; you can see the errors there.

Image

Image

### Expected Behavior

So that the errors pass

### Actual Behavior

Code doesn't compile

### Steps to Reproduce

Visible in screenshot

### TUnit Version

1.12.15

### .NET Version

.NET 10

### Operating System

Windows

### IDE / Test Runner

JetBrains Rider

```shell

```

### Additional Context

_No response_

### IDE-Specific Issue?

thomhurst:
## Summary
- Fix `IndexOutOfRangeException` crash in `TestDataAnalyzer.CheckMethodDataSource()` when attribute constructor arguments are missing or cannot be resolved
- Add defensive checks to handle malformed attributes gracefully

## Root Cause
The analyzer was accessing `ConstructorArguments[0]` and `ConstructorArguments[1]` without checking if the array had elements. This caused crashes when:
- The attribute's constructor arguments couldn't be resolved during compilation
- There were other compilation errors in the user's code
- The attribute was malformed

The exception was being caught by the analyzer framework and converted into a cryptic `TUnit0001` error message.

## Test plan
- [x] Verify the analyzer no longer crashes with missing constructor arguments
- [ ] Run existing analyzer tests to ensure no regression

Fixes #4540

### DIFF
--- a/TUnit.Analyzers/TestDataAnalyzer.cs
+++ b/TUnit.Analyzers/TestDataAnalyzer.cs
@@ -410,12 +410,24 @@ public class TestDataAnalyzer : ConcurrentDiagnosticAnalyzer
         IPropertySymbol? propertySymbol = null)
     {
         {
+            // Guard against missing constructor arguments (can happen during compilation errors)
+            if (attribute.ConstructorArguments.IsDefaultOrEmpty)
+            {
+                return;
+            }
+
             var type = attribute.AttributeClass?.IsGenericType == true
                 ? attribute.AttributeClass.TypeArguments.First()
                 : attribute.ConstructorArguments[0].Value as INamedTypeSymbol ?? testClassType;
 
             var methodName = attribute.ConstructorArguments[0].Value as string
-                         ?? attribute.ConstructorArguments[1].Value as string;
+                         ?? (attribute.ConstructorArguments.Length > 1 ? attribute.ConstructorArguments[1].Value as string : null);
+
+            // If method name couldn't be resolved, skip validation (compilation error will be reported separately)
+            if (string.IsNullOrEmpty(methodName))
+            {
+                return;
+            }
 
             var argumentsNamedArgument = attribute.NamedArguments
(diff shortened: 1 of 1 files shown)


thomhurst:
## Summary
Fixes crash in analyzer when MethodDataSource attribute has malformed or unresolved constructor arguments during compilation.

## Critical Issues
None found ✅

## Suggestions

**Consider adding a length check for ConstructorArguments[0] access (line 415)**

While the PR fixes the `IndexOutOfRangeException` for `ConstructorArguments[1]`, it still accesses `ConstructorArguments[0]` without verifying the array has at least one element (line 415 in the diff context):

```csharp
var type = attribute.AttributeClass?.IsGenericType == true
    ? attribute.AttributeClass.TypeArguments.First()
    : attribute.ConstructorArguments[0].Value as INamedTypeSymbol ?? testClassType;
```

The guard check on line 413-416 (after the fix) returns early if the array is empty, which protects line 415. However, for defensive coding and clarity, you might want to reorder the logic to check the array before any access, or add an explicit length check before line 415.

**Example:**
```csharp
if (attribute.ConstructorArguments.IsDefaultOrEmpty || attribute.ConstructorArguments.Length == 0)
{
    return;
}

var type = attribute.AttributeClass?.IsGenericType == true
    ? attribute.AttributeClass.TypeArguments.First()
    : (attribute.ConstructorArguments.Length > 0 ? attribute.ConstructorArguments[0].Value as INamedTypeSymbol : null) ?? testClassType;
```

This is a minor suggestion - the current fix does protect against the crash through the early return.

## Verdict

The fix appropriately addresses the reported bug by adding defensive checks for missing constructor arguments. The analyzer will now gracefully skip validation when attributes are malformed due to compilation errors, allowing the compiler to report the actual error instead of causing an analyzer crash.

PR #4543, opened 2026-01-23
perf: use static `Task` in `MethodAssertionGenerator` generated code

TimothyMakkison:
Improve generated code by `MethodAssertionGenerator` to sometimes emit `private static readonly Task<AssertionResult> _passedTask = Task.FromResult(AssertionResult.Passed);` and return it.

- This could be a `protected static readonly Task<AssertionResult> _passedTask` inherited from `Assertion<T>`, but I didn't want to change the existing api.
- It's been a while since I've worked on a source generator, but could this one use caching?

Again I'll ask you to update the snapshot tests, I've rebased and I still get the Sourcy error 😅 

### Before
image

### After
image

### DIFF
--- a/TUnit.Assertions.SourceGenerator/Generators/MethodAssertionGenerator.cs
+++ b/TUnit.Assertions.SourceGenerator/Generators/MethodAssertionGenerator.cs
@@ -569,6 +569,13 @@ public sealed class MethodAssertionGenerator : IIncrementalGenerator
 
         sb.AppendLine("{");
 
+        if (data.ReturnTypeInfo.Kind is ReturnTypeKind.Bool)
+        {
+            sb.AppendLine(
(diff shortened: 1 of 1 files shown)


thomhurst:
## Summary
Optimizes generated code by caching the successful assertion result task in a static field.

## Critical Issues

Missing snapshot test updates - The PR modifies source generator output but does not include updated verified.txt files. TUnit CRITICAL RULE 2 requires snapshot tests to be updated when source generator output changes.

The generated code will now include a static field and changed return logic affecting MethodAssertionGeneratorTests snapshot files.

Required: Run snapshot tests, review changes, update verified.txt files and commit them.

## Suggestions

The optimization follows TUnit Performance First principle. Caching the success case is good. The failure case correctly allocates new Task for unique error messages.

## Verdict
REQUEST CHANGES - Missing required snapshot test updates

PR #4550, opened 2026-01-24
fix: actually load assemblies in InfrastructureGenerator

### LINKED ISSUE #4541: [Bug]: The method marked with [Before (TestSession)] never runs
### Description

After upgrading TUnit from 0.19.2 to the latest version, this issue occurred where the method marked as [Before (TestSession)] never ran. Before all testing, I used this method to start a postgres container for the testing environment, but it never executed, causing the test to fail. After the breakpoint display test starts, the method will be skipped directly and breakpoints located within the method will not be triggered

### Expected Behavior

The method marked as [Before (TestSession)] runs before the start of the test session

### Actual Behavior

Marked as [Before (TestSession)], it has never been run from the beginning to the end of the test

### Steps to Reproduce

Here is a simple test example I used:

```
public sealed class AlbaBootstrap : IntegrationTestBase, IAsyncInitializer, IAsyncDisposable
{
    public IAlbaHost Host { get; private set; } = null!;

    public async Task InitializeAsync() 
    {
        Host = await AlbaHost.For<WebApp.Program>();
    }

    public async ValueTask DisposeAsync()
    {
        await Host.DisposeAsync();
    }
}
```

```
public abstract class AlbaTestBase(AlbaBootstrap albaBootstrap)
{
    protected IAlbaHost Host => albaBootstrap.Host;
}

[ClassDataSource<AlbaBootstrap>(Shared = SharedType.PerTestSession)]
public class MyTestClass(AlbaBootstrap albaBootstrap) : AlbaTestBase(albaBootstrap)
{
    [Test]
    public async Task happy_path()
    {
        await Host.Scenario(_ =>
        {
            _.Get.Url("/fake/okay");
            _.StatusCodeShouldBeOk();
        });
    }
}
```

The following is the code for the base class IntegrationTestBase, where the method marked with [Before (TestSession)] is located:

```
public abstract class IntegrationTestBase
{
    private static RabbitMqContainer? rabbitMqContainer;
    private static PostgreSqlContainer? pgSqlContainer;
    protected static string? RabbitMqConnString { get; private set; }
    protected static string? PgSqlConnString { get; private set; }

    [Before(TestSession)]
    public static async Task SessionSetup()
    {
        // db setup
        rabbitMqContainer = new RabbitMqBuilder("rabbitmq:latest").Build();
        pgSqlContainer = new PostgreSqlBuilder("postgres:latest")
            .WithUsername("test")
            .WithPassword("test")
            .WithDatabase("testDB")
            .WithCleanUp(true)
            .Build();
        await rabbitMqContainer.StartAsync();
        await pgSqlContainer.StartAsync();
        PgSqlConnString = pgSqlContainer.GetConnectionString();
        RabbitMqConnString = rabbitMqContainer.GetConnectionString();
    }

    [After(TestSession)]
    public static async Task SessionCleanup()
    {
        if (rabbitMqContainer is not null)
            await rabbitMqContainer.DisposeAsync();

        if (pgSqlContainer is not null)
            await pgSqlContainer.DisposeAsync();
    }
}
```

### TUnit Version

1.12.15

### .NET Version

.Net 10

### Operating System

Windows

### IDE / Test Runner

```shell

```

### Additional Context

none of them

### IDE-Specific Issue?

thomhurst:
## Problem

Hooks (`[Before(TestSession)]`, `[After(TestSession)]`, etc.) in referenced library assemblies were not executing, even though the source generator correctly generated hook registration code with ModuleInitializers.

Reported in #4541

## Root Cause

The `InfrastructureGenerator` has a critical bug on line 264:

```csharp
sourceBuilder.AppendLine($"global::TUnit.Core.SourceRegistrar.RegisterAssembly(() => global::System.Reflection.Assembly.Load(\"{assemblyName}\"));");
```

This queues a loader function to `Sources.AssemblyLoaders`, but **this queue is never processed anywhere in the codebase**. The assemblies are never actually loaded, so their ModuleInitializers (which register hooks) never execute.

### The Flow

1. ✅ Source generator runs in library and generates hooks with `[ModuleInitializer]`
2. ✅ `InfrastructureGenerator` detects referenced assemblies with TUnit.Core
3. ❌ It queues loaders via `RegisterAssembly()` but they're never invoked
4. ❌ Library assembly never loads → ModuleInitializers never run → Hooks never register

## Solution

Changed the generated code to **directly load assemblies** in the ModuleInitializer:

```csharp
sourceBuilder.AppendLine($"_ = global::System.Reflection.Assembly.Load(\"{assemblyName}\");");
```

Now assemblies are loaded immediately during module initialization, triggering their ModuleInitializers which register hooks globally.

## Testing

- ✅ Added `AbstractTestSessionHookTests.cs` to verify hooks in abstract base classes work
- ✅ Verified hooks from `TUnit.TestProject.Library` execute correctly
- ✅ User's reproduction case now passes with this fix

## Impact

This fixes a long-standing bug where hooks in library projects referenced by test projects would silently fail to execute. Users organizing their test infrastructure into separate libraries can now use hooks as expected.

Closes #4541


thomhurst:
## Summary
Fixes hooks in library assemblies by directly loading assemblies in the ModuleInitializer instead of queueing them to an unprocessed queue.

## Critical Issues
None found ✅

## Analysis

The fix correctly addresses the root cause identified in #4541:

**Problem:** Hooks defined in referenced library assemblies weren't executing because:
1. The source generator generated code to queue assembly loaders: `SourceRegistrar.RegisterAssembly(() => Assembly.Load(...))`
2. This queue (`Sources.AssemblyLoaders`) is **never processed** anywhere in the codebase (verified - no `.Dequeue()` calls exist)
3. Assemblies never load → ModuleInitializers never run → Hooks never register

**Solution:** Changed to directly load assemblies:
```csharp
_ = global::System.Reflection.Assembly.Load("{assemblyName}");
```

This triggers ModuleInitializers immediately, allowing hooks to register.

**TUnit Rules Compliance:**
- ✅ No dual-mode concerns (source-gen only change)
- ✅ No snapshot tests need updating (InfrastructureGenerator output not snapshot tested)
- ✅ No public API changes
- ✅ AOT compatible (Assembly.Load is reflection, but guarded by `RuntimeFeature.IsDynamicCodeSupported` check in SourceRegistrar.cs:27-30)
- ✅ No VSTest usage
- ✅ Performance: Minimal impact - assemblies load during module init (one-time cost)

**Testing:**
- New test `AbstractTestSessionHookTests.cs` validates the fix
- Tests hooks on abstract base classes (the reported scenario)

## Verdict
✅ **APPROVE** - Clean fix for a critical bug affecting library-based test organization.

thomhurst:
## Summary
Fixes hooks in library assemblies by directly loading assemblies in ModuleInitializer instead of queueing them to an unprocessed queue.

## Critical Issues
**Silent Exception Swallowing** - The try-catch block at InfrastructureGenerator.cs:265-275 silently swallows all exceptions with only a comment. This could hide assembly name typos, missing dependencies, version conflicts, or security issues.

**Impact:** Users won't know why hooks aren't executing if assembly loading fails.

**Recommendation:** Either fail fast with AggregateException or only catch specific expected exceptions like FileNotFoundException or BadImageFormatException.

## Previous Review Status
The previous review approved the fix, correctly identifying that it solves the root cause. This concern about exception handling is an additional consideration not covered in that review.

## Verdict
REQUEST CHANGES - The fix is correct, but silent exception swallowing could mask real issues.

PR #4553, opened 2026-01-24
fix: flush console output after test completion to prevent truncation (#4545)

### LINKED ISSUE #4545: [Bug]: Log outputs when running multiple tests in parallel are either missing or incomplete

thomhurst:
## Problem

Console output was being truncated when tests completed, particularly when tests ended with `Console.Write()` without a newline. Users reported that "the end of the output is cut" even after version 1.12.41 (which included the previous fix in c01a6fc86).

**Example scenario:**
```csharp
Console.WriteLine("Start");
Console.Write("End (no newline)"); // This output was lost!
```

## Root Cause

While commit c01a6fc86 fixed **parallel output mixing** by using per-context buffers, it didn't address **output truncation** at test completion.

When a test ends, any buffered console output (from `Console.Write()` without newline) remains in the test context's line buffer but is never flushed to the sinks. This buffered output is lost and doesn't appear in test results or IDE output.

## Solution

Added explicit flushing of console interceptors (stdout and stderr) in `TestCoordinator.ExecuteTestInternalAsync()` finally block:

```csharp
try
{
    await Console.Out.FlushAsync().ConfigureAwait(false);
    await Console.Error.FlushAsync().ConfigureAwait(false);
}
catch (Exception flushEx)
{
    await _logger.LogErrorAsync($"Error flushing console output for {test.TestId}: {flushEx}").ConfigureAwait(false);
}
```

This ensures all buffered output is captured before test results are reported.

**Why this location?**
- ✅ Runs after test execution completes
- ✅ Runs in finally block (always executes, even on exception)
- ✅ Runs BEFORE result reporting (so output is available)
- ✅ Runs while `TestContext.Current` still points to the test
- ✅ Handles both stdout and stderr
- ✅ Error handling prevents flush failures from breaking tests

## Testing

- **Added** `OutputTruncationTests.cs` - Tests specifically for output ending without newline
- **Enhanced** `ParallelConsoleOutputTests.cs` - Added Test4 for truncation scenario
- **Existing tests** - All parallel output isolation tests continue to pass

## Impact

- ✅ Tests ending with `Console.Write()` now capture all output
- ✅ No output loss in Visual Studio Test Explorer
- ✅ Works with both parallel and `[NotInParallel]` tests
- ✅ Maintains existing parallel output isolation

Fixes #4545

### DIFF
--- a/TUnit.Engine/Services/TestExecution/TestCoordinator.cs
+++ b/TUnit.Engine/Services/TestExecution/TestCoordinator.cs
@@ -150,6 +150,18 @@ internal sealed class TestCoordinator : ITestCoordinator
         {
             var cleanupExceptions = new List<Exception>();
 
+            // Flush console interceptors to ensure all buffered output is captured
+            // This is critical for output from Console.Write() without newline
+            try
+            {
+                await Console.Out.FlushAsync().ConfigureAwait(false);
+                await Console.Error.FlushAsync().ConfigureAwait(false);
+            }
+            catch (Exception flushEx)
+            {
+                await _logger.LogErrorAsync($"Error flushing console output for {test.TestId}: {flushEx}").ConfigureAwait(false);
+            }
+
             await _objectTracker.UntrackObjects(test.Context, cleanupExceptions).ConfigureAwait(false);
 
             var testClass = test.Metadata.TestClassType;
--- a/TUnit.TestProject/Bugs/Issue4545/OutputTruncationTests.cs
+++ b/TUnit.TestProject/Bugs/Issue4545/OutputTruncationTests.cs
@@ -1,0 +1,40 @@
+namespace TUnit.TestProject.Bugs.Issue4545;
+
+/// <summary>
+/// Tests to reproduce the issue where console output is truncated/missing
+/// when tests complete, especially when they end with Console.Write (no newline).
+/// </summary>
+[NotInParallel]
+public class OutputTruncationTests
+{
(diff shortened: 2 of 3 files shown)


thomhurst:
## Summary
Adds console output flushing in the test cleanup path to prevent loss of buffered output from `Console.Write()` calls without newlines.

## Critical Issues
None found ✅

## Suggestions
None - the implementation is solid:
- Correctly placed early in the `finally` block (before message bus reporting)
- Proper async/await with `ConfigureAwait(false)`
- Error handling prevents flush failures from breaking tests
- Good test coverage for the fix

The fix addresses the root cause described in the PR: buffered output (from `Console.Write()` without newline) now gets flushed before test results are reported, ensuring all console output is captured.

## Verdict

PR #4572, opened 2026-01-27
fix: pass empty array to params parameter when [Arguments] has no values

### LINKED ISSUE #4561: [Bug]: `params` test argument is null when `Arguments` list is empty
### Description

When the list of arguments to the `Arguments` attribute is empty, and the test has a `params` parameter, the test is invoked with `null` rather than an empty array.

Consider the following example:
```csharp
[Test]
[Arguments]            // this one will fail, following ones will pass
[Arguments("a")]
[Arguments("a", "B")]
[Arguments("a", "b", "c")]
public async Task SomeTest(params string[] args)
{
    // do some test using the "args" array
    // if no arguments are provided, expect it to be empty, not null 
    await Assert.That(args).IsNotNull();
}
```

### Expected Behavior

Expected the test to be invoked with an empty array - the standard C# behavior when a method with a `params` argument is invoked with no arguments.

### Actual Behavior

The test's `params` argument is `null`

### Steps to Reproduce

Run the following test. It should never fail, but it does.

```csharp
[Test]
[Arguments]
[Arguments("a")]
[Arguments("a", "B")]
[Arguments("a", "b", "c")]
public async Task SomeTest(params string[] args)
{
    // do some test using the "args" array
    // if no arguments are provided, expect it to be empty, not null 
    await Assert.That(args).IsNotNull();
}
```

### TUnit Version

1.12.15

### .NET Version

.NET 9.0

### Operating System

Windows

### IDE / Test Runner

JetBrains Rider

```shell

```

### Additional Context

No AOT / trimming.

### IDE-Specific Issue?

thomhurst:
## Summary

- Fixed bug where `[Arguments]` with no values passed `null` instead of an empty array to `params` parameters
- The `ArgumentsAttribute` constructor now distinguishes between empty arguments (`[]`) and explicit null (`[null]`)
- Added test case to verify the fix

Fixes #4561

## Test plan

- [x] Verified new test `ParamsOnlyWithEmptyArguments` passes with 4 test cases
- [x] Verified all 12 `ParamsArgumentsTests` pass
- [x] Verified 105 source generator tests pass
- [x] No snapshot test failures

### DIFF
--- a/TUnit.Core/Attributes/TestData/ArgumentsAttribute.cs
+++ b/TUnit.Core/Attributes/TestData/ArgumentsAttribute.cs
@@ -62,9 +62,13 @@ public sealed class ArgumentsAttribute : Attribute, IDataSourceAttribute, ITestR
 
     public ArgumentsAttribute(params object?[]? values)
     {
-        if (values == null || values.Length == 0)
+        if (values == null)
         {
             Values = [null];
+        }
+        else if (values.Length == 0)
+        {
+            Values = [];
         }
         else
         {
--- a/TUnit.TestProject/ParamsArgumentsTests.cs
+++ b/TUnit.TestProject/ParamsArgumentsTests.cs
@@ -45,6 +45,17 @@ public class ParamsArgumentsTests
     }
 
     [Test]
+    [Arguments]
+    [Arguments("a")]
+    [Arguments("a", "b")]
+    [Arguments("a", "b", "c")]
+    public async Task ParamsOnlyWithEmptyArguments(params string[] args)
+    {
+        // When [Arguments] has no values, params should be an empty array, not null
+        await Assert.That(args).IsNotNull();
+    }
+
+    [Test]
     [Arguments(1, "single")]
     public async Task SingleStringInParamsArray(int id, params string[] values)
     {


thomhurst:
## Summary
Fixes bug where `[Arguments]` with no values passed `null` instead of an empty array to `params` parameters.

## Critical Issues
None found ✅

## Analysis

**Changes Review:**
The fix correctly distinguishes between three cases in the `ArgumentsAttribute` constructor (TUnit.Core/Attributes/TestData/ArgumentsAttribute.cs:63-73):
1. `null` passed → wraps as `[null]` (explicit null value)
2. Empty array `[]` → keeps as `[]` (empty params)
3. Non-empty array → uses values as-is

This aligns with standard C# behavior where calling a method with a `params` parameter and no arguments passes an empty array, not null.

**Test Coverage:**
The new test `ParamsOnlyWithEmptyArguments` (TUnit.TestProject/ParamsArgumentsTests.cs:47-58) validates the fix with 4 test cases covering the empty, single, double, and triple argument scenarios.

**TUnit Rules Compliance:**
- ✅ **No dual-mode concerns** - This is a runtime attribute change, not source generator output modification
- ✅ **No snapshot test updates needed** - No source generator output changed (verified PR description: "No snapshot test failures")
- ✅ **No public API changes** - Internal behavior fix only
- ✅ **No reflection concerns** - No new reflection usage
- ✅ **No performance concerns** - Single conditional check, minimal overhead
- ✅ **No VSTest usage**

**Correctness:**
The logic is sound. The original code treated both `null` and empty array identically (wrapping as `[null]`), which violated C# params semantics. The fix preserves the ability to pass an explicit null value via `ArgumentsAttribute(null)` while correctly handling the no-arguments case `ArgumentsAttribute()`.

## Verdict

PR #4585, opened 2026-01-28
perf: remove `CodeWriter` bad initialisation

TimothyMakkison:
Each time `CiodeWriter` is intialised it will try to pre seed `_indentCache` with different length strings.
- This isn't really needed
- Most code writers are one of two different `indentString`
- This would alllocated 10 different `Enumerable.RepeatIterator` and `string` making this an anti optimisation


thomhurst:
## Summary
Removes unnecessary pre-seeding of the static indent cache in CodeWriter constructor.

## Critical Issues
None found ✅

## Suggestions
None - this is a good optimization.

Analysis: The removed code was pre-seeding a static readonly ConcurrentDictionary with 10 entries on every CodeWriter instantiation. Since the cache is static and shared across all instances, most of these TryAdd calls would fail after the first instance. The GetIndentation method (line 54-58) already handles lazy initialization via GetOrAdd, making this pre-seeding redundant.

This removes allocations from the source generator hot path (10 Enumerable.RepeatIterator + 10 strings per constructor call).

## Verdict
✅ APPROVE - Valid performance optimization with no behavioral changes.

TimothyMakkison:
Ran the benchmarks and this saves 10MB


### Before (timing is inaccurate, laptop not plugged in and ran once.

| Method  | Mean     | Error    | StdDev   | Gen0       | Gen1      | Allocated |
|-------- |---------:|---------:|---------:|-----------:|----------:|----------:|
| Compile | 499.5 ms | 19.50 ms | 57.48 ms | 18000.0000 | 4000.0000 | 165.61 MB |



### After
| Method  | Mean     | Error    | StdDev   | Gen0       | Gen1      | Allocated |
|-------- |---------:|---------:|---------:|-----------:|----------:|----------:|
| Compile | 504.7 ms | 27.82 ms | 80.25 ms | 17000.0000 | 4000.0000 | 155.99 MB |

PR #4624, opened 2026-01-31
chore: add `AotConverterGeneratorBenchmarks`

TimothyMakkison:
Add `AotConverterGeneratorBenchmarks` 

| Method       | Mean     | Error    | StdDev  | Gen0       | Gen1      | Allocated |
|------------- |---------:|---------:|--------:|-----------:|----------:|----------:|
| RunGenerator | 576.4 ms | 10.45 ms | 8.72 ms | 11000.0000 | 3000.0000 | 101.98 MB |


TimothyMakkison:
There might be an issue when benchmarking source generators, it appears that all generators withtin `TUnit.Core.SourceGenerator` are running each time you run the benchmark. I'm not sure why.


Edit: I think this is only an issue when your profile the application

PR #4629, opened 2026-01-31
feat: make `AotConverterGenerator` incremental and optimise

TimothyMakkison:
- Add system to make `AotConverterGenerator` incremental
- Descend root node once instead of descending twice

### Before

| Method       | Mean     | Error    | StdDev  | Gen0       | Gen1      | Allocated |
|------------- |---------:|---------:|--------:|-----------:|----------:|----------:|
| RunGenerator | 576.4 ms | 10.45 ms | 8.72 ms | 11000.0000 | 3000.0000 | 101.98 MB |

### After

| Method       | Mean     | Error   | StdDev   | Median   | Gen0       | Gen1      | Allocated |
|------------- |---------:|--------:|---------:|---------:|-----------:|----------:|----------:|
| RunGenerator | 329.8 ms | 5.83 ms | 13.04 ms | 325.1 ms | 11000.0000 | 3000.0000 | 101.19 MB |

### DIFF
--- a/TUnit.Core.SourceGenerator/Generators/AotConverterGenerator.cs
+++ b/TUnit.Core.SourceGenerator/Generators/AotConverterGenerator.cs
@@ -673,9 +673,9 @@ public class AotConverterGenerator : IIncrementalGenerator
         return new TypeMetadata(globallyQualified, type.ToDisplayString(), patternTypeName);
     }
 
-    private record TypeMetadata(string GloballyQualified, string DisplayString, string PatternTypeName);
+    public record TypeMetadata(string GloballyQualified, string DisplayString, string PatternTypeName);
 
-    private record ConversionMetadata
+    public record ConversionMetadata
     {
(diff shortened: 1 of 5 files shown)


TimothyMakkison:
Added `AotConverterGeneratorIncrementalTests` to verify that incremental caching is occuring.

TimothyMakkison:
Last update, added assertion that for `EquatbableArray<ConversionMetada>` length to avoid mistakes I made when debgging these tests.

For the future: might be worth looking at using `CreateSyntaxProvider` or better yet `ForAttributeWithMetadataName` instead of descending every syntax tree.
That and using `WellKnownTypes`, `SymbolAccessor` (when added) and avoiding calling `Members` and `GetAttributes` repeatedly.

thomhurst:
Yeah definitely. These source generators are definitely not optimized! I was learning them as I went along.

Appreciate the help!

PR #4638, opened 2026-02-01
perf: make `StaticPropertyInitializationGenerator` incremental

TimothyMakkison:
- Add `PropertyWithDataSourceModel` and associated types
- Return `ImmutableArray<PropertyWithDataSource>` from `GetSemanticTargetForGeneration`
- Move `enabledProvider` check to earlier in pipeline to reduce work
- `ToUniquePropertyModels` ensures that only distinct properties are converter into `PropertyWithDataSourceModel`
- Only call `GetStaticPropertyDataSources` instead of doing a unnecessary call in `GenerateStaticPropertyInitialization`

### DIFF
--- a/TUnit.Core.SourceGenerator/CodeGenerators/StaticPropertyInitializationGenerator.cs
+++ b/TUnit.Core.SourceGenerator/CodeGenerators/StaticPropertyInitializationGenerator.cs
@@ -13,6 +13,8 @@ namespace TUnit.Core.SourceGenerator.CodeGenerators;
 [Generator]
 public class StaticPropertyInitializationGenerator : IIncrementalGenerator
 {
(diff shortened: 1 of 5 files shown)


TimothyMakkison:
- Fixed run length assertion in `AotConverterTests`
- Refactor and rename methods in the parse step
- Use a reusable `properties` list and keep track of already visited nodes
- Add tests to ensure that the generator is incremental




| Method       | Mean     | Error    | StdDev   | Gen0     | Allocated |
|------------- |---------:|---------:|---------:|---------:|----------:|
| RunGenerator | 20.42 ms | 0.354 ms | 0.562 ms | 250.0000 |   2.38 MB |


Time and memory usage seems suspiciously low 🤔 but the tests appear to pass

PR #4644, opened 2026-02-02
feat: confirm `InterfaceCache` memory leak

TimothyMakkison:
The `public static readonly ConcurrentDictionary` are permanently keeping references to outdated `Compilations` causing a memory leak

### Current
340 references to stale compilations. Total 123 MB
573 references to stale compilations. Total 127 MB
806 references to stale compilations. Total 134 MB
1039 references to stale compilations. Total 135 MB
1272 references to stale compilations. Total 142 MB
1505 references to stale compilations. Total 149 MB
1738 references to stale compilations. Total 156 MB
1971 references to stale compilations. Total 163 MB
2204 references to stale compilations. Total 171 MB
2437 references to stale compilations. Total 178 MB
2670 references to stale compilations. Total 185 MB
2903 references to stale compilations. Total 192 MB
3136 references to stale compilations. Total 199 MB
3369 references to stale compilations. Total 206 MB
3602 references to stale compilations. Total 213 MB
3835 references to stale compilations. Total 220 MB
4068 references to stale compilations. Total 227 MB

### With #4645
0 references to stale compilations. Total 116 MB
0 references to stale compilations. Total 116 MB
0 references to stale compilations. Total 116 MB
0 references to stale compilations. Total 110 MB
0 references to stale compilations. Total 110 MB
0 references to stale compilations. Total 110 MB
0 references to stale compilations. Total 110 MB
0 references to stale compilations. Total 110 MB
0 references to stale compilations. Total 110 MB
0 references to stale compilations. Total 111 MB
0 references to stale compilations. Total 111 MB
0 references to stale compilations. Total 111 MB
0 references to stale compilations. Total 111 MB
0 references to stale compilations. Total 111 MB
0 references to stale compilations. Total 111 MB
0 references to stale compilations. Total 111 MB

PR #4645, opened 2026-02-02
fix: make `InterfaceCache` stateless and prevent memoryleak

TimothyMakkison:
Make `InterfaceCache` stateless to prevent creating static references to `INamedTypeSymbol` and creating a memory leak

See #4644

### DIFF
--- a/TUnit.Core.SourceGenerator/Helpers/InterfaceCache.cs
+++ b/TUnit.Core.SourceGenerator/Helpers/InterfaceCache.cs
@@ -1,4 +1,3 @@
-using System.Collections.Concurrent;
 using Microsoft.CodeAnalysis;
 using TUnit.Core.SourceGenerator.Extensions;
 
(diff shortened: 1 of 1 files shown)


thomhurst:
Thanks for finding this issue!

TimothyMakkison:
> Thanks for finding this issue!

No problem, easy mistake to made. I am a little worried that a smaller memory leak is happening as #4644 still shows a small increase each time we rerun the generator. 

This might just be a quirk of the test or compilation, or it could be another leak. I don't see many likely culprits though 🤔

thomhurst:
Can we run it through `dotnet-trace` on a loop or something to measure memory?

TimothyMakkison:
Good idea, never used it before but it seems appropriate. I fruitlessly tried comparing dot memory snapshots, needless I'm no expert in these kinds of issues.

> This might just be a quirk of the test or compilation, or it could be another leak. I don't see many likely culprits though

On the plus side, the leak (if it exists) seems fairly small. If it's proportionate to the size of the project then it'll be fairly benign in most real world smaller test projects


Again this might be a quirk of the test. Has anyone ever reported an ide memory usage issue?

PR #4648, opened 2026-02-03
fix: enable Windows long path support in GitHub Actions

thomhurst:
## Summary
Fixes PathTooLongException errors on Windows CI by enabling long path support (>260 characters) via registry setting.

## Problem
Windows CI was failing with:
```
System.IO.PathTooLongException: The specified path, file name, or both are too long. The fully qualified file name must be less than 260 characters, and the directory name must be less than 248 characters.
```

This occurred during source generation for test methods with complex generic signatures like `GenericMethodTests.AggregateBy_HasExpectedOutput<TSource, TKey, TAccumulate>(...)`.

## Solution
Enable Windows long path support in GitHub Actions by setting the `LongPathsEnabled` registry key before building. This is a cleaner solution than modifying source generator code to hash long filenames.

## Test Plan
- [x] Windows CI should pass without PathTooLongException
- [x] Source generator tests with long method names should work

Fixes: https://github.com/thomhurst/TUnit/actions/runs/21610514651

### DIFF
--- a/.github/workflows/dotnet.yml
+++ b/.github/workflows/dotnet.yml
@@ -100,6 +100,12 @@ jobs:
         run: npx playwright install
         if: steps.playwright-cache.outputs.cache-hit == 'true'
 
+      - name: Enable Long Paths on Windows
+        if: matrix.os == 'windows-latest'
+        run: |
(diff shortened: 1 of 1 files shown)


Frulfump:
Will this force users to also use long paths to avoid potential issues? Could your CI now hide issues that users would run into otherwise?

PR #4659, opened 2026-02-03
fix: prevent substring matching in UID filter causing incorrect test inclusion

### LINKED ISSUE #4656: [Bug]: [After(TestDiscovery)] TestDiscoveryContext.AllTests is incorrect
### Description

When you do the following:
- Group tests into different classes that have overlapping test method names
- Give those tests dependencies which have an overlapping test name
- Run specific subsets of those tests

It can occasionally trip up the Test Discovery phase and cause tests that do not execute to incorrectly appear in `TestDiscoveryContext.AllTests`.

### Expected Behavior

1. Run tests `ABCD.B2`, `ABCV.B1`, `ABCVC.B2`, and `VPCU.A1000`
2. `ABCV.B2` _is not_ included in `TestDiscoveryContext.AllTests`
3. `ABCV.B2` is not executed

### Actual Behavior

1. Run tests `ABCD.B2`, `ABCV.B1`, `ABCVC.B2`, and `VPCU.A1000`
2. `ABCV.B2` _is_ included in `TestDiscoveryContext.AllTests`
3. `ABCV.B2` is not executed

### Steps to Reproduce

```cs
public class ABCD
{
    [Test]
    public async Task B1()
    {
        await Assert.That(true).IsEqualTo(true);
    }

    [Test]
    public async Task B2()
    {
        await Assert.That(true).IsEqualTo(true);
    }
}


public class VPCU
{
    [Test, DependsOn<ABCV>(nameof(ABCV.B1))]
    public async Task A1000()
    {
        await Assert.That(true).IsEqualTo(true);
    }
}

public class ABCV
{
    [Test]
    public async Task A1()
    {
        await Assert.That(true).IsEqualTo(true);
    }

    [Test, DependsOn(nameof(A1))]
    public async Task B1()
    {
        await Assert.That(true).IsEqualTo(true);
    }

    [Test, DependsOn(nameof(A1))]
    public async Task B2()
    {
        await Assert.That(true).IsEqualTo(true);
    }
}

public class ABCVC
{
    [Test]
    public async Task B0()
    {
        await Assert.That(true).IsEqualTo(true);
    }

    [Test, DependsOn(nameof(B0))]
    public async Task B1()
    {
        await Assert.That(true).IsEqualTo(true);
    }

    [Test, DependsOn(nameof(B0))]
    public async Task B2()
    {
        await Assert.That(true).IsEqualTo(true);
    }
}

public static class TestSetup
{
    [After(TestDiscovery)]
    public static void Setup(TestDiscoveryContext context) => ...;
}
```

### TUnit Version

1.12.125

### .NET Version

.NET 10

### Operating System

Windows

### IDE / Test Runner

Visual Studio

```shell

```

### Additional Context

I played with this for a while to try and create the smallest reproduction, and this is what I came up with. I'm not entirely sure what triggers this situation to happen, but this example does reproduce it consistently.

### IDE-Specific Issue?

thomhurst:
## Summary

Fixes #4656

When filtering tests by UID (e.g., from Visual Studio), class names like `ABCV` would incorrectly match `ABCVC` because the filter used `string.Contains()` for matching. This caused `TestDiscoveryContext.AllTests` to include tests that shouldn't be there.

### Root Cause

In `MetadataFilterMatcher.CouldMatchUidFilter()`, the code used substring matching:
```csharp
if (uidValue.Contains(className) && uidValue.Contains(methodName))
```

This meant that when checking if test `ABCV.B2` matches a filter containing UID for `ABCVC.B2`, it would incorrectly return `true` because "ABCVC" contains "ABCV" as a substring.

### The Fix

Changed to use **word-boundary matching**:
- Class names must be bounded by `.` (namespace/indices), `<` (generics), or `+` (nested types)
- Method names must be preceded by `.` and followed by `.`, `<`, or `(`

This ensures `ABCV` only matches the exact class name, not substrings like `ABCVC`.

## Test plan

- [x] Added reproduction test case in `TUnit.TestProject/Bugs/4656/`
- [x] Added integration tests in `TUnit.Engine.Tests/OverlappingClassNameFilterTests.cs`
- [x] Tests verify that filtering for `ABCVC/B2` runs exactly 2 tests (not 4 if bug exists)
- [x] Tests verify that filtering for `ABCV/*` runs exactly 3 tests (not 6 if bug exists)
- [x] All tests pass in both SourceGenerated and Reflection modes

### DIFF
--- a/TUnit.Engine/Services/MetadataFilterMatcher.cs
+++ b/TUnit.Engine/Services/MetadataFilterMatcher.cs
@@ -213,6 +213,7 @@ internal sealed class MetadataFilterMatcher : IMetadataFilterMatcher
     private static bool CouldMatchUidFilter(TestNodeUidListFilter filter, TestMetadata metadata)
     {
         var classMetadata = metadata.MethodMetadata.Class;
+        var namespaceName = classMetadata.Namespace ?? "";
         var className = metadata.TestClassType.Name;
         var methodName = metadata.TestMethodName;
 
@@ -224,14 +225,27 @@ internal sealed class MetadataFilterMatcher : IMetadataFilterMatcher
             classNameForMatching = className.Substring(0, backtickIndex);
         }
 
+        // Build expected prefix: {Namespace}.{ClassName}. or just {ClassName}. for empty namespace
+        // This ensures we match the exact class in the exact namespace
+        var expectedClassPrefix = string.IsNullOrEmpty(namespaceName)
+            ? $"{classNameForMatching}."
+            : $"{namespaceName}.{classNameForMatching}.";
+
+        // Also handle generic class names in UIDs (e.g., Namespace.MyClass<System.Int32>.0.0...)
+        var expectedGenericClassPrefix = string.IsNullOrEmpty(namespaceName)
+            ? $"{classNameForMatching}<"
+            : $"{namespaceName}.{classNameForMatching}<";
+
         foreach (var uid in filter.TestNodeUids)
         {
             var uidValue = uid.Value;
 
-            // Check for class name with word boundaries to avoid substring false positives
-            // e.g., "ABCV" should not match "ABCVC"
-            // Class names in TestIds are bounded by: '.' (namespace/indices), '<' (generics), '+' (nested)
-            if (!HasClassNameMatch(uidValue, classNameForMatching))
+            // Check for exact namespace.classname prefix to avoid matching
+            // same class name in different namespaces
+            var hasClassPrefix = uidValue.StartsWith(expectedClassPrefix, StringComparison.Ordinal) ||
+                                 uidValue.StartsWith(expectedGenericClassPrefix, StringComparison.Ordinal);
+
+            if (!hasClassPrefix)
             {
                 continue;
             }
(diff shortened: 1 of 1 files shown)


TUnit.Engine/Services/MetadataFilterMatcher.cs
```diff
@@ -217,15 +217,78 @@ private static bool CouldMatchUidFilter(TestNodeUidListFilter filter, TestMetada
         var className = metadata.TestClassType.Name;
         var methodName = metadata.TestMethodName;
 
+        // Handle generic types: Type`1 -> need to match Type< in the UID
+        var classNameForMatching = className;
+        var backtickIndex = className.IndexOf('`');
+        if (backtickIndex > 0)
+        {
+            classNameForMatching = className.Substring(0, backtickIndex);
```
TimothyMakkison, line 225: `classNameForMatching = className.AsSpan().Slice(0, backtickIndex);` and make `ReadOnlySpan<char> classNameForMatching`

Could prolly use `ValueStringBuilder` for `expectedClassPrefix` and `expectedGenericClassPrefix` but that's pretty pendanitc

PR #4692, opened 2026-02-06
perf: remove `List` usage from `AttributeWriter`

TimothyMakkison:
Remove `List<AttributeData>` usage in `AttributeWriter.WriteAttributes`

## Benchmarks
Note that timing is inaccurate, I didn't bother running them under benchmark conditions
### Before

| Method       | Mean     | Error   | StdDev   | Median   | Gen0       | Gen1      | Allocated |
|------------- |---------:|--------:|---------:|---------:|-----------:|----------:|----------:|
| RunGenerator | 187.1 ms | 5.75 ms | 16.97 ms | 178.6 ms | 11000.0000 | 4000.0000 | 101.25 MB |

### After

| Method       | Mean     | Error   | StdDev   | Median   | Gen0       | Gen1      | Allocated |
|------------- |---------:|--------:|---------:|---------:|-----------:|----------:|----------:|
| RunGenerator | 203.4 ms | 6.20 ms | 17.99 ms | 194.7 ms | 11000.0000 | 4000.0000 | 100.62 MB |

### DIFF
--- a/TUnit.Core.SourceGenerator/CodeGenerators/Writers/AttributeWriter.cs
+++ b/TUnit.Core.SourceGenerator/CodeGenerators/Writers/AttributeWriter.cs
@@ -13,50 +13,46 @@ public class AttributeWriter(Compilation compilation, TUnit.Core.SourceGenerator
     public void WriteAttributes(ICodeWriter sourceCodeWriter,
         IEnumerable<AttributeData> attributeDatas)
     {
-        var attributesToWrite = new List<AttributeData>();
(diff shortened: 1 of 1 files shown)


TimothyMakkison:
Is Claude hallucinating a bug here? The snapshot tests haven't failed and I don't think my changes would create that output.

thomhurst:
> Is Claude hallucinating a bug here? The snapshot tests haven't failed and I don't think my changes would create that output.

Yep 🤣

Issue #4811, opened 2026-02-15
[Bug]: xUnit fixer incorrectly converts `Assert.All` with item action


ascott18:
### Description

The following xUnit test will not have the inner item assertions converted.

``` c#
    [Fact]
    public void TestWithAssertAll()
    {
        var items = new[] { 1, 2, 3 };

        Assert.All(items, item =>
        {
            Assert.True(item > 0);
            Assert.True(item < 10);
            Assert.NotEqual(0, item);
        });
    }
```

### Expected Behavior

Ideally, an overload of `All` that takes an async delegate and allows for nested awaited TUnit assertions, perhaps working in a similar way for `Assert.Multiple()` for each nested item? Like this:

``` c#
        await Assert.That(items).All(async item =>
        {
            await Assert.That(item > 0).IsTrue();
            await Assert.That(item < 10).IsTrue();
            await Assert.That(item).IsNotEqualTo(0);
        });
```

Or, failing that, to just have the assertion be converted to a foreach loop:

``` c#
        foreach (var item in items)
        {
            await Assert.That(item > 0).IsTrue();
            await Assert.That(item < 10).IsTrue();
            await Assert.That(item).IsNotEqualTo(0);
        };
```

### Actual Behavior

Fixer produces the following, which is not a valid way to use TUnit's `All` assertion, and inner assertions are left as xUnit assertions.

``` c#
        await Assert.That(items).All(item =>
        {
            Assert.True(item > 0);
            Assert.True(item < 10);
            Assert.NotEqual(0, item);
        });
```

### Steps to Reproduce

Run xUnit code fixer on the code in the description above.

### TUnit Version

1.14.0

### .NET Version

.NET 10

### Operating System

Windows

### IDE / Test Runner

dotnet CLI (dotnet test / dotnet run)

### Error Output / Stack Trace

```shell

```

### IDE-Specific Issue?

- [ ] I've confirmed this issue occurs when running via `dotnet test` or `dotnet run`, not just in my IDE

thomhurst:
This should be sorted in the latest version. Let me know if not.

Issue #4806, opened 2026-02-15
[Bug]: xUnit code fixer drops `Assert.Collection` assertions


ascott18:
### Description

An xunit assertion with item inspectors like this...

```c#
        Assert.Collection(orderings,
            ordering =>
            {
                var prop = Assert.Single(ordering.Properties);
                Assert.Equal("Parent", prop.Name);
            }
        );
```

... has the item inspectors dropped entirely, leaving only `await Assert.That(orderings).HasCount(1);`

This leads to tests that still pass, but in a large test refactor, its easy to miss that the other assertions were lost.

### Expected Behavior

Item assertions are preserved in some form, or the `Assert.Collection` is left as-is for the user to convert to TUnit manually.

### Actual Behavior

Assertions are dropped in a way that still leaves passing tests in place, but missing additional critical test logic.

### Steps to Reproduce

Run xUnit code fixer on an assertion like this:

```c#
        Assert.Collection(orderings,
            ordering =>
            {
                var prop = Assert.Single(ordering.Properties);
                Assert.Equal("Parent", prop.Name);
            }
        );
```

### TUnit Version

1.14.0

### .NET Version

NET 10

### Operating System

Windows

### IDE / Test Runner

dotnet CLI (dotnet test / dotnet run)

### Error Output / Stack Trace

```shell

```

### IDE-Specific Issue?

- [ ] I've confirmed this issue occurs when running via `dotnet test` or `dotnet run`, not just in my IDE

thomhurst:
This should be sorted in the latest version. Let me know if not.

Issue #4805, opened 2026-02-15
[Bug]: xUnit code fixer drops lambda from Assert.Single


ascott18:
### Description

The lambda from the xUnit method `Assert.Single<T>(IEnumerable<T> collection, Predicate<T> predicate)` is dropped by the xUnit code fixer.

### Expected Behavior

`Assert.Single(properties, p => p.Prop == "myKey");` should become `await Assert.That(properties.Where(p => p.Prop == "myKey")).HasSingleItem();`.

Or, even better, an equivalent `HasSingleItem` overload that accepts a filter lambda is added to TUnit, allowing `await Assert.That(properties).HasSingleItem(p => p.Prop == "myKey");`.

### Actual Behavior

Lambda was dropped entirely, leaving only `await Assert.That(properties).HasSingleItem();`

### Steps to Reproduce

Apply xUnit code fixers against an xUnit assertion like `Assert.Single(properties, p => p.Prop == "myKey");`

### TUnit Version

1.14.0

### .NET Version

NET 10

### Operating System

Windows

### IDE / Test Runner

dotnet CLI (dotnet test / dotnet run)

### Error Output / Stack Trace

```shell

```

### IDE-Specific Issue?

- [ ] I've confirmed this issue occurs when running via `dotnet test` or `dotnet run`, not just in my IDE

thomhurst:
This should be sorted in the latest version. Let me know if not.

PR #4821, opened 2026-02-15
fix: xUnit code fixer bugs for Assert.Single, Assert.Collection, and Assert.All

thomhurst:
## Summary

Fixes three xUnit code fixer bugs reported in #4805, #4806, and #4811:

- **#4805 - Assert.Single drops predicate lambda**: `Assert.Single(collection, predicate)` now correctly converts to `Assert.That(collection.Where(predicate)).HasSingleItem()` instead of silently dropping the predicate
- **#4806 - Assert.Collection drops element inspectors**: The TODO comment now preserves the full text of each element inspector (e.g. `[0]: x => Assert.Equal(1, x)`) so assertions aren't silently lost during migration
- **#4811 - Assert.All produces invalid code**: When `Assert.All` has a multi-statement lambda body that can't be converted to a simple predicate, the two-phase analyzer now skips conversion instead of producing invalid `Assert.That(items).All(item => { Assert.True(...); })` code

Both the assertion code fixer (`XUnitAssertionCodeFixProvider`) and the two-phase migration analyzer (`XUnitTwoPhaseAnalyzer`) are updated where applicable.

## Test plan

- [x] New tests added for `Assert.Single` with and without predicate in assertion code fixer
- [x] Existing `Assert_Collection_Adds_Todo_Comment` test updated to match improved TODO message
- [x] All 52 assertion code fixer tests pass
- [x] All 268 two-phase migration analyzer tests pass

### DIFF
--- a/TUnit.Analyzers.CodeFixers/TwoPhase/XUnitTwoPhaseAnalyzer.cs
+++ b/TUnit.Analyzers.CodeFixers/TwoPhase/XUnitTwoPhaseAnalyzer.cs
@@ -593,6 +593,13 @@ public class XUnitTwoPhaseAnalyzer : MigrationAnalyzer
         // Try to extract predicate from Assert.True/False patterns
         var predicate = TryConvertActionToPredicate(actionExpression);
 
+        // If the lambda body can't be converted to a simple predicate,
+        // skip conversion to avoid producing invalid code
+        if (predicate == null)
+        {
+            return (AssertionConversionKind.All, null, false, null);
+        }
+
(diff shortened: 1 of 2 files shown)


thomhurst:
The review concern about missing `using System.Linq;` is now resolved — after rebasing onto main (which includes #4832 `feat: add HasSingleItem(predicate) overload`), the code no longer generates `.Where(predicate).HasSingleItem()`. Instead, both the code fixer and two-phase analyzer now generate `HasSingleItem(predicate)` directly, which doesn't require `System.Linq`.

All tests pass after rebase (13 assertion code fixer tests, 67 migration analyzer tests).

PR #4849, opened 2026-02-18
Cleanup global usings in Analyzers

SimonCropp:
This pull request adds support for detecting and removing global using directives for test frameworks (NUnit, MSTest, and xUnit) in both the analyzers and their tests. The analyzers now flag global usings like `global using Xunit;` as migration issues, and the code fixers can remove them.


thomhurst:
Thanks 😁

PR #4961, opened 2026-02-19
Preserve newline for EOF preprocessor directives

SimonCropp:
after migrating i got 

```
#if DEBUG
public class HotKeyControlTests
{
    [Fact]
    public async Task Default()
    {
        using var target = new HotKeyControl();
        await Verify(target);
    }
}#endif
```

note the `}#endif`


SimonCropp:
@thomhurst while the fix works, i am not experienced with writing analyzers, so a critical eye on the approach i have take would be appreciated

thomhurst:
Thanks @SimonCropp !

PR #4991, opened 2026-02-23
feat(mocks): unified Mock<T> API — remove Setup/Verify/Raise surfaces

thomhurst:
## Summary

- **Remove `.Setup.`, `.Verify.`, and `.Raise.` intermediate properties** from `Mock<T>`. Methods are now called directly on the mock via source-generated extension methods, with chain methods (`.Returns()` vs `.WasCalled()`) disambiguating between setup and verification.
- **Add unified return types** (`MockMethodCall<T>`, `VoidMockMethodCall`, `PropertyMockCall<T>`, `PropertySetterMockCall<T>`) that expose both setup and verification surfaces on a single object.
- **Replace `MockSetupBuilder` + `MockVerifyBuilder`** with a single `MockMembersBuilder` that generates extension methods on `Mock<T>` and unified sealed classes per qualifying method.
- **Void methods and property setters eagerly register** their setups in the constructor, since they're commonly used without chaining (e.g., `mock.Log(Arg.Any<string>())` to allow calls in strict mode).

### Before
```csharp
mock.Setup.Greet(Arg.Any<string>()).Returns("Hello");
mock.Verify.Greet(Arg.Any<string>()).WasCalled();
mock.Raise.OnMessage("hi");
```

### After
```csharp
mock.Greet(Arg.Any<string>()).Returns("Hello");
mock.Greet(Arg.Any<string>()).WasCalled();
mock.RaiseOnMessage("hi");
```

## Test plan

- [x] All 475 runtime tests pass (`TUnit.Mocks.Tests` on net10.0)
- [x] All 10 source generator snapshot tests pass (`TUnit.Mocks.SourceGenerator.Tests`)
- [x] Snapshot `.verified.txt` files updated to reflect new generated output
- [ ] Verify CI passes across all target frameworks (net8.0, net9.0, net10.0)

### DIFF
--- a/TUnit.Mocks.SourceGenerator.Tests/Snapshots/Interface_Inheriting_Multiple_Interfaces.verified.txt
+++ b/TUnit.Mocks.SourceGenerator.Tests/Snapshots/Interface_Inheriting_Multiple_Interfaces.verified.txt
@@ -98,7 +98,7 @@ namespace TUnit.Mocks.Generated
         private readonly int _memberId;
         private readonly string _memberName;
         private readonly global::TUnit.Mocks.Arguments.IArgumentMatcher[] _matchers;
-        private global::TUnit.Mocks.Setup.VoidMethodSetupBuilder? _builder;
+        private readonly global::System.Lazy<global::TUnit.Mocks.Setup.VoidMethodSetupBuilder> _lazyBuilder;
 
         internal IReadWriter_Write_M2_MockCall(global::TUnit.Mocks.IMockEngineAccess engine, int memberId, string memberName, global::TUnit.Mocks.Arguments.IArgumentMatcher[] matchers)
         {
@@ -106,20 +106,17 @@ namespace TUnit.Mocks.Generated
             _memberId = memberId;
             _memberName = memberName;
             _matchers = matchers;
-            EnsureSetup();
-        }
-
-        private global::TUnit.Mocks.Setup.VoidMethodSetupBuilder EnsureSetup()
-        {
-            if (_builder is null)
+            _lazyBuilder = new global::System.Lazy<global::TUnit.Mocks.Setup.VoidMethodSetupBuilder>(() =>
             {
                 var setup = new global::TUnit.Mocks.Setup.MethodSetup(_memberId, _matchers, _memberName);
                 _engine.AddSetup(setup);
-                _builder = new global::TUnit.Mocks.Setup.VoidMethodSetupBuilder(setup);
+                return new global::TUnit.Mocks.Setup.VoidMethodSetupBuilder(setup);
             }
-
-            return _builder;
+            );
+            _ = _lazyBuilder.Value;
         }
+
+        private global::TUnit.Mocks.Setup.VoidMethodSetupBuilder EnsureSetup() => _lazyBuilder.Value;
 
         /// <inheritdoc />
         public IReadWriter_Write_M2_MockCall Throws<TException>() where TException : global::System.Exception, new() { EnsureSetup().Throws<TException>(); return this; }
--- a/TUnit.Mocks.SourceGenerator.Tests/Snapshots/Interface_With_Async_Methods.verified.txt
+++ b/TUnit.Mocks.SourceGenerator.Tests/Snapshots/Interface_With_Async_Methods.verified.txt
@@ -319,7 +319,7 @@ namespace TUnit.Mocks.Generated
         private readonly int _memberId;
         private readonly string _memberName;
         private readonly global::TUnit.Mocks.Arguments.IArgumentMatcher[] _matchers;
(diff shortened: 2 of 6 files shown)


thomhurst:
Addressed in f225a2c:

**Issue 1 — Thread-safety race in `EnsureSetup()`:** Fixed. Non-void `MockMethodCall<TReturn>` and generated non-void sealed classes now use `Lazy<T>` for thread-safe single initialization. Void methods and property setters were already safe (eager init in the constructor completes before the object is visible to other threads).

**Issue 2 — Void vs non-void asymmetry:** Added XML doc comment on `MockMethodCall<TReturn>` explaining the distinction and suggesting `.Returns(default!)` as the explicit escape hatch for non-void methods in strict mode.

**Issue 3 — Verification object per call:** Skipping. These `MockMethodCall` instances are ephemeral per-call objects, so the verification builder is equally short-lived. Caching adds complexity for no measurable benefit in test code.

**Issue 4 — Hardcoded `MockMemberNames`:** Acknowledged as a maintenance concern. Deriving from `Mock<T>` symbol analysis at generator time is doable but non-trivial (the generator runs against the user's compilation, not TUnit.Mocks itself). Tracking as a follow-up.

**Issue 5 — `Getter` re-registration:** Added clarifying comment explaining this is intentional — each access allows chaining different behaviors that stack as additional setups.

All 475 runtime tests and 10 snapshot tests pass.

thomhurst:
Addressed in f5f06f4 (responding to reviews 2 and 3):

**Review 2 Issue 1 (Getter re-registration):** This is pre-existing behavior from `PropertySetupAccessor`. Each access intentionally allows stacking setups (last wins). Already documented in prior commit.

**Review 2 Issue 2 (Void eager + VerifyAll):** The specific edge case is `mock.Log(Arg.Any<string>()).WasNeverCalled()` followed by `mock.VerifyAll()` — the eager registration creates a setup that was never invoked. In practice, if `WasCalled()` passes, the setup was also invoked, so `VerifyAll()` agrees. The `WasNeverCalled()` + `VerifyAll()` combo is the real concern, but no existing test hits this pattern (475/475 pass). A proper fix requires the engine to distinguish setup-registrations from verification-only objects — tracking as a follow-up.

**Review 2 Issue 3 (Allocation regression):** Deliberate. These are test-time ephemeral objects, not engine hot paths. The `readonly struct` → `sealed class` change was required for `Lazy<T>` and mutable builder fields.

**Review 2 Issue 4 / Review 3 Issue 2 (Member name inconsistency):** Fixed. Setup paths now use `"get_Name"` / `"set_Name"` matching the verification paths and the engine's call records.

**Review 2 Issue 5 (Non-void silent no-op):** Already documented. A `[MustUseReturnValue]`-style Roslyn analyzer is a good follow-up idea.

**Review 3 Issue 1 (Setter re-registration):** Fixed. `PropertyMockCall.Setter` now creates `VoidMockMethodCall` with `eagerRegister: false`. Accessing `.Setter` for verification no longer registers a setup. Setup methods (`.Callback()`, `.Throws()`) still trigger registration on demand.

**Review 3 Issue 3 (VoidMockMethodCall consistency):** Fixed. Now uses `Lazy<T>` matching the non-void pattern, with an `eagerRegister` parameter that forces evaluation in the constructor when `true`.

All 475 runtime tests and 10 snapshot tests pass.

thomhurst:
**Addressed review 4 feedback:**

**Issue 1 — `PropertySetterMockCall<T>` consistency:** Fixed. Replaced manual null-check `_builder` field with `Lazy<VoidMethodSetupBuilder>`, matching the pattern in `VoidMockMethodCall` and `MockMethodCall<TReturn>`. Commit: 78c57a36b.

**Issue 2 — `VerifyAll` + void eager registration:** Acknowledged as a known limitation. Proper fix requires engine-level changes (distinguishing setup vs verification objects, or adding an `.Allow()` method). Tracked as follow-up work.

thomhurst:
**Addressed review 5 feedback:**

**Issue 1 — `PropertySetterMockCall` allocation per verification call:** Fixed. Cached `_matchers` array and `_setterName` string as `readonly` fields in the constructor, matching the pattern in `VoidMockMethodCall` and `MockMethodCall<TReturn>`. No more per-call allocation or string interpolation. Commit: c6690cc18.

**Issue 2 — `GetSafeMemberName` cascading collision:** Acknowledged as edge case. The current `MockMemberNames` set contains names from `Mock<T>` base (`Object`, `Engine`, `Reset`, etc.) — a user having a method named e.g. `Engine_` is astronomically unlikely. Tracked as follow-up alongside deriving the name set from symbol analysis.

**Observation — getter-only verification doc:** Fixed. Updated `PropertyMockCall<T>` XML doc to explicitly state that `WasCalled()`/`WasNeverCalled()` verify the getter only, with guidance to use `.Setter` or `.Set()` for setter verification. Same commit.

PR #5017, opened 2026-02-26
feat(mocks): concise API — inline lambdas, untyped Any(), global static import

thomhurst:
## Summary

- **Inline lambda predicates**: Generate `Func<T, bool>` extension method overloads in the source generator so `mock.Add(x => x > 5, Any()).Returns(100)` compiles directly (sidesteps CS1660). For N eligible parameters, generates all 2^N - 1 overloads (capped at N=8).
- **Untyped `Any()`**: Add `Arg.Any()` returning a sentinel that implicitly converts to `Arg<T>`, so `Any()` works without type arguments
- **Implicit `Func<T, bool>` conversion**: Add `implicit operator Arg<T>(Func<T, bool>)` for predicate matching via variable assignment
- **Global static import**: Add `global using static Arg` to `TUnit.Mocks.targets` so matchers (`Any()`, `Is<T>()`, `IsInRange()`, etc.) work without the `Arg.` prefix
- **Update all tests and docs** to use the concise syntax

## Test plan

- [x] Source generator builds with 0 errors
- [x] All 11 snapshot tests pass with updated `.verified.txt` files
- [x] 13 new `FuncOverloadTests.cs` + new `ImplicitArgConversionTests.cs` + new `UntypedAnyTests.cs`
- [x] Full suite: 658/658 tests pass (no regressions)
- [x] All documentation updated to showcase concise syntax

### DIFF
--- a/Library.props
+++ b/Library.props
@@ -3,7 +3,6 @@
         <TargetFrameworks>netstandard2.0;net8.0;net9.0;net10.0</TargetFrameworks>
 
         <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-        <NoWarn>$(NoWarn);AD0001</NoWarn>
 
         <TrimmerSingleWarn>false</TrimmerSingleWarn>
 
--- a/TUnit.Mocks.SourceGenerator/Builders/MockMembersBuilder.cs
+++ b/TUnit.Mocks.SourceGenerator/Builders/MockMembersBuilder.cs
@@ -13,6 +13,7 @@ namespace TUnit.Mocks.SourceGenerator.Builders;
 internal static class MockMembersBuilder
 {
     private const int MaxTypedParams = 8;
+    private const int MaxFuncOverloadParams = 4;
 
     private static readonly HashSet<string> MockMemberNames = new(System.StringComparer.Ordinal)
     {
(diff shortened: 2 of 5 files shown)


thomhurst:
## Response to review

### "Missing implicit operator" — already in the PR

`ArgOfT.cs` is modified in the second commit (`4ad1bcd`). The PR adds:

```csharp
public static implicit operator Arg<T>(Func<T, bool> predicate) => new(new PredicateMatcher<T>(predicate!));
```

The generated `__fa_name = name` assignments rely on this and compile correctly. All 658 tests pass.

### "2^N overload explosion — implicit operator is the cleaner solution" — incorrect; CS1660

The implicit operator alone does **not** enable inline lambdas. C# specification (CS1660):

> Cannot convert lambda expression to type 'Arg\<int\>' because it is not a delegate type

Verified with a minimal repro just now — even with `implicit operator Arg<T>(Func<T, bool>)` present, `mock.Add(x => x > 5, default(Arg<int>))` produces CS1660. Lambda expressions cannot participate in user-defined implicit conversions to non-delegate types. This is the entire motivation for the source-gen approach.

The implicit operator IS useful for `Func<T, bool>` **variables** (which already have delegate type), but not for inline lambdas. Both are needed:
- Implicit operator: `Func<int, bool> pred = x => x > 5; mock.Add(pred, Any());`
- Generated overloads: `mock.Add(x => x > 5, Any());`

In practice, most interfaces have 1-3 params per method (1, 3, 7 overloads), well within reasonable bounds.

### Logic duplication — acknowledged

Fair observation. `EmitSingleFuncOverload` shares return-type resolution and return-statement logic with `EmitMemberMethodBody`. Could extract shared helpers. Acceptable for now since the methods serve distinct purposes and the code is straightforward, but worth revisiting if the emission logic grows.

### Missing test coverage — Func-typed parameter case added

Verified that a mocked method with `Func<int, bool>` as a parameter type compiles without ambiguity (Func value targets `Arg<Func<int,bool>>` via implicit T→Arg\<T\>, lambda targets the generated `Func<Func<int,bool>, bool>` overload). Added a test for this.

The null predicate case will produce a `NullReferenceException` from `PredicateMatcher<T>.Matches()` — consistent with how null values behave elsewhere in the matcher pipeline.

### Minor items

- `__fa_` prefix: intentionally matches the `__` compiler-generated convention. Users shouldn't encounter it in diagnostics since the generated code compiles or fails at the source generator level.
- `HashSet<int>`: used for O(1) `Contains()` lookups in the matcher-building loop. A List would work but the HashSet makes the intent clearer.
- Generic method guard: generic methods are intentionally included — `Func<T, bool>` works with type parameters. `ShouldGenerateTypedWrapper` excludes generics from the *sealed wrapper class* (different concern — generic type args in class names).

thomhurst:
## Addressed in 637c3e7

**AD0001 suppression** — Removed from `Library.props`. Confirmed no AD0001 errors fire without it.

**Overload explosion** — Added separate `MaxFuncOverloadParams = 4` (max 15 overloads per method), keeping `MaxTypedParams = 8` for typed wrapper classes which are a different concern. Methods with 5+ params can still use `Is<T>(predicate)` or `Func<T, bool>` variables.

**Arg class design** — Refactored `Arg` back to `static class`. Extracted `AnyArg` as a dedicated sentinel type with `[EditorBrowsable(Never)]`, private constructor, and singleton instance. `Arg<T>` now converts from `AnyArg` instead of `Arg`.

**Logic duplication** — Extracted `GetReturnTypeInfo()` returning `(bool UseTypedWrapper, string ReturnType, string SetupReturnType)`, used by both `EmitMemberMethodBody` and `EmitSingleFuncOverload`.

**Ordering test** — Now verifies both setups are active: `Add(5, 0)` returns 50 (first only), `Add(15, 0)` returns 100 (last wins).

All 659 integration tests + 11 snapshot tests pass.

PR #5057, opened 2026-03-02
perf: add ParameterMetadataFactory and lazy ReflectionInfo resolution

thomhurst:
## Summary

- **New `ParameterMetadataFactory`** — replaces inline `new ParameterMetadata { ... }` object initializers in generated code with `ParameterMetadataFactory.Create(...)` calls, reducing per-parameter IL size and JIT-compiled native code (follows the existing `TestMetadataFactory`/`MethodMetadataFactory` pattern)
- **Lazy `ReflectionInfo` via `field` keyword** — all `GetMethod().GetParameters()[i]` reflection is deferred to first access via `ReflectionInfoFactory = static () => ...` lambdas, eliminating eager reflection at module init in source-gen mode (where `ReflectionInfo` is rarely accessed at runtime)
- **Hoist helpers** — added `GenerateParameterInfoArrayExpression()` and `GenerateParameterInfoHoistStatement()` to `MetadataGenerationHelper` for future use when eager parameter info arrays are needed

### Generated code before (per parameter):
```csharp
new ParameterMetadata(typeof(int))
{
    Name = "value",
    TypeInfo = new ConcreteType(typeof(int)),
    IsNullable = false,
    ReflectionInfo = typeof(X).GetMethod("M", flags, null, paramTypes, null)!.GetParameters()[0]
}
```

### Generated code after (per parameter):
```csharp
ParameterMetadataFactory.Create(typeof(int), "value",
    new ConcreteType(typeof(int)), false,
    reflectionInfoFactory: static () => typeof(X).GetMethod("M", flags, null, paramTypes, null)!.GetParameters()[0])
```

Net reduction: **-7,136 lines** / **+1,156 lines** across 88 files (mostly snapshot shrinkage from compact factory calls replacing multi-line object initializers).

## Test plan
- [x] `dotnet build TUnit.Core.SourceGenerator -c Release` — compiles
- [x] `TUnit.Core.SourceGenerator.Tests` — 456/456 pass (4 pre-existing skips)
- [x] `TUnit.PublicAPI` — 12/12 pass
- [x] `TUnit.TestProject` compiles on net10.0 (proves generated code is valid C#)
- [x] `DataDrivenTests` runtime — 15/15 pass
- [x] `MatrixTests` runtime — 275/275 pass (exercises lazy `ReflectionInfo` resolution path)

### DIFF
--- a/TUnit.Core/Models/TestModels/ParameterMetadata.cs
+++ b/TUnit.Core/Models/TestModels/ParameterMetadata.cs
@@ -26,7 +26,9 @@ public record ParameterMetadata([DynamicallyAccessedMembers(DynamicallyAccessedM
     public required TypeInfo TypeInfo { get; init; }
     public ParameterInfo ReflectionInfo
     {
-        get => field ??= ReflectionInfoFactory?.Invoke() ?? null!;
+        get => field ??= ReflectionInfoFactory?.Invoke()
+            ?? throw new InvalidOperationException(
+                $"ReflectionInfo for parameter '{Name}' was not set and no ReflectionInfoFactory was provided.");
         set;
     } = null!;
 


thomhurst:
Addressed the review feedback:

**Item 1 (`?? null!` landmine):** Fixed in `9b048dd` — replaced with `throw new InvalidOperationException(...)` with a descriptive message including the parameter name. Failures are now loud and informative.

**Item 2 (removal of `required`):** The `required` modifier can't coexist with the `field` keyword lazy pattern — the compiler requires the property to be set in every object initializer, but source-gen callers set `ReflectionInfoFactory` instead of `ReflectionInfo` directly. The reflection-mode callers (`ReflectionMetadataBuilder`, `ReflectionHookDiscoveryService`, `DataGeneratorMetadataCreator`, `ClassMetadataHelper`) all continue to set `ReflectionInfo` directly via the setter. The new `throw` in the getter (item 1) provides a runtime safety net for any caller that forgets to set either property.

**Item 3 (thread safety):** The `field ??=` pattern without synchronization is consistent with the rest of TUnit.Core (`Context.cs:39,42`, `HookMethod.cs:16,23`, `LastTestInClassAdapter.cs:14`, etc.). The race is benign — `ParameterInfo` from CLR reflection is immutable and deterministic, so double-invocation produces the same result. Adding `Interlocked` would be inconsistent with the codebase convention.

**Item 4 (dead code):** Already removed in `774af06` (the simplification commit). Both `GenerateParameterInfoArrayExpression` and `GenerateParameterInfoHoistStatement` were deleted.

**Item 5 (`WriteParameterMetadataGeneric` hardcodes `false`):** Already fixed in `774af06`. `WriteParameterMetadataGeneric` now delegates to `WriteParameterMetadata`, which correctly computes `isNullable` from `parameter.Type.IsNullable()`.

PR #5059, opened 2026-03-02
feat: distributed trace collection for HTML report

thomhurst:
## Summary

- Expands ActivityCollector to capture traces from all instrumented .NET libraries (HttpClient, ASP.NET Core, EF Core, etc.) — not just TUnit's own spans
- Adds `TestContext.Activity` (public) and `TestContext.RegisterTrace()` API for explicit external trace correlation
- Renders linked external traces in the HTML report timeline

## How it works

Activities from instrumented libraries (e.g. `System.Net.Http`, `Microsoft.AspNetCore`) automatically become child spans when they execute under a test's activity context (via `Activity.Current` AsyncLocal propagation). The smart sampling approach returns `AllDataAndRecorded` for known traces and `PropagationData` for everything else — enabling context propagation with near-zero overhead for unrelated activities.

For external services that don't share the test's trace context, users can explicitly link traces via:
```csharp
TestContext.Current!.RegisterTrace(externalActivity.Context.TraceId);
```

## Changes

| File | Change |
|------|--------|
| `TUnit.Core/TraceRegistry.cs` | **New** — static bidirectional traceId↔testNodeUid registry |
| `TUnit.Core/TestContext.cs` | Public `Activity` property + `RegisterTrace` method (`#if NET`) |
| `TUnit.Engine/.../ActivityCollector.cs` | Listen to all sources, smart sampling, gated collection |
| `TUnit.Engine/.../HtmlReportDataModel.cs` | `AdditionalTraceIds` on `ReportTestResult` |
| `TUnit.Engine/.../HtmlReporter.cs` | Populate `AdditionalTraceIds` from TraceRegistry |
| `TUnit.Engine/.../HtmlReportGenerator.cs` | `renderExternalTrace` JS + `renderDetail` update |

## Test plan

- [ ] `dotnet build` TUnit.Core and TUnit.Engine — 0 errors, 0 warnings
- [ ] Run test project with HttpClient calls, verify HTTP spans appear in test trace timeline
- [ ] Verify `TestContext.RegisterTrace` links external traces as "Linked Trace" sections
- [ ] Run existing snapshot/unit tests — no regressions
- [ ] Verify existing TUnit-only trace timelines render unchanged

### DIFF
--- a/TUnit.AspNetCore/Http/ActivityPropagationHandler.cs
+++ b/TUnit.AspNetCore/Http/ActivityPropagationHandler.cs
@@ -12,6 +12,9 @@ namespace TUnit.AspNetCore;
 /// </summary>
 internal sealed class ActivityPropagationHandler : DelegatingHandler
 {
+    // Intentionally process-scoped: lives for the test process lifetime and is
+    // cleaned up on process exit. Not disposed explicitly because multiple handler
+    // instances share this source across concurrent tests.
     private static readonly ActivitySource HttpActivitySource = new("TUnit.AspNetCore.Http");
 
     protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
--- a/TUnit.Engine/Reporters/Html/HtmlReporter.cs
+++ b/TUnit.Engine/Reporters/Html/HtmlReporter.cs
@@ -86,10 +86,18 @@ internal sealed class HtmlReporter(IExtension extension) : IDataConsumer, ITestH
                 return;
             }
 
-            var reportData = BuildReportData();
+            ReportData reportData;
+            try
+            {
+                reportData = BuildReportData();
+            }
+            finally
+            {
 #if NET
-            TraceRegistry.Clear();
+                TraceRegistry.Clear();
 #endif
+            }
+
             var html = HtmlReportGenerator.GenerateHtml(reportData);
 
             if (string.IsNullOrEmpty(html))


thomhurst (commented):
## Response to Follow-up Review

### Item 1: `ACTIONS_RUNTIME_TOKEN` — No change

This is the standard pattern for Microsoft.Testing.Platform CI integration (same in `dotnet.yml`). The token is run-scoped, short-lived, and the only subsequent steps are `dotnet build` and `dotnet run` — both trusted. The reviewer's suggested `env:` block scoping is functionally equivalent since the token is already available in the Actions runner environment.

### Item 2: `FindTestCaseAncestor` caching + GC concern — No change (pushback)

**a) Caching by traceId is incorrect.** All tests share a single traceId (rooted at the "test session" activity). Caching `traceId → testCaseSpanId` would return the first test's span for every subsequent activity, completely breaking per-test cap enforcement.

**b) `Activity.Parent` GC concern is unfounded.** `Activity.Parent` is a strong reference field on the child. The child Activity is alive during `OnActivityStopped` (it's the callback argument), so the entire parent chain is held alive by the GC. Parent activities cannot be collected while any child holds a reference.

### Item 3: Display-name string matching — Fixed ✓

Switched `FindTestCaseAncestor` from `DisplayName.StartsWith("test case")` to checking for the `tunit.test.node_uid` tag (set at `TestExecutor.cs:133`). Tag-based detection is robust against future activity renaming. Fixed in `6e39f88`.

### Item 4: Stale docs — Fixed ✓

Updated `docs/docs/guides/html-report.md` to reflect the current cap: 100 external spans per test (TUnit spans uncapped). Removed stale reference to 1,000/50,000 caps. Fixed in `6e39f88`.

### Item 5: Breaking change on `Factory` property type — Noted

`TUnit.AspNetCore` is a new unreleased package, so there are no existing consumers to break. The `Factory.Inner` escape hatch provides access to the raw `WebApplicationFactory<TEntryPoint>` when needed.

PR #5075, opened 2026-03-03
perf: remove `Action` allocation for `token.Register`

TimothyMakkison:
Use ` CancellationToken.Register(Action<object?> callback, object? state)` overload to pass a boxed `ValueTuple` of state into the function. This removes the `Action` allocation.

### Before
image

### After
image


TimothyMakkison:
FYI, gitguardian is flagging TUnit, I've marked it as erroneous, hopefully the issue won't persist

image

TimothyMakkison:
Added `UnconditionalSuppressMessage` tests should pass now

PR #5077, opened 2026-03-04
Improve HTML report perf

slang25:
## Description

This change is to address the stutteriness of the html report when toggling between dark and light mode. The biggest win is by deferring the test details to keep the dom elements/depth to a minimum.

_That said_, maybe we should embed a lightweight library for this kind of thing, like preact 🤔  Interested to hear your views.


thomhurst:
Ty!

PR #5095, opened 2026-03-06
fix: improve CreateTestVariant API and fix void/ValueTask return types

### LINKED ISSUE #5093: [Bug]: CreateTestVariant not working
### Description

I've been trying to use the method TestContext.CreateTestVariant to create variants of tests which encounter multiple paths to go forward. The idea is that when a test encounters multiple choices it would pick the first one and create test variants for the other choices. The new tests would have the specific choice saved in the statebag.

Here is the example I'm using. It's called inside a running test. I get the exception message "Could not extract method info from dynamic test expression".

```
stateBag.Add("Decider", CreateNewTree(newDecision));

await TestContext.Current!.CreateTestVariant(
    properties: stateBag,
    displayName: TestContext.Current!.Metadata.DisplayName + "_variant",
    relationship: TestRelationship.Generated);
```

I did notice now though that the TestVariant is no longer part of the documentation (I'm on 1.18.21 now). Is it no longer supported? I see that the dynamic tests are still there and I could use those but then I would have to use reflection to create them since the generic type parameter will be different depending on the test.

### Expected Behavior

I expected a new test to be created.

### Actual Behavior

I get an error when calling CreateTestVariant with the message "Could not extract method info from dynamic test expression".

### Steps to Reproduce

With the code below the message is shown.
If I change to use await in a async task then the test passes but no variants are created.
If I don't await I don't get any variant tests.

```csharp
public class VariantTestFixture
{
    [Test]
    public void VariantTest()
    {
        TestContext.Current!.CreateTestVariant(
            displayName: "VariantTestLeft",
            relationship: TestRelationship.Generated,
            properties: new()
            {
                { "decision", "go left" },
            }).Wait();

        TestContext.Current!.CreateTestVariant(
            displayName: "VariantTestRight",
            relationship: TestRelationship.Generated,
            properties: new()
            {
                { "decision", "go right" },
            }).Wait();

        TestContext.Current!.StateBag.TryGetValue("decision", out string? decision);

        Console.WriteLine($"Decisions is {decision ?? "nothing"}");

#pragma warning disable TUnitAssertions0005
#pragma warning disable TUnitAssertions0002
        Assert.That(10).IsEqualTo(10);
#pragma warning restore TUnitAssertions0002
#pragma warning restore TUnitAssertions0005
    }
}
```

### TUnit Version

1.18.21 

### .NET Version

.net 10.0

### Operating System

Windows

### IDE / Test Runner

Visual Studio

```shell
Message: 
[Test Failure] Could not extract method info from dynamic test expression

  Stack Trace: 
TestRegistry.CreateMetadataFromDynamicDiscoveryResult(DynamicDiscoveryResult result)
TestRegistry.ProcessPendingDynamicTests()
TestRegistry.CreateTestVariantInternal[T](TestContext currentContext, MethodInfo methodInfo, Object[] variantMethodArguments, Object[] classArguments, Dictionary`2 properties, TestRelationship relationship, String displayName)
TestRegistry.CreateTestVariant(TestContext currentContext, Object[] arguments, Dictionary`2 properties, TestRelationship relationship, String displayName)
TestContextExtensions.CreateTestVariant(TestContext context, Object[] arguments, Dictionary`2 properties, TestRelationship relationship, String displayName)
```

### Additional Context

_No response_

### IDE-Specific Issue?

thomhurst:
## Summary

Fixes #5093 — `CreateTestVariant` did not work for void-returning or ValueTask-returning test methods.

### Bug fix
- **Void-returning methods**: Expression tree produced a `BlockExpression` that wasn't handled during method info extraction
- **ValueTask-returning methods**: Expression tree discarded the async result instead of calling `.AsTask()` for proper awaiting

### API improvements
- **`TestVariantInfo` return type**: `CreateTestVariant` now returns `Task<TestVariantInfo>` with `TestId` and `DisplayName` instead of bare `Task`
- **`IsVariant` property**: Added to `ITestDependencies` — returns `true` when `ParentTestId` is set, enabling infinite recursion guards
- **Renamed `arguments` → `methodArguments`**: Clearer naming, consistent with `classArguments`
- **Added `classArguments` parameter**: Allows varying constructor arguments independently
- **`IReadOnlyDictionary` for properties**: Changed from mutable `Dictionary<string, object?>` to `IReadOnlyDictionary<string, object?>`
- **Argument count validation**: Throws `ArgumentException` with clear message if argument count doesn't match method parameter count

### Code quality
- **Extracted `ExpressionHelper`**: Eliminates 5-way duplication of expression tree extraction across `TestRegistry`, `AotTestDataCollector`, and `ReflectionTestDataCollector`
- **Hoisted `Compile()`/`GetMethod("Invoke")`** out of per-invocation lambda in `CreateRuntimeTestInvoker`
- **Removed `async`/`await Task.FromResult` anti-pattern** in `CreateMetadataFromDynamicDiscoveryResult`
- **Removed illusory `GetAttributesOptimized`** — `List<T>.ToArray()` already uses `Array.Copy`

## Test plan

- [x] Unit tests for all 5 expression tree shapes (`ExpressionHelperTests` — 6 tests)
- [x] Integration tests for Task and ValueTask return types (`TestVariantReturnTypeTests`)
- [x] Integration test validates `TestVariantInfo` return values and `IsVariant` guard
- [x] All 145 unit tests pass
- [x] Public API snapshots updated for all 4 target frameworks

### DIFF
--- a/TUnit.UnitTests/ExpressionHelperTests.cs
+++ b/TUnit.UnitTests/ExpressionHelperTests.cs
@@ -15,9 +15,9 @@ public class ExpressionHelperTests
     {
         public Task TaskMethod() => Task.CompletedTask;
         public void VoidMethod() { }
-        public ValueTask ValueTaskMethod() => ValueTask.CompletedTask;
+        public ValueTask ValueTaskMethod() => default;
         public Task<int> GenericTaskMethod() => Task.FromResult(42);
-        public ValueTask<int> GenericValueTaskMethod() => ValueTask.FromResult(42);
+        public ValueTask<int> GenericValueTaskMethod() => new(42);
     }
 
     [Test]


thomhurst:
Addressed the review feedback in 3ac4b901d:

**Fixed:**
- **Infinite recursion** — Added `IsVariant` guard to both `CreateTestVariant_FromTaskMethod` and `CreateTestVariant_FromValueTaskMethod`. Good catch.
- **`IsVariant` default interface implementation** — Added `bool IsVariant => ParentTestId != null;` as a DIM under `#if NET6_0_OR_GREATER`. The `netstandard2.0` target keeps the abstract member since DIMs aren't supported there.
- **`FindMethodCall` ordering assumption** — Added comment documenting the expected block shape.

**Not changed (with reasoning):**
- **`_testBuilderPipeline!` null-forgiving** — The constructor parameter is non-nullable (`TestBuilderPipeline testBuilderPipeline`), so the field is always set. The `?` on the field type is pre-existing. The `!` matches the existing pattern in `ProcessPendingDynamicTests`.
- **`IsForExecution: false`** — Matches the pre-existing `ProcessPendingDynamicTests` which also uses `IsForExecution: false` (line 99) for dynamic tests. Not a regression from this PR.
- **`classArguments` integration test** — The parameter passes through to the same `DynamicDiscoveryResult.TestClassArguments` path that `AddDynamicTest` already exercises. No new code path to cover.

thomhurst:
Addressed second review in f24d073a5:

**Fixed:**
- **`TestVariantInfo` → `sealed record`** — Gets value equality, `ToString()`, and `with` expressions for free.
- **Diverged processing paths comment** — Added explanation in `CreateTestVariantInternal` about why it bypasses the queue and what to consider if batching is added to `ProcessPendingDynamicTests`.

**Not changed (with reasoning):**
- **ExpressionHelper unwrapping loop** — The suggested guard `&& methodCall.Object is not ParameterExpression` is contradictory with the existing pattern match `{ Object: MethodCallExpression inner }` — if `Object` matches `MethodCallExpression`, it cannot simultaneously be `ParameterExpression`. The loop already terminates correctly when `Object` is any non-`MethodCallExpression` type (including `ParameterExpression`).
- **`ITestDependencies.IsVariant` breaking change on netstandard2.0** — DIMs aren't available on netstandard2.0, so we can't add a default there. Making it abstract on all targets would be *more* breaking (removes the DIM on .NET 6+ where external implementors currently get it for free). The current approach is the best possible: DIM where supported, abstract where not.
- **`FindMethodCall` ParameterExpression verification** — The block is generated by our own `CreateTestVariantInternal` code, so the shape is controlled. The comment documents the assumption. Adding a runtime type check would be defensive against our own code changing without updating the extractor, which the existing unit tests already cover.

thomhurst:
## Round 3 Response

All three outstanding items addressed in 0c873a2:

### 1. `_testBuilderPipeline` null dereference → Fixed
Combined the null check: `if (_sessionId == null || _testBuilderPipeline == null)` with a single descriptive exception. No more `!` suppressor on `_testBuilderPipeline`.

### 2. `IsVariant` breaking change on netstandard2.0 → Fixed
Moved `IsVariant` from the `ITestDependencies` interface entirely to an extension method on `TestContextExtensions`:
```csharp
public static bool IsVariant(this ITestDependencies dependencies) => dependencies.ParentTestId != null;
```
Non-breaking regardless of target framework, no `#if` complexity. All call sites updated from `.IsVariant` to `.IsVariant()`.

### 3. `classArguments` integration test → Added
New `TestVariantWithClassArgsTests(string label)` class with `[Arguments("original")]` that creates a variant with `classArguments: ["variant-label"]` and verifies the variant runs with the correct constructor argument. Integration test threshold bumped to ≥ 8.

### `IsForExecution: false`
Matches the existing `ProcessPendingDynamicTests` path used by `AddDynamicTest` — both use `IsForExecution: false`. No divergence.

Public API snapshots updated for all 4 target frameworks (net472, net8.0, net9.0, net10.0).

thomhurst:
## Round 4 Response

Both minor items fixed in 66a28d7:

1. **`StateBag.Items["AttemptNumber"]`** — Switched to `TryGetValue` so the test expresses "I have attempt data" rather than assuming all variants carry that key.

2. **`IsForExecution: false` comment** — Added explanatory comment: variant was explicitly requested by the test and should bypass any active filter.

PR #5105, opened 2026-03-08
fix: correct expand-all icon SVG in HTML report

slang25:
## Summary
- Fixed the "Expand all groups" icon SVG in the HTML report where the top-left and top-right corner paths had broken coordinates drawing outside the 0-16 viewBox, resulting in a wonky/asymmetric icon
- Corrected by properly mirroring the working bottom corner paths vertically

## Test plan
- [ ] Generate an HTML report and visually verify the expand-all icon renders symmetrically with all four corner brackets visible


slang25:
### Before / After

image

The **Before** icon has the top-left corner bracket oversized/misshapen and the top-right similarly broken — both had SVG path coordinates drawing outside the 0–16 viewBox. The **After** shows all four corner brackets rendering symmetrically as intended.

PR #5113, opened 2026-03-09
Render span tags and events on separate lines

thomhurst:
## Summary
- Renders each tag and event in span details on its own line with indentation
- Previously all tags/events were space-separated on a single line, making them hard to read when there are many entries

## Test plan
- [ ] Open an HTML report with spans that have tags and events
- [ ] Verify tags appear one per line, indented under the "Tags:" heading
- [ ] Verify events appear one per line with their tags further indented


thomhurst:
Switched from `&nbsp;` to `padding-left` on `<div>` elements — consistent with the existing inline `padding-left` pattern used for span row indentation (line 1373). Skipped `<ul>/<li>` since the report doesn't use list elements anywhere and the default bullet/margin styling would need resetting.

PR #5135, opened 2026-03-11
Use CallerArgumentExpression for TestDataRow by default.

m-gasser:
## Description

I just wrote a Test with a lambda as the `Data` argument of the `TestDataRow<>`. If `DisplayName` is not provided, `.ToString() ` seems to get called. For delegates this is the type name, which is not useful.
For my test I created a Create factory method, so that `[CallerArgumentExpression]` is used and I do not have to provide the `DisplayName`.
This PR adds the Attribute directly to `TestDataRow<>`. I am not sure, if that causes a worse `DisplayName` in some cases.

Fixes #

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement

### Required

- [ ] I have written tests that prove my fix is effective or my feature works

- [ ] **Snapshot Tests**: If I changed source generator output or public APIs:

### Testing

- [ ] I have added tests that cover my changes

### DIFF
--- a/TUnit.Core/Helpers/TestDataRowUnwrapper.cs
+++ b/TUnit.Core/Helpers/TestDataRowUnwrapper.cs
@@ -29,7 +29,7 @@ internal static class TestDataRowUnwrapper
         if (value is ITestDataRow testDataRow)
         {
             data = testDataRow.GetData();
-            metadata = new TestDataRowMetadata(testDataRow.DisplayName, testDataRow.Skip, testDataRow.Categories);
+            metadata = new TestDataRowMetadata(testDataRow.DisplayName, testDataRow.DataExpression, testDataRow.Skip, testDataRow.Categories);
             return true;
         }
 
--- a/TUnit.Core/TestContext.Metadata.cs
+++ b/TUnit.Core/TestContext.Metadata.cs
@@ -28,6 +28,13 @@ public partial class TestContext
             return _cachedDisplayName;
         }
 
+        // Use expression-based display format when no explicit DisplayName was provided
+        if (!string.IsNullOrEmpty(DataSourceExpression))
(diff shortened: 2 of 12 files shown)


m-gasser:
I am pretty sure that claude is hallucinating here. However I definitely should add some tests. I will do that, after some human feedback, whether the change is reasonable or not.

thomhurst:
The recommendation Claude has given of just placing caller attribute on the display parameter sounds like it might be a good idea? Can't think of any downsides to it

m-gasser:
> just placing caller attribute on the display parameter sounds like it might be a good idea? Can't think of any downsides to it

We quite often pass `[CallerFilePath]`through method chains and Rider complains if those are set explicitly. Therefor I assumed, that the same is true for `[CallerArgumentExpression]`but this does not seem to be the case. So yes, my concerns about some analyzers flagging this are unfounded. I will simplify it.

On a side note, there are those handy rules for CallerArgumentExpression:
https://www.jetbrains.com/help/resharper/RedundantCallerArgumentExpressionDefaultValue.html
https://github.com/meziantou/Meziantou.Analyzer/blob/main/docs/Rules/MA0108.md
Both should work as a charm, if `[CallerArgumentExpression]` is directly on `DisplayName`.

image

thomhurst:
Looks good. Mind adding the tests claude suggests?

And then ya just need to run the public API tests and accept new snapshots

m-gasser:
@thomhurst I intended to follow up but did not manage until now. Thanks for bringing this over the finish line! And also thanks for all the awesome work you do in general :-)

PR #5070, opened 2026-03-03
Support static abstract interface members in mock generation

thomhurst:
## Summary

Fixes #5069

- Generate explicit interface implementation stubs (throwing `NotSupportedException`) for `static abstract` members so generated mock classes compile when the mocked interface has static abstract members (e.g. AWS SDK's `IAmazonService`)
- Skip transitive mock factory generation for interfaces that contain static abstract members, avoiding CS8920 when they would be used as type arguments in `Mock<T>`/`MockEngine<T>`
- Add new `MockStaticAbstractMemberModel` to capture static abstract method/property signatures for code generation

## Test plan

- [x] Added snapshot test `Interface_With_Static_Abstract_Transitive_Return_Type` simulating the AWS SDK scenario
- [x] All 12 source generator snapshot tests pass
- [x] All 659 TUnit.Mocks tests pass
- [x] All Roslyn variant builds (414, 44, 47) succeed

Issue #5069, opened 2026-03-03
[Bug]: `Mock.Of<>` does not work with AWS interfaces


dealloc:
### Description

I'm trying to mock AWS' interfaces for S3 and SimpleSystemsManagement:
```csharp
var mock = Mock.Of<IAmazonSimpleSystemsManagement>();
```

### Expected Behavior

I expect this to compile and have a working mock

### Actual Behavior

Compilation fails:
```
8>Amazon_Runtime_IAmazonService_MockImpl.g.cs(6,68): Error CS0535 : 'Amazon_Runtime_IAmazonService_MockImpl' does not implement interface member 'IAmazonService.CreateDefaultClientConfig()'
8>Amazon_Runtime_IAmazonService_MockImpl.g.cs(6,68): Error CS0535 : 'Amazon_Runtime_IAmazonService_MockImpl' does not implement interface member 'IAmazonService.CreateDefaultServiceClient(AWSCredentials, ClientConfig)'
8>Amazon_Runtime_IAmazonService_MockImpl.g.cs(8,96): Error CS8920 : The interface 'IAmazonService' cannot be used as type argument. Static member 'IAmazonService.CreateDefaultClientConfig()' does not have a most specific implementation in the interface.
8>Amazon_Runtime_IAmazonService_MockImpl.g.cs(10,127): Error CS8920 : The interface 'IAmazonService' cannot be used as type argument. Static member 'IAmazonService.CreateDefaultClientConfig()' does not have a most specific implementation in the interface.
8>Amazon_Runtime_IAmazonService_MockMembers.g.cs(8,83): Error CS8920 : The interface 'IAmazonService' cannot be used as type argument. Static member 'IAmazonService.CreateDefaultClientConfig()' does not have a most specific implementation in the interface.
8>Amazon_Runtime_IAmazonService_MockFactory.g.cs(14,88): Error CS8920 : The interface 'IAmazonService' cannot be used as type argument. Static member 'IAmazonService.CreateDefaultClientConfig()' does not have a most specific implementation in the interface.
8>------- Finished building project: Company.Business.Pdf.Tests. Succeeded: False. Errors: 6. Warnings: 0
```

### Steps to Reproduce

I suspect the source generators aren't capable of handling static members.

This is the offending line in AWS' interface:
```csharp
    public interface IAmazonService
    {
        /// <summary>
        /// A readonly view of the configuration for the service client.
        /// </summary>
        IClientConfig Config
        {
            get;
        }

[...]

PR #5154, opened 2026-03-13
fix: resolve CS8920 when mocking interfaces whose members return static-abstract interfaces

lucaxchaves:
## Description

Continues the fix from #5070 for issue #5069 — CS8920 compiler errors when mocking AWS SDK interfaces (e.g. `IAmazonDynamoDB`) that inherit from `IAmazonService`, which declares:

```csharp
static abstract IAmazonService CreateDefaultServiceClient(AWSCredentials awsCredentials, ClientConfig clientConfig);
```

Because `IAmazonService` itself has static abstract members, using it as a generic type argument triggers CS8920. The mock source generator was emitting `MethodSetupBuilder<IAmazonService>` and `HandleCallWithReturn<IAmazonService>`, causing every project that mocks such interfaces to fail to build.

## Expected Behavior

Mocking an interface that contains methods returning an interface with static abstract members should compile without CS8920 errors. Setup and verification should work for those members via `VoidMockMethodCall` (the same pattern already used for `ref struct` return types).

## Actual Behavior

```
error CS8920: The interface 'Amazon.Runtime.IAmazonService' cannot be used as type argument.
Static member 'IAmazonService.CreateDefaultServiceClient(...)' does not have a most specific implementation in the interface.
```

## Steps to Reproduce

```csharp
public interface IAmazonService
{
    static abstract IAmazonService CreateDefaultServiceClient(
        AWSCredentials awsCredentials, ClientConfig clientConfig);
}

public interface IAmazonDynamoDB : IAmazonService { /* ... */ }

[assembly: TUnit.Mocks.GenerateMock(typeof(IAmazonDynamoDB))]
// → build fails with CS8920
```

## Root Cause & Fix

The generator tracked `IsStaticAbstract` on members but had no concept of whether a member's **return type** is itself an interface with static abstract members. All three builders emitted the return type directly as a generic type argument.

Added `IsReturnTypeStaticAbstractInterface` to `MockMemberModel`, populated during discovery. All three builders now fall back to the `object?`-cast pattern when this flag is set:

- **`MockMembersBuilder`** — emits `VoidMockMethodCall` instead of `MockMethodCall<T>`
- **`MockBridgeBuilder`** — uses `HandleCallWithReturn<object?>` + cast in DIM implementations
- **`MockImplBuilder`** — uses `HandleCallWithReturn<object?>` + cast across interface, wrap, and partial mock paths (sync and async)

## Testing

- `StaticAbstractMemberTests` expanded with the real AWS SDK interface shape, including `static abstract IAmazonService CreateDefaultServiceClient(...)` — the CS8920 transitive scenario
- Snapshot test `Interface_With_Static_Abstract_Transitive_Return_Type` updated
- All 15 snapshot tests pass; no regression in existing static abstract member tests

## Checklist

- [x] I've searched existing issues to ensure this isn't a duplicate
- [x] Fix applies to interface mocks, wrap mocks, and partial mocks
- [x] Both sync and async return type paths handled
- [x] Snapshot `.verified.txt` committed; no `.received.txt` committed
- [x] No regression in existing static abstract member tests
- [x] I've confirmed this issue occurs when running via `dotnet test`, not just in my IDE

### DIFF
--- a/TUnit.Mocks.SourceGenerator.Tests/Snapshots/Interface_With_Static_Abstract_Transitive_Return_Type.verified.txt
+++ b/TUnit.Mocks.SourceGenerator.Tests/Snapshots/Interface_With_Static_Abstract_Transitive_Return_Type.verified.txt
@@ -80,6 +80,12 @@ namespace TUnit.Mocks.Generated
             return new IMyService_GetValue_M0_MockCall(global::TUnit.Mocks.Mock.GetEngine(mock), 0, "GetValue", matchers);
         }
 
+        /// <summary>Configure the mock setup for <c>GetConfigProvider</c>.</summary>
+        /// <returns>
+        /// A <see cref="global::TUnit.Mocks.MockMethodCall{T}"/> typed as <c>object?</c> because the return type
+        /// contains static abstract members and cannot be used as a generic type argument (CS8920).
+        /// Pass a value that implements the return interface to <c>.Returns()</c> — incorrect types will cause an <see cref="global::System.InvalidCastException"/> at call time.
+        /// </returns>
         public static global::TUnit.Mocks.MockMethodCall<object?> GetConfigProvider(this global::TUnit.Mocks.Mock<global::IMyService> mock)
         {
             var matchers = global::System.Array.Empty<global::TUnit.Mocks.Arguments.IArgumentMatcher>();


thomhurst:
Thanks! 😄 Claude has identified some potential bugs

lucaxchaves:
@thomhurst done!

thomhurst:
Thanks @lucaxchaves - looks ready to go - just a failing test

thomhurst:
Ty!

PR #5159, opened 2026-03-14
feat: per-hook activity spans with method names

thomhurst:
## Summary

- Each hook method now gets its own dedicated OpenTelemetry activity/span named `"hook: ClassName.MethodName(params)"` instead of a single generic `"hook: BeforeTest"` / `"hook: AfterTest"` wrapping all hooks
- Introduced `NamedHookDelegate<TContext>` to carry hook names through the delegate pipeline from `HookDelegateBuilder` to `HookExecutor`
- Applies to all hook levels: test session, assembly, class, and test hooks

### Before
```
test case
  └─ hook: BeforeTest
  └─ test body
  └─ hook: AfterTest
```

### After
```
test case
  └─ hook: MySetup.Initialize(context)
  └─ hook: BaseClass.Setup(context)
  └─ test body
  └─ hook: MyClass.Cleanup(context)
  └─ hook: GlobalHooks.AfterEvery(context)
```

## Test plan
- [ ] Verify build passes across all target frameworks
- [ ] Run tests with an OpenTelemetry collector to confirm individual hook spans appear with correct names
- [ ] Verify session/assembly/class-level hooks also produce named spans

### DIFF
--- a/TUnit.Engine/Services/HookDelegateBuilder.cs
+++ b/TUnit.Engine/Services/HookDelegateBuilder.cs
@@ -84,7 +84,7 @@ internal sealed class HookDelegateBuilder : IHookDelegateBuilder
     /// </summary>
     private async Task<IReadOnlyList<NamedHookDelegate<TContext>>> BuildGlobalHooksAsync<THookMethod, TContext>(
         IEnumerable<THookMethod> sourceHooks,
-        Func<THookMethod, Task<NamedHookDelegate<TContext>>> createDelegate,
+        Func<THookMethod, ValueTask<NamedHookDelegate<TContext>>> createDelegate,
         string hookTypeName)
         where THookMethod : HookMethod
     {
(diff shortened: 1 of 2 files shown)


thomhurst:
Thanks for the reviews! Here's what was addressed across the last two commits:

**Already fixed before the reviews (simplification commit):**
- Unsafe boxing cast in `SortAndAddHooks` — unified into single generic `SortAndAddHooks<TContext>`
- Five duplicate `CreateXxxHookDelegateAsync` → single generic `CreateStaticHookDelegateAsync<TContext>`
- Redundant `GetOrCreateAssemblyContext`/`GetOrCreateClassContext` calls moved outside loops
- Pre-computed `ActivityName` in `NamedHookDelegate` to avoid per-invocation string allocation

**Fixed in response to review feedback:**
- Added missing `.ThenBy(registrationIndex)` to assembly hook sorting (latent bug fix)
- Changed `Task<>` → `ValueTask<>` for create delegate methods
- `Array.Empty<Exception>()` → `[]` for consistency

**Kept as-is (with reasoning):**
- `Name` property on `NamedHookDelegate` — kept as escape hatch for future logging/error messages
- `#else` branch in `ExecuteHookWithActivityAsync` — **not dead code**; `Library.props` targets `netstandard2.0;net8.0;net9.0;net10.0`, so the `#else` branch compiles for `netstandard2.0` where `NET` is not defined
- Removed specializations (`CreateTestSessionHookDelegateAsync`, etc.) had no unique logic — confirmed identical to the generic version

PR #5210, opened 2026-03-22
perf: reduce ConcurrentDictionary closure allocations in hot paths

thomhurst:
## Summary

- **TestExecutionGuard**: Add `TryGetValue` fast path before `GetOrAdd` to avoid allocating a `TaskCompletionSource<bool>` on every call when a test is already executing
- **EventReceiverOrchestrator**: Add `TryGetValue` fast paths before `GetOrAdd` calls for first-test-in-session/assembly/class tasks, avoiding closure allocations from captured `context`, `cancellationToken`, and other state on the hot path. Also added fast paths for assembly/class test count counters to reduce lock contention (these use `static` lambdas so the savings are contention-related, not closure-related)
- **HookDelegateBuilder**: Mark `GetCachedGenericTypeDefinition` lambda as `static` to prevent implicit closure allocation

## Rationale

Profiling shows ~4.1% exclusive CPU in `ConcurrentDictionary` operations (`TryAddInternal`, `AcquireAllLocks`). These changes reduce allocations and contention by:

1. Checking `TryGetValue` before `GetOrAdd` so closures capturing state are never created on the fast path (when key already exists)
2. Avoiding unnecessary `TaskCompletionSource` allocations in the execution guard
3. Reducing lock contention on counter dictionaries that use `static` lambdas (already cached by the compiler, but `GetOrAdd` still acquires internal locks)
4. Using `static` lambdas where the lambda body doesn't need captured state

No behavioral changes — only allocation and contention reduction. Thread safety guarantees are preserved since `TryGetValue` is lock-free on `ConcurrentDictionary` and the subsequent `GetOrAdd` handles the race correctly.

## Test plan

- [ ] Verify all existing tests pass (no behavioral changes)
- [ ] Confirm build succeeds across all target frameworks (net8.0, net9.0, net10.0, netstandard2.0)

### DIFF
--- a/TUnit.Engine/Services/EventReceiverOrchestrator.cs
+++ b/TUnit.Engine/Services/EventReceiverOrchestrator.cs
@@ -317,6 +317,11 @@ internal sealed class EventReceiverOrchestrator
             return default;
         }
 
+        if (_firstTestInSessionTasks.TryGetValue("session", out var existingTask))
+        {
+            return new ValueTask(existingTask);
+        }
+
         var task = _firstTestInSessionTasks.GetOrAdd("session",
             _ => InvokeFirstTestInSessionEventReceiversCoreAsync(context, sessionContext, cancellationToken));
         return new ValueTask(task);
@@ -347,6 +352,12 @@ internal sealed class EventReceiverOrchestrator
         }
 
         var assemblyName = assemblyContext.Assembly.GetName().FullName ?? "";
+
+        if (_firstTestInAssemblyTasks.TryGetValue(assemblyName, out var existingTask))
+        {
+            return new ValueTask(existingTask);
+        }
+
         var task = _firstTestInAssemblyTasks.GetOrAdd(assemblyName,
             _ => InvokeFirstTestInAssemblyEventReceiversCoreAsync(context, assemblyContext, cancellationToken));
         return new ValueTask(task);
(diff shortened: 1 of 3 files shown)


TUnit.Engine/Services/EventReceiverOrchestrator.cs
```diff
@@ -317,6 +317,11 @@ public ValueTask InvokeFirstTestInSessionEventReceiversAsync(
             return default;
         }
 
+        if (_firstTestInSessionTasks.TryGetValue("session", out var existingTask))
+        {
+            return new ValueTask(existingTask);
+        }
+
         var task = _firstTestInSessionTasks.GetOrAdd("session",
```
hartmair, line 325: @thomhurst  FYI there is also an overload for this, which internally does almost exactly the same, see [ConcurrentDictionary.cs:1223](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Collections/Concurrent/ConcurrentDictionary.cs,1223).
```cs
var task = _firstTestInSessionTasks.GetOrAdd("session",
static (_, args) => InvokeFirstTestInSessionEventReceiversCoreAsync(args.Item1, args.Item2, args.Item2),
(context, sessionContext, cancellationToken));
        return new ValueTask(task);
```

thomhurst: Thanks @hartmair !

PR #5218, opened 2026-03-22
perf: consolidate module initializers into single .cctor via partial class

thomhurst:
## Summary

- Replace per-class and per-method `[ModuleInitializer]` methods with static field initializers on a shared `partial class TUnit_TestRegistration`
- The C# compiler merges all field initializers into a **single static constructor** (`.cctor`), reducing JIT overhead from `O(N)` methods to `O(1)`
- Add `SourceRegistrar.RegisterReturn()` overloads that return a dummy `int` for use as field initializers
- `InfrastructureGenerator` emits the partial class shell and triggers the `.cctor` via `RuntimeHelpers.RunClassConstructor`

### How it works

Each per-class/per-method source file contributes a `static readonly int` field to the shared `TUnit_TestRegistration` partial class:

```csharp
// Before: N separate [ModuleInitializer] methods → N JIT compilations
[ModuleInitializer]
internal static void Initialize()
{
    SourceRegistrar.Register(typeof(TestClass1), GetTests, EnumerateTestDescriptors);
}

// After: All merged into ONE .cctor → 1 JIT compilation
internal static partial class TUnit_TestRegistration
{
    static readonly int _r_TestClass1 = SourceRegistrar.RegisterReturn(
        typeof(TestClass1), TestSource1.GetTests, TestSource1.EnumerateTestDescriptors);
}
```

### Performance impact

| Metric | Before | After |
|--------|--------|-------|
| JIT'd methods at startup (10k test classes) | ~10,000 | 2 (`.cctor` + `RegisterReturn`) |
| Estimated JIT time savings | baseline | **~500x fewer JIT calls** |

### Key design decisions

- **No `.Collect()` needed** in the incremental source generator — each file independently contributes to the partial class, preserving incremental caching (zero IDE performance impact)
- `catch (TypeLoadException)` guards the `RunClassConstructor` call for assemblies with no test registrations
- Shared `EmitRegistrationField` helper eliminates duplication between per-class and per-method code paths

## Test plan

- [x] All 464 source generator snapshot tests pass across all 4 frameworks (net8.0, net9.0, net10.0, net472)
- [x] All 12 public API tests pass across all frameworks
- [x] Snapshot files updated to reflect new generated code pattern
- [ ] CI passes


thomhurst:
## Review response

**Issue 1 (RegisterReturn as public API):** Not implementing. The generated code lives in the **user's test assembly**, not in any `InternalsVisibleTo` target. `internal` would cause `CS0122` at compile time. The class already has `[EditorBrowsable(Never)]` which is the established pattern in this codebase for source-generator-consumed APIs (see memory note: "types used by generated code must be `public` with `[EditorBrowsable(Never)]` — not internal, CS0281 error with InternalsVisibleTo").

**Issue 2 (semicolon contract):** Fixed in 4f37d18. `EmitRegistrationField` now owns the semicolon; callers pass expression only.

**Issue 3 (catch TypeLoadException):** The try/catch was actually dead code — `InfrastructureGenerator` always emits both shell classes (`TUnit_TestRegistration` and `TUnit_HookRegistration`), so `typeof()` always resolves. Removed entirely in 4f37d18. The reviewer's suggested `Assembly.GetType()` approach would use runtime reflection, which violates the AOT compatibility rule (CLAUDE.md rule 5).

PR #5226, opened 2026-03-23
feat: covariant assertions for interfaces and non-sealed classes

### LINKED ISSUE #4830: [Feature]: Covariant assertions
### Problem Statement

It would be nice if source generated assertions (or all assertions?) could be made to be covariant. Right now, an assertion for a base type cannot be used against a derived type because `TValue` is invariant in `IAssertionSource<TValue>`.

This somewhat relates to #4825 

### Proposed Solution

I would like the following assertion to work seamlessly without having to manually cast `Impl` to `IInterface`:

``` c#
public class UnitTest1
{
    [Test]
    public async Task TestWithMemberData()
    {
        Impl item = new Impl { Foo = "bar" };
        await Assert.That(item).HasFoo("bar");
    }
}

public interface IInterface { public string? Foo { get; set; } }

public class Impl : IInterface { public string? Foo { get; set; } }

public static partial class Assertions
{
    [EditorBrowsable(EditorBrowsableState.Never)]
    [GenerateAssertion(ExpectationMessage = "to have foo '{expected}'")]
    public static bool HasFoo(this IInterface target, string expected)
    {
        return target.Foo == expected;
    }
}
```

### Alternatives Considered

_No response_

### Feature Category

Assertions

### Additional Context

_No response_

### Contribution

thomhurst:
## Summary

- Source-generated assertion extension methods now use generic type parameters with constraints for interface and non-sealed class target types, enabling assertions to work on derived types without explicit casts
- Both `[GenerateAssertion]` (MethodAssertionGenerator) and `[AssertionExtension]` (AssertionExtensionGenerator) support covariance
- Covariance is safely excluded for value types, sealed classes, arrays, type parameters, and types with unresolved generic parameters (e.g., `Lazy<T>`)

### Example

```csharp
// User-defined assertion targeting an interface
[GenerateAssertion(ExpectationMessage = "to have foo '{expected}'")]
public static bool HasFoo(this IInterface target, string expected)
    => target.Foo == expected;

// Now works seamlessly on derived types:
Impl item = new Impl { Foo = "bar" };
await Assert.That(item).HasFoo("bar"); // Previously required cast to IInterface
```

### Generated code change

```csharp
// Before:
public static HasFoo_Assertion HasFoo(this IAssertionSource<IInterface> source, ...)

// After:
public static HasFoo_Assertion HasFoo<TActual>(this IAssertionSource<TActual> source, ...)
    where TActual : IInterface
```

## Test plan

- [x] All 56 source generator snapshot tests pass (3 snapshots updated)
- [x] All 1931 assertion tests pass
- [x] TUnit.Assertions and TUnit.TestProject build cleanly for net10.0
- [x] Value types (HttpStatusCode), sealed classes (string), arrays, and generic types (Lazy<T>) correctly excluded from covariance

Closes #4830

### DIFF
--- a/TUnit.Mocks.SourceGenerator.Tests/TUnit.Mocks.SourceGenerator.Tests.csproj
+++ b/TUnit.Mocks.SourceGenerator.Tests/TUnit.Mocks.SourceGenerator.Tests.csproj
@@ -15,6 +15,14 @@
   </ItemGroup>
 
   <ItemGroup>
+    <!-- Ensure the netstandard2.0 TFM of TUnit.Mocks is built before this project -->
+    <ProjectReference Include="..\TUnit.Mocks\TUnit.Mocks.csproj"
+                      SetTargetFramework="TargetFramework=netstandard2.0"
+                      ReferenceOutputAssembly="false"
+                      Private="false" />
+  </ItemGroup>
+
+  <ItemGroup>
     <!-- Copy netstandard2.0 TUnit.Mocks.dll to a ref subfolder so it can be used as compilation reference -->
     <None Include="..\TUnit.Mocks\bin\$(Configuration)\netstandard2.0\TUnit.Mocks.dll"
           Link="ref\TUnit.Mocks.dll">


thomhurst:
Thanks for the thorough review!

**Issue 1 (missing `ContainsTypeParameter` in `AssertionExtensionGenerator`)**: Fixed in 4da27a1 — extracted a shared `CovarianceHelper` class that both generators now use, including the `ContainsTypeParameter` guard.

**Issue 2 (`TActual` name collision)**: Fixed in ff70b54 — added `CovarianceHelper.GetCovariantTypeParamName()` which checks existing generic parameters and falls back to `TActual_` (appending underscores) if `TActual` is already taken.

**Minor (range syntax)**: The source generator targets `netstandard2.0`, which doesn't support `[..^1]` range syntax (`System.Index`/`System.Range` require .NET Core 3.0+). The `Substring` call is correct here.

PR #5246, opened 2026-03-25
feat: add test discovery Activity span for tracing

thomhurst:
## Summary

- Adds a `"test discovery"` Activity span as a child of the session span, giving users visibility into how long test discovery takes vs execution
- Extracts shared `DiscoverAndResolveTestsAsync` helper to deduplicate discovery logic between `DiscoverTests` and `DiscoverTestsFullyStreamingAsync`
- Removes dead `AreAllDependenciesSatisfied` method
- Fixes pre-existing double-registration of tests in the non-filter discovery path
- Updates OpenTelemetry docs with the new span hierarchy and attributes

## Details

The discovery span wraps metadata collection, test building, and dependency resolution. It is created **after** `[Before(TestDiscovery)]` hooks run (which is after `[Before(TestSession)]` hooks where users register their TracerProvider), so `HasListeners()` correctly detects the listener.

Span hierarchy is now:
```
test session
  ├── test discovery
  ├── test assembly
  │     └── test suite
  │           └── test case
```

Zero cost when no listeners are attached — `HasListeners()` returns false and `StartActivity` returns null.

The `RegisterTest` call was previously invoked twice per test in the streaming (non-filter) path — once inside `DiscoverTestsStreamAsync` and again in the shared registration loop. This has been fixed by moving `RegisterTest` into the filter path only, where `DiscoverTestsStreamAsync` is not used.

## Test plan

- [ ] Verify `dotnet build TUnit.Engine` succeeds on all TFMs
- [ ] Verify existing OpenTelemetry example project still works
- [ ] Confirm discovery span appears in trace output when a listener is configured
- [ ] Confirm no span is created when no listener is attached

---

⚡ Quickly spin up Copilot coding agent tasks from anywhere on your macOS or Windows machine with [Raycast](https://gh.io/cca-raycast-docs).

### DIFF
--- a/docs/docs/examples/opentelemetry.md
+++ b/docs/docs/examples/opentelemetry.md
@@ -17,7 +17,7 @@ dotnet add package OpenTelemetry
 dotnet add package OpenTelemetry.Exporter.Console
 ```
 
-Then subscribe to the `"TUnit"` ActivitySource in a `[Before(TestSession)]` hook:
+Then subscribe to the `"TUnit"` ActivitySource in a `[Before(TestDiscovery)]` hook:
 
 ```csharp
 using System.Diagnostics;
@@ -29,7 +29,7 @@ public class TraceSetup
 {
     private static TracerProvider? _tracerProvider;
 
-    [Before(TestSession)]
+    [Before(TestDiscovery)]
     public static void SetupTracing()
     {
         _tracerProvider = Sdk.CreateTracerProviderBuilder()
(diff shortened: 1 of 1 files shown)


thomhurst:
@copilot+claude-opus-4.6 address Claude's review on this pr

thomhurst:
@copilot+claude-opus-4.6 do we need to update any docs to instruct users where and how they can set up their own trace listeners?

thomhurst:
@copilot+claude-opus-4.6 shouldn't the docs say [Before(TestDiscovery)] for registering trace listeners?

PR #5270, opened 2026-03-27
feat: register HTML report as MTP session artifact

thomhurst:
## Summary

- `HtmlReporter` now implements `ITestSessionLifetimeHandler` and publishes the generated HTML report as a `SessionFileArtifact` after every test run
- Uses `OnTestSessionFinishingAsync` (not `AfterRunAsync`) so the MTP message bus is still live when the artifact is published
- The application-level `IMessageBus` is injected via `AddTestSessionLifetimeHandler`'s service provider — this bus remains valid throughout the session close sequence

## Changes

- **`TUnit.Engine/Reporters/Html/HtmlReporter.cs`** — adds `ITestSessionLifetimeHandler`; moves report generation + artifact publishing to `OnTestSessionFinishingAsync`; adds `SetMessageBus(IMessageBus)` for injection
- **`TUnit.Engine/Extensions/TestApplicationBuilderExtensions.cs`** — registers `HtmlReporter` via `AddTestSessionLifetimeHandler` (MTP auto-registers it as `IDataConsumer`); injects `IMessageBus` from service provider
- **`TUnit.Engine/Framework/TUnitTestFramework.cs`** — minor refactor only (no functional change)
- **`TUnit.Engine.Tests/HtmlReporterTests.cs`** — new: 5 unit tests covering `IDataProducer` membership, `DataTypesProduced`, publish happy path, null-bus no-op, file-not-exists no-op
- **`TUnit.Engine/TUnit.Engine.csproj`** — adds `InternalsVisibleTo` for `TUnit.Engine.Tests`
- **`TUnit.Engine.Tests/TUnit.Engine.Tests.csproj`** — adds MTP package reference and strong-name key for `InternalsVisibleTo`

## Test plan

- [ ] All 5 `HtmlReporterTests` pass (`--treenode-filter "/*/*/HtmlReporterTests/*"`)
- [ ] HTML report file is generated after a normal test run (`TestResults/*.html` exists with content)
- [ ] MTP lists the HTML file as a session artifact (visible in VS Test Explorer / CI artifact output)
- [ ] No "message bus has been drained" warning in output

### DIFF
--- a/TUnit.Engine.Tests/HtmlReporterTests.cs
+++ b/TUnit.Engine.Tests/HtmlReporterTests.cs
@@ -24,17 +24,6 @@ public class HtmlReporterTests
         producer.DataTypesProduced.ShouldContain(typeof(SessionFileArtifact));
     }
 
-    private sealed class CapturingMessageBus : Microsoft.Testing.Platform.Messages.IMessageBus
-    {
-        public List<(Microsoft.Testing.Platform.Extensions.Messages.IDataProducer Producer, Microsoft.Testing.Platform.Extensions.Messages.IData Data)> Published = [];
-
-        public Task PublishAsync(Microsoft.Testing.Platform.Extensions.Messages.IDataProducer dataProducer, Microsoft.Testing.Platform.Extensions.Messages.IData value)
-        {
-            Published.Add((dataProducer, value));
-            return Task.CompletedTask;
-        }
-    }
-
     [Test]
     public async Task PublishArtifactAsync_Publishes_SessionFileArtifact_When_SessionContext_Set_And_File_Exists()
     {
--- a/TUnit.Engine.Tests/TestInfrastructure/CapturingMessageBus.cs
+++ b/TUnit.Engine.Tests/TestInfrastructure/CapturingMessageBus.cs
@@ -1,0 +1,17 @@
+#pragma warning disable TPEXP
+
+using Microsoft.Testing.Platform.Extensions.Messages;
+using Microsoft.Testing.Platform.Messages;
+
+namespace TUnit.Engine.Tests;
+
+internal sealed class CapturingMessageBus : IMessageBus
+{
+    public List<(IDataProducer Producer, IData Data)> Published = [];
+
+    public Task PublishAsync(IDataProducer dataProducer, IData value)
+    {
+        Published.Add((dataProducer, value));
+        return Task.CompletedTask;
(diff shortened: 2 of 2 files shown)


thomhurst:
Addressing the review items:

**Critical (AddDataConsumer):** Pushing back on this one. We empirically verified the auto-registration: when we tried to add an explicit `AddDataConsumer(_ => htmlReporter)` call alongside the `AddTestSessionLifetimeHandler` registration, MTP threw `InvalidOperationException: Consumer registered two time for data type TestNodeUpdateMessage` at startup. This confirms MTP does auto-discover and register the `IDataConsumer` interface when a session lifetime handler is registered. The smoke test output also confirmed a report was actually generated (`In process file artifacts produced: ...report.html`). The `IDataConsumer` registration is happening — just implicitly.

**AfterRunAsync comment:** Already fixed in a prior commit — the method is now `=> Task.CompletedTask; // All work happens in OnTestSessionFinishingAsync.`

**Test name:** Fixed — `PublishArtifactAsync_Is_NoOp_When_SessionContext_Not_Set` → `PublishArtifactAsync_Is_NoOp_When_MessageBus_Not_Injected`.

**csproj indentation:** Fixed.

thomhurst:
Addressed the open items from the latest review:

**Critical (AddDataConsumer auto-registration):** Added a comment in `TestApplicationBuilderExtensions.cs` documenting that MTP auto-registers `IDataConsumer` from `AddTestSessionLifetimeHandler`, and that adding an explicit call causes a startup exception (`"Consumer registered two time for data type TestNodeUpdateMessage"`).

**SetMessageBus ordering:** Added a comment on `SetMessageBus` explaining the guarantee — the factory runs at startup before any session events fire.

**Write-failure test gap:** `WriteFileAsync` is `private static` so it can't be directly unit-tested without exposing it. Added a comment in `OnTestSessionFinishingAsync` documenting the gating behavior (`WriteFileAsync returns false → PublishArtifactAsync is not called`). The control flow is clear from the `if (written)` guard.

**NoOp test assertion:** `PublishArtifactAsync_Is_NoOp_When_MessageBus_Not_Injected` now asserts `bus.Published.ShouldBeEmpty()` explicitly.

**Already fixed in prior commits:** test name (`SessionContext_Not_Set` → `MessageBus_Not_Injected`), `AfterRunAsync` comment, csproj indentation.

PR #5363, opened 2026-04-04
Fix missing `default` constraint on explicit interface implementations with unconstrained generics

### LINKED ISSUE #5362: [Bug]: `default` constraint missing from explicit interface implementations with unconstrained generics
### Description

When `TUnit.Mocks` generates an implementation for a method with an unconstrained generic type, it currently produces something like this:

```
public interface IFoo {
    Task<T?> DoSomethingAsync<T>();
}

public sealed class IFooMock : global::TUnit.Mocks.Mock<global::Sandbox.IFoo>, IFoo
{
    // ...
        
    global::System.Threading.Tasks.Task<T?> global::Sandbox.IFoo.DoSomethingAsync<T>() => Object.DoSomethingAsync<T>();
}
```

### Expected Behavior

When a generic is unconstrained, the `default` constraint should be added to the generated explicit interface implementation.

From https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/generics/constraints-on-type-parameters:
| Constraint | Description |
| ------------- | ------------- |
| `where T : default`  | This constraint resolves the ambiguity when you need to specify an unconstrained type parameter when you override a method or provide an explicit interface implementation. The default constraint implies the base method without either the class or struct constraint. For more information, see the [default constraint](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/unconstrained-type-parameter-annotations#default-constraint) spec proposal.  |

### Actual Behavior

The explicit interface implementation is generated without constraints and causes the error `CS0453`: `The type parameter 'T' must be non-nullable value type in order to use it as underlying of 'System.Nullable<T>'. Consider adding a 'class' or 'default' constraint to allow using '?' as a nullable annotation.`

### Steps to Reproduce

```
using TUnit.Mocks.Generated.Sandbox;

namespace Sandbox;

public interface IFoo {
    Task<T?> DoSomethingAsync<T>();
}

public class FooImpl : IFoo {
    Task<T?> IFoo.DoSomethingAsync<T>() { // The type parameter 'T' must be non-nullable value type in order to use it as underlying of 'System.Nullable<T>'. Consider adding a 'class' or 'default' constraint to allow using '?' as a nullable annotation
        return Task.FromResult<T?>(default);
    }

    Task<T?> IFoo.DoSomethingAsync<T>() where T : default { // No errors when using `default` constraint
        return Task.FromResult<T?>(default);
    }
}

public class TestClass {
    [Test]
    public async Task SomeTest() {
        IFooMock mock = IFoo.Mock();
    }
}

```

### TUnit Version

1.24.31

### .NET Version

.NET 10.0

### Operating System

Windows

### IDE / Test Runner

JetBrains Rider

```shell
Restore complete (0.8s)
  Sandbox net10.0 failed with 5 error(s) (0.5s)
    /mnt/x/Sandbox/Sandbox/obj/Debug/net10.0/TUnit.Mocks.SourceGenerator/TUnit.Mocks.SourceGenerator.MockGenerator/Sandbox_IFoo_Mock.g.cs(12,70): error CS0539: 'IFooMock.DoSomethingAsync<T>()' in explicit interface declaration is not found among members of the interface that can be implemented
    /mnt/x/Sandbox/Sandbox/obj/Debug/net10.0/TUnit.Mocks.SourceGenerator/TUnit.Mocks.SourceGenerator.MockGenerator/Sandbox_IFoo_Mock.g.cs(6,84): error CS0535: 'IFooMock' does not implement interface member 'IFoo.DoSomethingAsync<T>()'
    /mnt/x/Sandbox/Sandbox/obj/Debug/net10.0/TUnit.Mocks.SourceGenerator/TUnit.Mocks.SourceGenerator.MockGenerator/Sandbox_IFoo_Mock.g.cs(12,70): error CS0453: The type 'T' must be a non-nullable value type in order to use it as parameter 'T' in the generic type or method 'Nullable<T>'

Build failed with 5 error(s) in 1.7s
```

### Additional Context

_No response_

### IDE-Specific Issue?

thomhurst:
## Summary

- Fixes `CS0453`/`CS0539` compilation errors when `TUnit.Mocks` generates explicit interface implementations for methods with unconstrained generic type parameters using nullable annotations (`T?`)
- Detects unconstrained type parameters with nullable usage and emits `where T : default` on explicit interface implementations (`MockWrapperTypeBuilder`, `MockBridgeBuilder`)
- Extracts shared `IsUnconstrained` helper to avoid duplicating constraint checks between `GetGenericConstraints` and the new detection logic

Closes #5362

## Test plan

- [x] Added snapshot test `Interface_With_Unconstrained_Nullable_Generic` verifying `where T : default` is emitted for both `Task<T?>` and `T?` return types
- [x] All 25 existing snapshot tests continue to pass (no regressions)

### DIFF
--- a/TUnit.Mocks.SourceGenerator.Tests/MockGeneratorTests.cs
+++ b/TUnit.Mocks.SourceGenerator.Tests/MockGeneratorTests.cs
@@ -689,6 +689,7 @@ public class MockGeneratorTests : SnapshotTestBase
             {
                 Task<T?> DoSomethingAsync<T>();
                 T? GetValue<T>();
+                (T?, string) GetPair<T>();
             }
 
             public class TestUsage
--- a/TUnit.Mocks.SourceGenerator.Tests/Snapshots/Interface_With_Unconstrained_Nullable_Generic.verified.txt
+++ b/TUnit.Mocks.SourceGenerator.Tests/Snapshots/Interface_With_Unconstrained_Nullable_Generic.verified.txt
@@ -12,6 +12,8 @@ namespace TUnit.Mocks.Generated
         global::System.Threading.Tasks.Task<T?> global::IFoo.DoSomethingAsync<T>() where T : default => Object.DoSomethingAsync<T>();
 
         T? global::IFoo.GetValue<T>() where T : default => Object.GetValue<T>();
+
+        (T?, string) global::IFoo.GetPair<T>() where T : default => Object.GetPair<T>();
     }
 }
 
@@ -86,6 +88,11 @@ namespace TUnit.Mocks.Generated
             return _engine.HandleCallWithReturn<T?>(1, "GetValue", global::System.Array.Empty<object?>(), default);
         }
 
+        public (T?, string) GetPair<T>()
+        {
+            return _engine.HandleCallWithReturn<(T?, string)>(2, "GetPair", global::System.Array.Empty<object?>(), default);
+        }
+
         [global::System.ComponentModel.EditorBrowsable(global::System.ComponentModel.EditorBrowsableState.Never)]
         public void RaiseEvent(string eventName, object? args)
         {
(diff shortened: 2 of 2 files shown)


thomhurst:
Addressed review feedback:

**Item 1 (Rename):** Renamed `NeedsDefaultConstraint` → `HasAnnotatedNullableUsage` across model, discovery, and builder. The model now describes what was detected; the builder decides whether to emit `where T : default` based on context.

**Item 2 (Properties):** No action needed — C# properties cannot have their own type parameters, so `TypeParameters` is always empty on property models.

**Item 3 (Tuples):** No action needed — Roslyn represents tuples as `INamedTypeSymbol` (with `IsTupleType == true`). There is no `ITupleTypeSymbol` interface. The existing `INamedTypeSymbol` branch already recurses into `TypeArguments`, so `(T?, string)` is handled.

**Item 4 (Bool → enum):** Skipping — single boolean with one call-site distinction doesn't warrant an enum yet.

**Item 5 (Early-exit):** Acknowledged as benign. The `IsUnconstrained()` early-exit handles the empty case, making the final `constraints.Count > 0` ternary unnecessary.

thomhurst:
Re: tuple elements — `ITupleTypeSymbol` does not exist in the Roslyn API. C# tuples are represented as `INamedTypeSymbol` (specifically `System.ValueTuple<...>`), so the existing `INamedTypeSymbol` branch already recurses into their `TypeArguments`.

Added a `(T?, string) GetPair<T>()` case to the snapshot test to prove it. The generated output correctly emits `where T : default`:

```csharp
(T?, string) global::IFoo.GetPair<T>() where T : default => Object.GetPair<T>();
```

No additional handling needed.

Issue #5394, opened 2026-04-04
perf: split filter data into separate lightweight class to avoid full .cctor during discovery


thomhurst:
## Summary

Follow-up to the lazy registration optimization (which defers per-class `.cctor` JIT from module init to discovery time).

Currently, when the engine needs to filter tests by method name/category, it must resolve the lazy factory for each matching class — which triggers the full `.cctor` (creating `ClassMetadata`, `MethodMetadata`, delegate captures, `TestEntry[]` arrays).

The idea is to **emit filter data in a separate nested class** with its own lightweight `.cctor`, so that discovery filtering only JITs the filter data — not the full test entries.

## Proposed approach

Split the generated `__TestSource` class into two parts:

```csharp
internal static class TUnit_TestProject_BasicTests__TestSource
{
    // Separate class = separate .cctor (just strings and booleans, no delegates)
    internal static class Filter
    {
        public static readonly int Count = 3;
        public static readonly TestEntryFilterData[] Data = new[]
        {
            new TestEntryFilterData { MethodName = "SynchronousTest", ClassName = "BasicTests", ... },
            new TestEntryFilterData { MethodName = "AsynchronousTest", ClassName = "BasicTests", ... },
            new TestEntryFilterData { MethodName = "ValueTaskAsynchronousTest", ClassName = "BasicTests", ... },
        };
    }
    
    // Full entries — only accessed when tests actually need to execute
    public static readonly TestEntry<BasicTests>[] Entries = /* ... existing ... */;
}
```

Registration becomes:
```csharp
SourceRegistrar.RegisterLazy<BasicTests>(
    TUnit_TestProject_BasicTests__TestSource.Filter.Data,     // ← cheap, separate .cctor
    static () => TUnit_TestProject_BasicTests__TestSource.Entries);  // ← deferred until execution
```

## Benefit

During discovery:
- **Type filter** — zero materialization (dict key check only)
- **Method/category filter** — only `Filter..cctor` JIT'd (just string arrays, ~microseconds)
- **Execution** — full `__TestSource..cctor` JIT'd (delegates, metadata) only for tests that will run

## Context

- Related runtime issue: https://github.com/dotnet/runtime/issues/126541
- jkotas's layered registration suggestion: https://github.com/dotnet/runtime/issues/126541#issuecomment-4187803850
- Phase 1 (lazy factory registration) is already implemented

PR #5395, opened 2026-04-04
perf: defer per-class JIT via lazy test registration + parallel resolution

thomhurst:
## Summary

- **Lazy registration:** Source generator now emits `RegisterLazy<T>(static () => Source.Entries)` instead of `RegisterEntries(Source.Entries)`. The lambda captures a function pointer without triggering the target class's static constructor — deferring all per-class JIT from module init to discovery time.
- **Parallel resolution:** Engine resolves lazy sources in parallel via `Parallel.ForEach` before filtering, turning sequential per-class JIT into parallel JIT on multi-core machines.
- **Filter-aware:** For filtered runs (common during development), only type-matched classes are resolved at all — unmatched classes never JIT their `.cctor`.

### Startup JIT cost (1000 test classes)

| Phase | Before (Eager) | After (Lazy) |
|-------|----------------|--------------|
| Module init | JIT 1000 `.cctor`s sequentially | JIT 1 `.cctor` + 1000 lambda allocs |
| Discovery (filtered to 1 class) | Already done | JIT 1 `.cctor` |
| Discovery (unfiltered) | Already done | JIT 1000 `.cctor`s **in parallel** |

At module init, only **4 methods** are JIT'd regardless of N (the shared `.cctor`, `RegisterLazy<__Canon>`, `LazyTestEntrySource<__Canon>..ctor`, `Func<__Canon>..ctor`) — O(1) due to .NET's reference-type generic code sharing.

### No impact on incremental generation

Each test class still generates its own file with its own partial contribution to `TUnit_TestRegistration`. No `Collect` needed. The only change per file is the registration expression.

### Files changed

| File | Change |
|------|--------|
| `TUnit.Core/LazyTestEntrySource.cs` | New — lazy wrapper with factory list support for generic merging |
| `TUnit.Core/SourceRegistrar.cs` | New `RegisterLazy<T>()` overload |
| `TUnit.Core.SourceGenerator/.../TestMetadataGenerator.cs` | Both registration call sites emit `RegisterLazy` |
| `TUnit.Engine/.../AotTestDataCollector.cs` | Parallel pre-resolution before filtering |
| 122 `.verified.txt` files | Updated registration lines in snapshots |

Context: https://github.com/dotnet/runtime/issues/126541
Follow-up: #5394

## Test plan

- [x] Snapshot tests pass on all 4 TFMs (net8.0, net9.0, net10.0, net472)
- [x] End-to-end: BasicTests, DataDrivenTests, GenericMethodTests, PropertySetterTests all pass
- [x] Build succeeds on netstandard2.0 and net10.0
- [ ] CI passes

### DIFF
--- a/TUnit.Core.SourceGenerator/Generators/TestMetadataGenerator.cs
+++ b/TUnit.Core.SourceGenerator/Generators/TestMetadataGenerator.cs
@@ -505,7 +505,6 @@ public sealed class TestMetadataGenerator : IIncrementalGenerator
         writer.Unindent();
         writer.AppendLine("}");
 
-        // Registration calls — factory wrapper defers per-class JIT until discovery
         registrationIndex = 0;
         foreach (var group in groupedByConcreteClass)
         {
@@ -3555,7 +3554,6 @@ public sealed class TestMetadataGenerator : IIncrementalGenerator
             writer.Unindent();
             writer.AppendLine("}");
 
-            // Registration — factory wrapper defers per-class JIT until discovery
             EmitRegistrationField(writer, classGroup.TestSourceName,
                 $"global::TUnit.Core.SourceRegistrar.RegisterEntries<{classGroup.ClassFullyQualified}>(static () => {classGroup.TestSourceName}.Entries)");
 
--- a/TUnit.Core/TestEntrySource.cs
+++ b/TUnit.Core/TestEntrySource.cs
@@ -39,10 +39,7 @@ public sealed class TestEntrySource<
             {
                 // Already resolved — merge eagerly
                 var additional = factory();
-                var combined = new TestEntry<T>[_entries.Length + additional.Length];
-                Array.Copy(_entries, 0, combined, 0, _entries.Length);
-                Array.Copy(additional, 0, combined, _entries.Length, additional.Length);
-                _entries = combined;
+                _entries = [.. _entries, .. additional];
                 return;
             }
 


thomhurst:
Thanks for the thorough review! Here's the status of each item:

**Critical: `Sum()` defeating laziness** — Fixed in a500ef504. The `Sum()` has been removed; `matching` now uses a growing list.

**`RegisterEntries` kept with no callers** — Addressed in c55d70012. The refactor consolidated `RegisterEntries` + `RegisterLazy` into a single `RegisterEntries<T>(Func<TestEntry<T>[]>)`. There is no longer a separate `RegisterLazy` method or `LazyTestEntrySource<T>` type. `TestEntrySource<T>` itself is now natively lazy.

**`AggregateException` from `Parallel.ForEach`** — The engine already unwraps `AggregateException` in `TUnitMessageBus` and `FailureCategorizer`. No action needed.

**Validation timing** — Intentional tradeoff. Empty-entries validation now happens at resolution time rather than module init. This only catches source-gen bugs, which are caught the moment the engine discovers the affected class.

**`ClassName` benign race** — Intentional. `GetNestedTypeName(typeof(T))` is pure/idempotent, string writes are atomic. No correctness risk.

**Bundled assertion API changes** — The `Tests.Assertions_Library_Has_No_API_Changes.Net4_7.verified.txt` change is pre-existing snapshot staleness on net472, not introduced by this PR. It just happened to get picked up when regenerating snapshots.

**Reflection path** — Not affected. Reflection-based discovery uses `Assembly.GetTypes()` / `MethodInfo` scanning, not pre-registered factories. The JIT cold-start issue is specific to the source-gen path where `[ModuleInitializer]` eagerly triggers all class static constructors.

**`ParallelOptions`** — Default `Parallel.ForEach` uses `Environment.ProcessorCount` which is appropriate for CPU-bound JIT work. Adding explicit options would be over-specification.

thomhurst:
Thanks for catching the Phase 1 issue! Fixed in 16d3e3454.

The `typeMatches` variable was computed before the loop but `source.Count` was still called unconditionally inside it, triggering `Resolve()` for non-matching types. Now non-type-matched sources are skipped via early `continue` before `source.Count` is ever accessed.

**Re: `ParallelOptions`** — Agreed this is low priority. The default `Environment.ProcessorCount` is appropriate for CPU-bound JIT work, and cancellation during discovery isn't a supported scenario.

PR #5412, opened 2026-04-05
fix: prevent StringBuilder race in console interceptor during parallel tests

### LINKED ISSUE #5411: Flaky ArgumentOutOfRangeException in StringBuilder.ToString() during parallel test execution
## Description

Intermittent `ArgumentOutOfRangeException` with parameter `chunkLength` at `System.Text.StringBuilder.ToString()` during parallel test execution with code coverage enabled. The exception is wrapped by TUnit as `[Test Failure]` and causes random test failures.

## Stack Trace

```
TUnit.Engine.Exceptions.TestFailedException: [Test Failure] ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than or equal to the size of the collection. (Parameter 'chunkLength')
    at System.Text.StringBuilder.ToString()
```

The stack trace is notably short — no TUnit frame, no application frame. Just `StringBuilder.ToString()`.

## Environment

- TUnit v1.22.19
- .NET 10.0.201
- Linux (GitHub Actions runner, Ubuntu 24.04)
- Code coverage enabled (`--coverage`)
- Parallel test execution with ASP.NET Core `WebApplicationFactory` integration tests

The test passes locally and passes most CI runs. Failure rate is low but reproducible over multiple runs.

## Root Cause Analysis

We believe the issue is in `OptimizedConsoleInterceptor` (line 168 of `TUnit.Engine/Logging/OptimizedConsoleInterceptor.cs`):

```csharp
public override void Write(StringBuilder? value) => Write(value?.ToString() ?? string.Empty);
```

And similarly at lines 171-172:

```csharp
public override Task WriteAsync(StringBuilder? value, CancellationToken cancellationToken = new())
    => WriteAsync(value?.ToString() ?? string.Empty);
```

### The race condition

ASP.NET Core's `ConsoleLogger` builds log messages in a pooled `StringBuilder` and writes them to `Console.Out` via the `TextWriter.Write(StringBuilder)` overload (available in .NET 8+). This avoids an intermediate string allocation by passing the `StringBuilder` directly.

TUnit's `OptimizedConsoleInterceptor` intercepts `Console.Out`. When it receives a `StringBuilder`, it calls `.ToString()` on the **caller's** `StringBuilder`. The caller (ASP.NET Core's logger) may return that `StringBuilder` to a pool immediately after the `Write` call returns. Under parallel test execution, another thread can pick up that pooled `StringBuilder` and start appending to it while `.ToString()` is still traversing the internal chunk linked list.

The sequence:

1. ASP.NET Core console logger builds a message in a pooled `StringBuilder`
2. Calls `Console.Out.Write(stringBuilder)` — intercepted by TUnit
3. `OptimizedConsoleInterceptor.Write(StringBuilder)` calls `stringBuilder.ToString()`
4. The caller returns the `StringBuilder` to the pool (the `Write` contract doesn't require the caller to keep the `StringBuilder` alive after the call)
5. Another thread picks up the pooled `StringBuilder` and starts appending
6. `.ToString()` sees inconsistent internal chunk state → `ArgumentOutOfRangeException` with parameter `chunkLength`

This only manifests under high concurrency (many parallel tests logging simultaneously) and is amplified by code coverage instrumentation, which changes thread scheduling.

### Related

This is similar in nature to #2670 (fixed by #2708), which added locking around the `ConcurrentStringWriter`. The difference is that the `StringBuilder` in this case is not TUnit's own — it's passed in by the caller.

## Suggested Fix

The `Write(StringBuilder)` and `WriteAsync(StringBuilder)` overloads should not call `.ToString()` on the caller's `StringBuilder` because it may be pooled and reused concurrently. The safest approach is to copy the content into a local string before the method returns, ensuring the caller's `StringBuilder` is not accessed after the synchronous call completes.

A secondary issue exists in `Context.GetStandardOutput()` and `Context.GetErrorOutput()` where `_outputBuilder.Length` is read outside the `ReaderWriterLockSlim`. While this is unlikely to be the primary cause, moving the `.Length` check inside the lock would be a small defensive improvement.

JSkimming:
Fixes #5411

## Problem

`OptimizedConsoleInterceptor.Write(StringBuilder)` calls `.ToString()` on the caller's `StringBuilder`. ASP.NET Core's `ConsoleLogger` pools its `StringBuilder` and may return it to the pool immediately after `Write` returns. Under parallel test execution with code coverage, another thread can pick up the pooled `StringBuilder` and mutate it while `.ToString()` is still traversing the internal chunk list, causing:

```
ArgumentOutOfRangeException: Index was out of range. (Parameter 'chunkLength')
    at System.Text.StringBuilder.ToString()
```

## Fix

**`OptimizedConsoleInterceptor.cs`**: Replace `value?.ToString()` with `CopyStringBuilderSafely(value)` for all `StringBuilder`-accepting overloads (`Write`, `WriteAsync`, `WriteLine`, `WriteLineAsync`). The helper copies via `string.Create` + `StringBuilder.CopyTo` and catches `ArgumentOutOfRangeException` if the race hits — losing one log entry rather than crashing the test.

**`Context.cs`**: Move the `_outputBuilder.Length` early-return check inside the `ReaderWriterLockSlim` in `GetStandardOutput()` and `GetErrorOutput()`. `StringBuilder.Length` is not thread-safe and was being read outside the lock.

## Test plan

- Existing `ContextThreadSafetyTests` covers concurrent read/write on the output builder
- The flaky failure is inherently difficult to reproduce locally — it requires high concurrency + code coverage instrumentation on Linux CI

### DIFF
--- a/TUnit.Core/Context.cs
+++ b/TUnit.Core/Context.cs
@@ -82,16 +82,13 @@ public abstract class Context : IContext, IDisposable
 
     public virtual string GetStandardOutput()
     {
-        if (_outputBuilder.Length == 0)
-        {
-            return string.Empty;
-        }
-
         _outputLock.EnterReadLock();
 
         try
         {
-            return _outputBuilder.ToString();
+            return _outputBuilder.Length == 0
+                ? string.Empty
+                : _outputBuilder.ToString();
         }
         finally
         {
@@ -101,16 +98,13 @@ public abstract class Context : IContext, IDisposable
 
     public virtual string GetErrorOutput()
     {
-        if (_errorOutputBuilder.Length == 0)
-        {
-            return string.Empty;
-        }
-
         _errorOutputLock.EnterReadLock();
 
         try
         {
-            return _errorOutputBuilder.ToString();
+            return _errorOutputBuilder.Length == 0
+                ? string.Empty
+                : _errorOutputBuilder.ToString();
         }
         finally
         {
--- a/TUnit.Engine/Logging/OptimizedConsoleInterceptor.cs
+++ b/TUnit.Engine/Logging/OptimizedConsoleInterceptor.cs
@@ -165,17 +165,50 @@ internal abstract class OptimizedConsoleInterceptor : TextWriter
 
 #if NET
     public override void Write(ReadOnlySpan<char> buffer) => Write(new string(buffer));
-    public override void Write(StringBuilder? value) => Write(value?.ToString() ?? string.Empty);
+    public override void Write(StringBuilder? value) => Write(CopyStringBuilderSafely(value));
     public override Task WriteAsync(ReadOnlyMemory<char> buffer, CancellationToken cancellationToken = new())
         => WriteAsync(new string(buffer.Span));
     public override Task WriteAsync(StringBuilder? value, CancellationToken cancellationToken = new())
-        => WriteAsync(value?.ToString() ?? string.Empty);
+        => WriteAsync(CopyStringBuilderSafely(value));
     public override void WriteLine(ReadOnlySpan<char> buffer) => WriteLine(new string(buffer));
-    public override void WriteLine(StringBuilder? value) => WriteLine(value?.ToString() ?? string.Empty);
+    public override void WriteLine(StringBuilder? value) => WriteLine(CopyStringBuilderSafely(value));
     public override Task WriteLineAsync(ReadOnlyMemory<char> buffer, CancellationToken cancellationToken = new())
         => WriteLineAsync(new string(buffer.Span));
     public override Task WriteLineAsync(StringBuilder? value, CancellationToken cancellationToken = new())
-        => WriteLineAsync(value?.ToString() ?? string.Empty);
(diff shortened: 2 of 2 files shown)


JSkimming:
OK, so this is the real me, not Claude on my behalf. I'm somewhat sceptical about ASP.NET mutating the string builder before the call completes, but with the stack tracing being so sparse, there's not much to go on. The change seems at least neutral. The locking changes are certainly worthwhile, as Length shouldn't be read outside of a lock, though it isn't a mutating operation and shouldn't cause the issue.

Happy to take feedback and make adjustments.

thomhurst:
If you can reproduce it again, run with `--detailed-stacktrace` to hopefully give you more info

JSkimming:
Superseded by #5414 — the owner took a cleaner approach: removing the StringBuilder overrides entirely rather than catching the exception. The base TextWriter handles StringBuilder chunk-by-chunk via GetChunks(), avoiding the race without any ToString() call.

PR #5447, opened 2026-04-07
perf: split filter data into separate nested class (#5394)

### LINKED ISSUE #5394: perf: split filter data into separate lightweight class to avoid full .cctor during discovery

thomhurst:
## Summary

Closes #5394. Eliminates per-class delegate/metadata `.cctor` JIT from the discovery hot path.

### The change

- **Split filter data into a sibling nested class.** The source generator now emits `Filter` (or `Filter_N` for grouped registrations) as a nested `static class` alongside the existing `Entries` field. `Filter.Data` holds only literal strings/ints, so its `.cctor` is trivially cheap. The heavy `__TestSource..cctor` (delegates + `MethodMetadata` graph) only runs for entries actually selected for materialization.
- **Lazy `Func<TestEntryFilterData[]>` factory.** `SourceRegistrar.RegisterEntries<T>` now takes a `Func<TestEntryFilterData[]>` instead of the array itself. `TestEntrySource<T>` resolves lazily via double-checked lock. Narrow-type filters that short-circuit at `CouldTypeMatch` never trigger the `Filter` cctor at all.
- **Parallelized `AotTestDataCollector`.**
  - **Phase 1** (no-hint path): `Parallel.For` over sources so each class's `Filter.Data` cctor JIT-compiles on a worker thread. When type/method hints are present, stays serial because short-circuiting dominates.
  - **Phase 3** (materialization): `Parallel.For` so each class's heavy `Entries` cctor JIT runs on a worker thread instead of serially on the caller.
- **API cleanup.** Replaced the initial `Count`/`GetFilterData(int)` pair with a single `FilterData` array property to eliminate a TOCTOU race on concurrent registration. `ClassName` moved off `TestEntryFilterData` (derivable from `typeof(T)`).

### Perf (531 synthetic classes, 11,952 tests, net9.0, measured via internal discovery duration)

| Scenario | Baseline | PR | Delta |
|---|---|---|---|
| Narrow-type filter | 687 ms | 694 ms | ~tie |
| Broad method filter | 698 ms | 712 ms | ~tie |
| `--list-tests` (all) | 703 ms | 684 ms | **−19 ms** |

The `--list-tests` worst case is the primary target: every class's metadata has to be materialized, so distributing the cctor JIT across cores is pure win. Filtered scenarios are neutral because they already skip most cctors — the lazy factory prevents regressing them.

### Review feedback addressed

- Volatile `_filterData` field + double-checked locking.
- TOCTOU on `Count`/`GetFilterData` → single `FilterData` array property.
- `Interlocked.Exchange` → `CompareExchange` on the `hasDependencies` flag.
- Documented that `AddSource` is safe because CLR serialises module `.cctor`s.
- Documented on `ITestEntrySource.FilterData` that the array is engine-internal and must not be mutated (kept as a bare array for hot-path dispatch cost).

## Test plan

- [x] Source generator snapshot tests (`TUnit.Core.SourceGenerator.Tests`, 116 passing)
- [x] `TUnit.PublicAPI` snapshot tests across net4.7, net8, net9, net10
- [x] Benchmark regression check on 3 scenarios (narrow-type, broad-method, `--list-tests`)

### DIFF
--- a/TUnit.Core/ITestEntrySource.cs
+++ b/TUnit.Core/ITestEntrySource.cs
@@ -22,6 +22,11 @@ public interface ITestEntrySource
     /// per-class delegate/metadata <c>.cctor</c>. Callers should snapshot this reference
     /// once and loop over it, rather than calling repeatedly, to avoid any torn reads
     /// if multiple sources are registered concurrently.
+    /// <para>
+    /// The returned array is internal engine state — callers <b>must not</b> mutate it.
+    /// It is exposed as a bare array (not <see cref="IReadOnlyList{T}"/>) purely to avoid
+    /// interface-dispatch overhead on the discovery hot path.
+    /// </para>
     /// </summary>
     TestEntryFilterData[] FilterData { get; }
 
--- a/TUnit.Core/TestEntrySource.cs
+++ b/TUnit.Core/TestEntrySource.cs
@@ -35,6 +35,9 @@ public sealed class TestEntrySource<
     /// Adds another (filterDataFactory, factory) pair for the same T. Used when multiple source-gen
     /// files register entries for the same class (e.g. generic instantiations).
     /// Thread-safe via lock since static field initializers may run concurrently.
+    /// Note: all calls to <c>AddSource</c> happen from module/type <c>.cctor</c>s, which the CLR
+    /// serialises before any discovery code runs. So there is no window where a reader could observe
+    /// a merged <c>_filterData</c> array but a not-yet-merged <c>_factories</c> list (or vice versa).
     /// </summary>
     internal void AddSource(Func<TestEntryFilterData[]> filterDataFactory, Func<TestEntry<T>[]> factory)
     {
--- a/TUnit.Engine/Building/Collectors/AotTestDataCollector.cs
+++ b/TUnit.Engine/Building/Collectors/AotTestDataCollector.cs
@@ -131,7 +131,9 @@ internal sealed class AotTestDataCollector : ITestDataCollector
                 perClassMatches[i] = local;
                 if (localHasDeps)
                 {
-                    Interlocked.Exchange(ref hasDependenciesFlag, 1);
+                    // CompareExchange so the write is a no-op once the flag is already set,
+                    // avoiding a redundant LOCK bus round-trip on hot paths.
+                    Interlocked.CompareExchange(ref hasDependenciesFlag, 1, 0);
                 }
             });
 


thomhurst:
Closing: after multiple rounds of benchmarking on 531 synthetic classes / ~12k tests, the measured impact across narrow-type, broad-method, and --list-tests scenarios was within noise (±2%). The theoretical win — avoiding heavy per-class .cctor JIT during discovery filtering — didn't translate into a clear wallclock improvement, and the Filter/Entries split added non-trivial codegen and public-API complexity for no measurable gain.

Leaving #5394 open. If the cctor JIT cost turns out to matter at much larger scale (5k+ classes) or on a real suite, a narrower change (just parallelizing AotTestDataCollector.CollectTestsFromTestEntries) would be the first thing to try.

PR #5494, opened 2026-04-10
feat(reporters): add minimap sidebar navigator to HTML report

### LINKED ISSUE #5488: HTML Report: Add minimap sidebar navigator for large test suites
## Summary

For test suites with hundreds of test classes, finding a specific group requires either using the search box (if you know the name) or scrolling through a long list. A collapsible sidebar minimap showing all groups with colored pass/fail indicators would let users visually scan the entire suite at a glance and jump directly to areas of interest.

### Sidebar panel
- A narrow, collapsible sidebar (or overlay panel) anchored to the left or right edge of the viewport.
- Toggled via a button in the toolbar (e.g. a minimap/list icon next to the expand/collapse buttons).
- Each test group is represented as a compact row: a colored dot (green/red/amber) indicating worst status in the group, plus a truncated class name.
- Clicking a row scrolls the main content to that group and expands it.

### Visual density
- The minimap should be compact enough to show 50+ groups without scrolling in most viewports.
- Use small font size and tight line height (similar to a code editor minimap or a table of contents).
- The currently visible group in the main viewport should be highlighted in the sidebar (scroll-spy behavior via IntersectionObserver).

### Responsiveness
- On narrow viewports (< 768px), the sidebar should be an overlay/drawer rather than taking up layout space.
- On wide viewports, it can sit alongside the main content within the existing `max-width: 1360px` shell, or extend beyond it.

### State
- The sidebar's open/closed state should persist in `localStorage` (like the theme preference).
- Hidden by default — only users who need it will toggle it on. Avoids adding visual noise for small suites.
- Only render the toggle button if the suite has more than ~10 groups (no point for small reports).

## Why This Matters

In CI for large projects with 200+ test classes, the report's main scroll area becomes very long. Users investigating failures often know roughly which area of the codebase broke (e.g. "something in the Auth namespace") but don't know the exact class name. The minimap provides:
- A birds-eye view of suite health — immediately spot clusters of red dots
- One-click navigation without typing
- Spatial memory — "the failing group was near the bottom" becomes actionable

## Implementation Notes

- Group data is already available in the JS `computeDisplayGroups()` function.
- Scroll-spy can reuse the existing `IntersectionObserver` pattern (used for lazy loading and the sticky header).
- The sidebar HTML can be generated in `render()` alongside the main content, or as a separate function called once after initial render.
- Consider using `position: sticky` for the sidebar so it stays visible while the main content scrolls.
- The minimap should respect the current filter/search/group state — only show groups that have visible tests.

thomhurst:
## Summary

- Adds a collapsible minimap sidebar to the HTML test report for navigating large test suites (200+ test classes)
- Shows colored status bars (green/red/amber/gradient) proportional to test count per group, with IntersectionObserver scroll-spy highlighting the currently visible group
- Hidden by default, toggle button only appears when >10 groups exist; state persists via localStorage; keyboard shortcut `m` to toggle; responsive with backdrop overlay on mobile

## Details

- **CSS**: Fixed sidebar with slide-in transition, 5px colored bars for density (~88 rows per 800px viewport), responsive breakpoints, print/high-contrast/forced-colors support
- **JS**: `updateMinimap()` builds bar-based minimap from all display groups (not just lazy-loaded subset), `countStatuses()` extracted as shared helper, IntersectionObserver for scroll-spy with `rootMargin: '0px 0px -75% 0px'`
- **Navigation**: Click-to-scroll with auto-expand of target group; if target is beyond lazy-load limit, extends `renderLimit` and re-renders first

Closes #5488

## Test plan

- [ ] Open HTML report with <10 test classes — minimap toggle button should be hidden
- [ ] Open HTML report with >10 test classes — minimap toggle button visible in toolbar
- [ ] Click toggle or press `m` — sidebar slides in from right with colored bars
- [ ] Scroll main content — active group highlighted in minimap via scroll-spy
- [ ] Click a minimap bar — scrolls to that group and expands it
- [ ] Click a bar for a group beyond the lazy-load limit — group loads and scrolls into view
- [ ] Toggle minimap closed, refresh page — minimap stays closed (localStorage)
- [ ] Toggle minimap open, refresh page — minimap reopens (localStorage)
- [ ] Resize to mobile width — backdrop overlay appears behind minimap
- [ ] Filter by status/search — minimap updates to show only matching groups
- [ ] Press `?` — help overlay shows minimap shortcut

### DIFF
--- a/TUnit.Engine/Reporters/Html/HtmlReportGenerator.cs
+++ b/TUnit.Engine/Reporters/Html/HtmlReportGenerator.cs
@@ -85,6 +85,14 @@ internal static partial class HtmlReportGenerator
         AppendJavaScript(sb);
 
         sb.AppendLine("</div>");
+
+        // Feature 11: Minimap sidebar navigator
+        sb.AppendLine("<div class=\"minimap-backdrop\" id=\"minimapBackdrop\"></div>");
+        sb.AppendLine("<div class=\"minimap\" id=\"minimap\" role=\"navigation\" aria-label=\"Test group navigator\">");
+        sb.AppendLine("<div class=\"minimap-hd\">Navigator</div>");
+        sb.AppendLine("<div id=\"minimapContent\"></div>");
+        sb.AppendLine("</div>");
+
         sb.AppendLine("</body>");
     }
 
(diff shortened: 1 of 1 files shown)


thomhurst:
## Review response

Addressed items 1, 4, 5 in 7b1b9c53a. Pushing back on 2 and 3:

### 1. Minimap rows not keyboard accessible — Fixed
Added `role="button" tabindex="0"` to all minimap rows and a `keydown` handler for Enter/Space that delegates to the same `navigateToRow()` function as click.

### 2. Rebuilding minimap on every render() — No change needed
The search input is already debounced, so `render()` doesn't fire on every keystroke. When it does fire, the main `container.innerHTML` rebuild (hundreds of test rows with detail panels) dwarfs the minimap rebuild (~200 tiny divs). The IntersectionObserver must be recreated because `container.innerHTML` destroys all observed elements. Separating concerns would add complexity without measurable benefit.

### 3. Multiple simultaneous active highlights — Intentional
Multiple active rows show the "viewport window" — which section of the report is currently visible. This is the same behavior as VS Code's minimap highlighting a range, not a single line. Collapsing to a single row would lose that spatial context.

### 4. parseInt without radix — Fixed
Added radix 10.

### 5. var → const/let — Fixed
All minimap code now uses `const`/`let` consistently with the rest of the file.

PR #5497, opened 2026-04-10
feat(html-report): omit redundant test body span from trace timeline

thomhurst:
## Summary
- When a test has no before/after hooks, the "test body" span is the only child of the test case span and adds no useful information to the trace timeline — omit it
- If the test body has child spans (e.g. HTTP calls, DB queries), those are re-parented directly under the test case span so they remain visible
- Add `SpanTestBody` constant to `TUnitActivitySource` alongside the other span name constants, eliminating the magic string in `TestExecutor`

## Test plan
- [ ] Run a test with no hooks — verify the trace timeline is hidden (no redundant "test body" bar)
- [ ] Run a test with before/after hooks — verify all spans (hooks + test body) still appear
- [ ] Run a test whose body produces child spans (e.g. HttpClient activity) — verify child spans appear directly under the test case span

### DIFF
--- a/TUnit.Core/TUnitActivitySource.cs
+++ b/TUnit.Core/TUnitActivitySource.cs
@@ -16,6 +16,7 @@ internal static class TUnitActivitySource
     internal const string SpanTestAssembly = "test assembly";
     internal const string SpanTestSuite = "test suite";
     internal const string SpanTestCase = "test case";
+    internal const string SpanTestBody = "test body";
 
     // Tag keys used across init/dispose spans and the HTML report.
     internal const string TagTestId = "tunit.test.id";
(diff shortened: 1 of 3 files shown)


thomhurst:
Addressed the review findings:

**1. Hardcoded string coupling** — The JS is a raw string literal (`"""`), so C# interpolation would require escaping hundreds of `{`/`}` across ~900 lines of JS — not practical. Added a comment linking the JS literal to `TUnitActivitySource.SpanTestBody` so the coupling is documented. (The same coupling already exists for `'test case'` on line 1566.)

**2. `<= 1` threshold** — This is actually fine as-is. `getDescendants` includes the root span itself, so after removing test body from a test with 1 child span: `sp = [testCase, httpChild]` → length 2 → the trace IS shown. The `<= 1` only hides the timeline when only the bare test case span remains with no children at all.

**3. `Object.assign` vs spread** — Fixed to `{...s, parentSpanId: rootSpanId}` for consistency with the rest of the JS.

PR #5519, opened 2026-04-12
Expose test arguments as filterable properties

SimonCropp:
Parameterized tests can now be filtered by their resolved argument
  values via `--treenode-filter`, e.g.:

      --treenode-filter "/*/*/MyTests/*[arg0=*foo*]"

  TestBuilder.InvokeDiscoveryEventReceiversAsync stringifies each entry
  in TestDetails.TestMethodArguments and adds it to the DiscoveredTestContext
  as arg0, arg1, ... so it flows into the existing property bag used by
  the filter matcher. Wildcards in the value are supported by the platform's
  built-in matcher.


thomhurst:
Just to avoid redundant work for those who don't need this, and also to not interfere with any possibilities of users having explicitly set these, can we wrap them in a feature flag that the user can control? I added settings functionality in #5522 

I'm thinking maybe something like:

```csharp
[Before(HookType.TestDiscovery)]
public static void Configure(BeforeTestDiscoveryContext context)
{
    context.Settings.Metadata.EnableArgumentProperties = true;
}
```

thomhurst:
Also should this work with class constructors too?

SimonCropp:
> Also should this work with class constructors too

for my purposes, i dont need it

SimonCropp:
@thomhurst  want me to reboot this?

PR #5532, opened 2026-04-13
refactor: move CorrelatedTUnitLogger to TUnit.Logging.Microsoft and auto-inject handlers

thomhurst:
## Summary

- **Move `CorrelatedTUnitLogger` and `CorrelatedTUnitLoggerProvider`** from `TUnit.AspNetCore.Core` to `TUnit.Logging.Microsoft`, removing the ASP.NET dependency so correlated logging works purely via Activity baggage and `TestContext.ResolveFromActivityBaggage()`
- **Auto-inject `ActivityPropagationHandler` and `TUnitTestIdHandler`** into all `HttpClient` instances created by `TestWebApplicationFactory` via `new` method shadows on `CreateDefaultClient`/`CreateClient`, obsoleting `CreateClientWithTestContext`
- **Eagerly capture `TestContext`** in `TUnitTestIdHandler` at construction time (with dynamic fallback), and deduplicate handler-prepend logic in `TracedWebApplicationFactory`
- **Add 7 comprehensive tests** covering Activity baggage resolution, correlated logger behavior, duplicate suppression, log level filtering, and error routing

## Test plan
- [x] All 13 ASP.NET Core tests pass (`dotnet run --framework net10.0` in `TUnit.AspNetCore.Tests`)
- [x] `TUnit.AspNetCore.Core` and `TUnit.Logging.Microsoft` build with 0 errors, 0 warnings
- [ ] CI passes across net8.0/net9.0/net10.0

### DIFF
--- a/TUnit.AspNetCore.Core/Extensions/WebApplicationFactoryExtensions.cs
+++ b/TUnit.AspNetCore.Core/Extensions/WebApplicationFactoryExtensions.cs
@@ -11,12 +11,11 @@ public static class WebApplicationFactoryExtensions
     /// <summary>
     /// Creates an <see cref="HttpClient"/> with a <see cref="TUnitTestIdHandler"/> that automatically
     /// propagates the current test context ID to the server via HTTP headers.
-    /// Use with <see cref="Logging.CorrelatedTUnitLoggingExtensions.AddCorrelatedTUnitLogging"/>
-    /// on the server side to enable the test context middleware.
     /// </summary>
     /// <typeparam name="TEntryPoint">The entry point class of the web application.</typeparam>
     /// <param name="factory">The web application factory.</param>
     /// <returns>An <see cref="HttpClient"/> configured with test context propagation.</returns>
+    [Obsolete("TestWebApplicationFactory now injects ActivityPropagationHandler and TUnitTestIdHandler automatically. Use CreateClient() or CreateDefaultClient() instead.")]
     public static HttpClient CreateClientWithTestContext<TEntryPoint>(
         this WebApplicationFactory<TEntryPoint> factory)
         where TEntryPoint : class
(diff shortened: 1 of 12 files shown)


thomhurst (commented):
Thanks for the detailed review. A few responses:

### `new` vs `override` — not applicable

`CreateClient()`, `CreateClient(WebApplicationFactoryClientOptions)`, `CreateDefaultClient(params DelegatingHandler[])`, and `CreateDefaultClient(Uri, params DelegatingHandler[])` are all **non-virtual** on `WebApplicationFactory<TEntryPoint>` (verified against `Microsoft.AspNetCore.Mvc.Testing` v10.0.5 in the NuGet cache). Using `override` produces CS0506. Method hiding with `new` is the only option.

This is also why `TracedWebApplicationFactory` was designed as a composition wrapper rather than inheriting from `WebApplicationFactory` — it avoids the non-virtual method limitation entirely.

In TUnit's test patterns, factories are always typed as the concrete `TestWebApplicationFactory<T>` subclass (via `[ClassDataSource]`), so the `new` methods are always resolved correctly.

### `CreateClient()` and `ClientOptions`

Since `CreateClient(WebApplicationFactoryClientOptions)` is also non-virtual, there's no way to intercept it. Our `CreateClient()` calls `CreateDefaultClient()` + `ConfigureClient(client)` which handles `BaseAddress`. The default `ClientOptions` behavior (auto-redirect, cookies) is handled by the base `HttpClientHandler` defaults, so the common case works. Users needing custom options can call `CreateDefaultClient(customHandlers)` directly.

### Handler deduplication

`TracedWebApplicationFactory` wraps a generic `WebApplicationFactory<T>` (not a `TestWebApplicationFactory<T>`), so it can't delegate to the base class's shadow methods. The 4-line array-prepend pattern is simple enough that a shared helper would add indirection without meaningful benefit.

### Test positive case

The positive case (correlated logger writes output when no per-test logger is active) is already covered by `CorrelatedLogger_WritesToCorrectTest_ViaActivityBaggage` and `CorrelatedLogger_RoutesErrorToStdErr` in the same test class, plus the HTTP integration tests in `CorrelatedLoggingResolverTests`.

PR #5572, opened 2026-04-16
refactor: separate test and lifecycle tracing

thomhurst:
## Summary
- keep exported backend traces as simple per-test parent-child trees for request navigation and log correlation
- auto-export the runner's `TUnit` test spans from `TUnit.Aspire` to the same OTLP backend used by the Aspire dashboard when `DOTNET_DASHBOARD_OTLP_ENDPOINT_URL` is present
- normalize the runner exporter endpoint to `.../v1/traces` so path-based OTLP backends like Seq also work correctly
- add a richer Aspire trace demo request that emits internal spans and nested HTTP calls for backend validation
- update the Aspire docs to describe the out-of-box backend trace behavior

## Out-of-Box Backend Validation
Local Jaeger validation on April 16, 2026 using only `DOTNET_DASHBOARD_OTLP_ENDPOINT_URL=http://127.0.0.1:4318` in the test runner and a real `OtlpCorrelationIntegrationTests/TraceDemo_Request_ProducesNestedCorrelatedLogs` run. No manual tracer provider was added to the test project.

Jaeger search/timeline view showing one trace with 10 spans across both services (`TUnit.Aspire.Tests` and `api-service`).

Jaeger search timeline

Jaeger trace waterfall showing the full request tree: `test case -> test body -> GET /trace-demo -> internal spans -> nested HTTP spans`.

Jaeger trace waterfall

## Testing
- `dotnet test TUnit.Aspire.Tests\TUnit.Aspire.Tests.csproj --framework net10.0 -- --treenode-filter "/*/*/TestTraceExporterTests/*"`
- `dotnet test TUnit.Aspire.Tests\TUnit.Aspire.Tests.csproj --framework net10.0 -- --treenode-filter "/*/*/BaggagePropagationHandlerTests/*"`
- `dotnet test TUnit.Aspire.Tests\TUnit.Aspire.Tests.csproj --framework net10.0 -- --treenode-filter "/*/*/OtlpReceiverTests/*"`
- `dotnet test TUnit.Aspire.Tests\TUnit.Aspire.Tests.csproj --framework net10.0 -- --treenode-filter "/*/*/OtlpCorrelationIntegrationTests/TraceDemo_Request_ProducesNestedCorrelatedLogs"`
- `DOTNET_DASHBOARD_OTLP_ENDPOINT_URL=http://127.0.0.1:4318 dotnet test TUnit.Aspire.Tests\TUnit.Aspire.Tests.csproj --framework net10.0 -- --treenode-filter "/*/*/OtlpCorrelationIntegrationTests/TraceDemo_Request_ProducesNestedCorrelatedLogs"`
- `dotnet build TUnit.Aspire\TUnit.Aspire.csproj -c Debug`


thomhurst:
@seahurricanelemonade @redoz I'm hoping this sorts it. The screenshots here look correct to me.

thomhurst:
Addressed in 80f5ae5:

**Issue 1** — `InjectBaggage` now guards `if (headers.Contains("baggage")) return;` so OTel `BaggagePropagator` output is preserved. Manual emit only fills the gap when the propagator (e.g. LegacyPropagator on older TFMs) didn't emit one.

**Issue 2** — Added `[InternalsVisibleTo("TUnit.AspNetCore.Tests")]` to `TUnit.AspNetCore.Core.csproj` and strong-named the test project (CS0281). Tests now reference `ActivityPropagationHandler` directly; reflection removed.

**Issue 3** — `InjectTraceContext` now guards with `!h.Contains(key)`, matching `TUnitBaggagePropagationHandler`'s never-overwrite policy. Pre-set `traceparent`/`tracestate` from test code is preserved.

**Issue 4** — Added defensive fallback in `FinishTestActivity`: when `result.Exception` is null but state is `Failed`/`Timeout`/`Cancelled`, span gets `error.type = <state>` so backends still flag the failure with attribution.

**Issue 5** — Extracted `TUnitActivitySource.DetachTestActivityFromAmbientContext()`. The `TestCoordinator` call site now self-documents the invariant.

**Minor (lifecycle source visibility)** — Acknowledged but not changed in this PR. Worth a follow-up: docs note + diagnostic when lifecycle spans exist with no listener.

thomhurst:
Addressed in efab86c (also folded in a code-reuse pass from \`/simplify\`):

**Concern 1 — TestTraceExporter static state, Dispose naming, parallel safety**
- Renamed \`Dispose\` → \`Stop\` (was misleading since it's not \`IDisposable\`).
- Extracted pure helpers: \`TryParseDashboardEndpoint(string?)\` and \`CreateTracerProvider(endpoint, context, sourceName)\`. Tests now build their own provider with a per-test \`Guid\` source name, never touch the env var, never touch the static field — they run **in parallel**. \`[NotInParallel]\` removed.
- Pushing back on the instance-class rewrite: hooks are static (TUnit's \`[Before(TestSession)]\` requires it), so the field still has to live in static space. \`TryStart\`/\`Stop\` already supports start→stop→start cycles for sequential sessions (\`Stop\` nulls the field). Same behaviour, less code.

**Concern 2 — Port TOCTOU race**
- \`OtlpTraceCaptureServer.CreateListener\` now retries up to 10× on \`HttpListenerException\`, disposing the failed listener between attempts. Loopback contention should be rare; bounded retry survives it.

**Concern 3 — Polling fragility**
- \`WaitForRequestAsync\` replaced with a \`TaskCompletionSource\` registered before the wait. \`ProcessRequestAsync\` signals matching waiters synchronously after enqueue. Timeout uses \`CancellationTokenSource\` registration. No more 50 ms busy-loop.

**Bonus (code-reuse cleanup)** — Both HTTP handlers had ~30 lines of duplicated W3C baggage encoding. Extracted to \`TUnitActivitySource.TryBuildBaggageHeader(activity)\` in TUnit.Core.

**Minor (\`GetServiceName\` fallback)** — Skipped per your acknowledgment that it can be a follow-up.

thomhurst:
Fixed in a5e10c6.

Replaced \`ConcurrentQueue\` + separate waiter lock with a single \`_stateLock\` covering both the request list and the waiter list. \`WaitForRequestAsync\` now scans the request list and registers the waiter atomically; \`ProcessRequestAsync\` calls a new \`EnqueueAndSignal\` that adds to the list and snapshots matching waiters under the same lock. Signalling runs outside the lock to avoid synchronous-continuation re-entry.

thomhurst:
Addressed in adcacb4.

**Issue 2 — Skip reason regression** — Confirmed real. \`FinishTestActivity\` moved from \`TestExecutor\` to \`TestCoordinator\`, so the local \`capturedException\` (which was the \`SkipTestException\`) is no longer reachable. \`TestCoordinator\`'s outer \`catch (SkipTestException)\` only fires for skips that bubble past \`TestExecutor\`, but \`TestExecutor\` catches them. Result: \`tunit.test.skip_reason\` was dropped for skips thrown from the test body.

Fix: in \`TestExecutor\`'s \`catch (SkipTestException)\`, also set \`Context.SkipReason ??= ex.Reason\`. \`FinishTestActivity\`'s existing first branch (\`if (skipReason is non-empty) ...\`) then tags the span correctly.

**Issue 3 — TOCTOU comment wording** — Changed "close the TOCTOU race" → "mitigate (cannot fully eliminate) the TOCTOU race". Retry loop reduces, doesn't eliminate.

**Pushing back**:
- **Issue 1 (session-scoped exporter state)** — Acknowledged as future-extensibility. Same response as last round: hooks are static so the field has to live in static space anyway, and \`TryStart\`/\`Stop\` already supports start→stop→start. Defer until in-process multi-session is an actual feature.
- **Issue 4 (rename \`Detach\` → \`Clear\`)** — \`Detach\` is the more accurate verb. The activity is **not** discarded — the doc explicitly says it's "kept alive so it can be finished with its final status and exported". Only the \`Activity.Current\` reference is severed; the activity object lives on. \`Clear\` would imply the opposite.
- **Issue 5** — Reviewer self-resolved ("looks fine on reflection").

PR #5598, opened 2026-04-17
fix: auto-suppress ExecutionContext flow for hosted services (#5589)

### LINKED ISSUE #5589: Auto-suppress ExecutionContext flow when starting IHostedService in TestWebApplicationFactory
## Problem

When integration tests use `WebApplicationFactory` with `IHostedService` implementations (e.g. message consumers like DotPulsar, timer-driven workers, connection pools), the hosted service's `StartAsync` captures the ambient `ExecutionContext` at the moment the factory is built. That context includes `Activity.Current`, which at that point is set to the test case Activity created by the TUnit engine.

Every subsequent activity emitted from the hosted service's background loop (on a thread-pool worker or long-running task) inherits that captured `Activity.Current` as its parent. Result:

- Spans from hosted-service work done during test B are attributed to test A's `TraceId`.
- When observed in Seq/Jaeger, activities appear mixed between tests.
- Static `ActivitySource` instances in third-party libraries amplify this because they are process-wide and capture the context of whichever test happened to start the service first.

See discussion #5479 for a concrete report of this symptom.

## Proposal

In `TestWebApplicationFactory<T>` (and `TracedWebApplicationFactory<T>`), decorate every registered `IHostedService` so its `StartAsync` runs with `ExecutionContext.SuppressFlow()` active. Background tasks spawned inside `StartAsync` then capture a clean context with `Activity.Current = null`, and any activities they later produce become orphan roots rather than leaking into a specific test's trace.

Sketch:

```csharp
// Inside TestWebApplicationFactory.ConfigureWebHost
builder.ConfigureServices(services =>
{
    services.Decorate<IHostedService>(inner => new FlowSuppressingHostedService(inner));
});

internal sealed class FlowSuppressingHostedService(IHostedService inner) : IHostedService
{
    public Task StartAsync(CancellationToken ct)
    {
        using var _ = ExecutionContext.SuppressFlow();
        return inner.StartAsync(ct);
    }

    public Task StopAsync(CancellationToken ct) => inner.StopAsync(ct);
}
```

## Considerations

- `Decorate` depends on `Scrutor` or a manual `ServiceDescriptor` replacement pass — avoid adding a dependency; iterate `IServiceCollection` and replace descriptors in place.
- Users who intentionally rely on ambient `Activity.Current` inside hosted services lose that linkage. Document and provide an opt-out (`WebApplicationTestOptions.SuppressHostedServiceFlow = false`).
- Hosted services registered inside `ConfigureTestServices` after our decorator runs will not be wrapped — need to run the decoration pass late (e.g. via `IStartupFilter` or `PostConfigure`).

## Related

- Discussion #5479 — "HTML report generation failed" / activities mixed between tests.
- Third-party static `ActivitySource` leak (tracked separately).

## Documentation

When this lands, update:
- `docs/docs/examples/opentelemetry.md` — replace the manual `ExecutionContext.SuppressFlow()` recipe in the "Spans from test A are showing up under test B" troubleshooting section with a note that `TestWebApplicationFactory<T>` now does this automatically.
- `docs/docs/guides/distributed-tracing.md` — update the "Static `ActivitySource`" limitation to reflect that hosted-service leak is auto-mitigated.

thomhurst:
## Summary

- `TestWebApplicationFactory<T>` now decorates every registered `IHostedService` so its `StartAsync` runs under `ExecutionContext.SuppressFlow()`. Background tasks spawned inside `StartAsync` capture a clean execution context, preventing spans from hosted-service work in test B being attributed to test A's `TraceId`.
- The wrapper (`FlowSuppressingHostedService`) also implements `IHostedLifecycleService` so `StartingAsync`/`StartedAsync`/`StoppingAsync`/`StoppedAsync` hooks keep firing for inner services that implement it — the Host uses an `is` check against the registered instance, and a plain `IHostedService` wrapper would silently drop those hooks.
- Opt-out via `protected virtual bool SuppressHostedServiceExecutionContextFlow => true` (override to return `false`).

## How

Decoration runs in an override of `CreateHost(IHostBuilder)`, which fires after all `ConfigureWebHost` / `WithWebHostBuilder` / `ConfigureTestServices` callbacks — so it catches hosted services registered via the isolated-factory path as well as the base factory.

Resolves #5589.

## Test plan

- [x] `StartAsync_SuppressesFlow_IntoSpawnedTasks` — spawned `Task.Run` inside `StartAsync` sees `Activity.Current == null` even when an outer Activity is active at factory-build time.
- [x] `OptOut_PreservesFlow_IntoSpawnedTasks` — when override returns `false`, spawned task inherits the outer Activity.
- [x] `RegisteredHostedServices_AreWrapped` — resolved `IHostedService` instances are of the wrapper type.
- [x] `IsolatedFactory_HostedServices_AreWrapped` — services added via `GetIsolatedFactory` / `ConfigureTestServices` are also wrapped.
- [x] Full `TUnit.AspNetCore.Tests` suite (20 tests) green on net10.0.

## Docs

- `docs/docs/examples/opentelemetry.md` — "Spans from test A are showing up under test B" section updated to note the auto-fix.
- `docs/docs/guides/distributed-tracing.md` — Static `ActivitySource` limitation updated.

### DIFF
--- a/TUnit.AspNetCore.Core/FlowSuppressingHostedService.cs
+++ b/TUnit.AspNetCore.Core/FlowSuppressingHostedService.cs
@@ -1,0 +1,51 @@
+using Microsoft.Extensions.Hosting;
+
+namespace TUnit.AspNetCore;
+
+/// <summary>
+/// Wraps an <see cref="IHostedService"/> so its <see cref="IHostedService.StartAsync"/>
+/// runs with <see cref="ExecutionContext.SuppressFlow"/> active. Background tasks
+/// spawned inside <c>StartAsync</c> capture a clean <see cref="ExecutionContext"/>,
+/// so activities they later emit do not inherit the test's ambient
+/// <see cref="System.Diagnostics.Activity.Current"/> as their parent.
+/// </summary>
+/// <remarks>
+/// Implements <see cref="IHostedLifecycleService"/> so the Host's lifecycle hooks keep
+/// firing for inner services that implement it — the Host uses an <c>is</c> check
+/// against the registered instance, so without passthrough wrapping would silently
+/// drop those hooks.
+/// </remarks>
+internal sealed class FlowSuppressingHostedService(IHostedService inner) : IHostedLifecycleService
+{
+    public Task StartAsync(CancellationToken cancellationToken) =>
+        Suppressed(inner.StartAsync, cancellationToken);
+
+    public Task StopAsync(CancellationToken cancellationToken) =>
+        inner.StopAsync(cancellationToken);
+
+    public Task StartingAsync(CancellationToken cancellationToken) =>
+        inner is IHostedLifecycleService lifecycle
+            ? Suppressed(lifecycle.StartingAsync, cancellationToken)
+            : Task.CompletedTask;
+
+    public Task StartedAsync(CancellationToken cancellationToken) =>
+        inner is IHostedLifecycleService lifecycle
+            ? Suppressed(lifecycle.StartedAsync, cancellationToken)
+            : Task.CompletedTask;
+
+    public Task StoppingAsync(CancellationToken cancellationToken) =>
+        inner is IHostedLifecycleService lifecycle
+            ? lifecycle.StoppingAsync(cancellationToken)
+            : Task.CompletedTask;
+
+    public Task StoppedAsync(CancellationToken cancellationToken) =>
+        inner is IHostedLifecycleService lifecycle
+            ? lifecycle.StoppedAsync(cancellationToken)
+            : Task.CompletedTask;
+
+    private static Task Suppressed(Func<CancellationToken, Task> op, CancellationToken ct)
+    {
+        using var _ = ExecutionContext.SuppressFlow();
+        return op(ct);
+    }
+}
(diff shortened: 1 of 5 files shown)


thomhurst:
Addressed in 81309cb:

**#1 (sync-only suppression):** Pushed further than the doc-note suggestion. Swapped the wrapper to combine `SuppressFlow` with `Task.Run`, so `inner.StartAsync` runs on a thread-pool worker whose ExecutionContext was captured under suppression. `Activity.Current` starts null on that thread and stays null through every await point — background tasks spawned anywhere inside `StartAsync` (before or after an `await`) inherit a clean context. Added `StartAsync_SuppressesFlow_WhenSpawnIsAfterAwait` covering the `await ...; _ = Task.Run(...)` pattern. Trade-off: always forces a thread-pool hop for hosted-service startup in tests — negligible vs. host build cost, and the thread-affinity case is vanishingly rare for hosted services.

**#2 (asymmetric suppression):** Added a comment on the stop-lifecycle methods explaining the intent — stop methods typically signal cancellation and await shutdown rather than spawning long-running background work, so Stop isn't the span-leak vector that Start is.

**#3 (structural tests):** Kept. The behavioral test (`StartAsync_SuppressesFlow_IntoSpawnedTasks`) is the primary assertion; the structural ones (`RegisteredHostedServices_AreWrapped`, `IsolatedFactory_HostedServices_AreWrapped`) verify a different question — *did wrapping happen at all?* — for the isolated-factory path where there's no outer Activity to assert against behaviorally. Both styles have value.

PR #5601, opened 2026-04-17
feat: TUnit0064 analyzer + code fix for direct WebApplicationFactory inheritance

### LINKED ISSUE #5596: Roslyn analyzer + code fix: convert WebApplicationFactory<T> to TestWebApplicationFactory<T>
## Problem

Users who inherit from `Microsoft.AspNetCore.Mvc.Testing.WebApplicationFactory<T>` directly silently lose:

- Trace propagation (no `traceparent`/`baggage` on outgoing requests).
- Per-test logging correlation.
- `TestContext.Current` resolution inside request handlers.

There is no compile-time signal. Discussion #5479 is one of multiple reports rooted in this exact mistake.

## Proposal

Extend the existing `WebApplicationFactoryAccessAnalyzer` (or add a sibling rule) in `TUnit.AspNetCore.Analyzers` that flags:

```csharp
public class MyFactory : WebApplicationFactory<Program> // TUNITxxxx
{
}
```

with severity `Warning` (not `Error` — some users may have legitimate reasons), and a code fix that converts to:

```csharp
public class MyFactory : TestWebApplicationFactory<Program>
{
}
```

The code fix should also offer the alternative wrapping pattern when the inheritance can't be changed (e.g. third-party base class):

```csharp
var factory = new TracedWebApplicationFactory<Program>(existingFactory);
```

## Considerations

- **Scope to test projects.** Detect via `IsTestProject`-style heuristics: `Microsoft.Testing.Platform` reference, or namespace contains `.Tests`. Avoid false positives in production code.
- **Opt-out.** Standard `#pragma warning disable` plus `.editorconfig` rule severity.
- **Documentation link in diagnostic.** `helpLinkUri` pointing to the new [Distributed Tracing](https://github.com/thomhurst/TUnit/blob/main/docs/docs/guides/distributed-tracing.md) page.
- **Diagnostic ID.** Allocate next free `TUNITxxxx` in the analyzer table.

## Related

- Discussion #5479.
- Existing analyzer: `TUnit.AspNetCore.Analyzers/WebApplicationFactoryAccessAnalyzer.cs`.

## Documentation

When this lands, update:
- `docs/docs/examples/aspnet.md` — update the existing "Use `TestWebApplicationFactory<T>`" warning callout to mention the new analyzer and its diagnostic ID, so users understand the warning they'll see in the IDE.
- `docs/docs/guides/distributed-tracing.md` — link the diagnostic ID from the "WebApplicationFactory without TUnit's wrapper" limitation entry.
- Analyzer rules documentation (if a per-rule reference exists) — add an entry for the new `TUNITxxxx` ID.

thomhurst:
## Summary

- New analyzer `TUnit0064` (Warning) — flags classes directly inheriting `Microsoft.AspNetCore.Mvc.Testing.WebApplicationFactory<T>` instead of `TUnit.AspNetCore.TestWebApplicationFactory<T>`.
- New code fix — rewrites the base type to `TestWebApplicationFactory<T>` and adds `using TUnit.AspNetCore;`.
- New `TUnit.AspNetCore.Analyzers.CodeFixers` project, bundled into the `TUnit.AspNetCore` nupkg.
- Docs updated in `examples/aspnet.md` and `guides/distributed-tracing.md`.

Closes #5596.

## Details

Direct inheritance silently loses:

- **Distributed tracing** — server-side spans no longer link back to the triggering test.
- **Per-test logging** — `ILogger` output from inside the app isn't routed to the right test output.
- **Test context** — `TestContext.Current` won't resolve inside request handlers.

The analyzer reports on the base type location (not the class identifier) so the squiggle sits where the offending inheritance is declared. It ignores `TestWebApplicationFactory<T>` itself. Transitive cases (class inherits a user type that inherits `TestWebApplicationFactory<T>`) are implicitly safe: the direct base isn't `WebApplicationFactory`, so the analyzer exits early.

## Test plan

- [x] Analyzer warns on direct inheritance
- [x] No warning when using `TestWebApplicationFactory<T>`
- [x] No warning on transitive `TestWebApplicationFactory<T>` inheritance
- [x] Diagnostic reports at base-type location (not class identifier)
- [x] Code fix rewrites base type and adds `using TUnit.AspNetCore;`
- [x] Verified codefix DLL is bundled at `analyzers/dotnet/cs/` in the `TUnit.AspNetCore` nupkg
- [x] 21/21 analyzer tests pass (net9.0)

### DIFF
--- a/TUnit.AspNetCore.Analyzers.CodeFixers/TUnit.AspNetCore.Analyzers.CodeFixers.csproj
+++ b/TUnit.AspNetCore.Analyzers.CodeFixers/TUnit.AspNetCore.Analyzers.CodeFixers.csproj
@@ -1,0 +1,37 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <Nullable>enable</Nullable>
+    <LangVersion>latest</LangVersion>
+    <EnforceExtendedAnalyzerRules>true</EnforceExtendedAnalyzerRules>
+    <IsRoslynComponent>true</IsRoslynComponent>
+    <DevelopmentDependency>false</DevelopmentDependency>
+    <NoPackageAnalysis>true</NoPackageAnalysis>
+    <RootNamespace>TUnit.AspNetCore.Analyzers.CodeFixers</RootNamespace>
+    <AssemblyName>TUnit.AspNetCore.Analyzers.CodeFixers</AssemblyName>
+    <NoWarn>RS2003</NoWarn>
+    <EnableTrimAnalyzer>false</EnableTrimAnalyzer>
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+  <ItemGroup>
+    <AssemblyAttribute Include="System.Runtime.CompilerServices.InternalsVisibleToAttribute">
+      <_Parameter1>TUnit.AspNetCore.Analyzers.Tests</_Parameter1>
+    </AssemblyAttribute>
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.Analyzers">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+    <PackageReference Include="Microsoft.CodeAnalysis.Common" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" />
+    <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.Common" />
+  </ItemGroup>
+  <ItemGroup>
+    <AdditionalFiles Remove="AnalyzerReleases.Unshipped.md" />
+    <AdditionalFiles Remove="AnalyzerReleases.Shipped.md" />
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\TUnit.AspNetCore.Analyzers\TUnit.AspNetCore.Analyzers.csproj" />
+  </ItemGroup>
+</Project>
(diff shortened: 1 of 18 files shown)


thomhurst:
Thanks for the review. Responses:

**1. Packaging:** The single `analyzers/dotnet/cs` path matches how `TUnit.Analyzers.CodeFixers` is already packed in `TUnit.Core.csproj` (line 48-50) — the versioned paths only host the analyzer DLL itself, and NuGet falls back from `roslynX.Y/cs` to `dotnet/cs` for the codefixer. Following the established pattern, so leaving as-is.

**2 & 3. Namespace-scoped usings + test:** Real gap, fixed in 5aeb7cdeb. `AddUsingIfMissing` now also scans `BaseNamespaceDeclarationSyntax.Usings` (and aliases/static directives are excluded from the match). Added two tests: one for pre-existing top-level using, one for namespace-scoped.

**4. Using order:** Agreed — purely cosmetic and downstream formatters sort. Leaving to avoid scope creep.

23/23 tests pass on net9.0.

thomhurst:
Thanks for round 2:

**1. Packaging (re-raised):** Same situation as the first review — the existing `TUnit.Analyzers.CodeFixers` ships only at `analyzers/dotnet/cs` (see `TUnit.Core.csproj` line 48-50) and works across all Roslyn variants. NuGet falls back from the versioned analyzer paths to the unversioned codefixer path. Matching the established pattern, not changing.

**2. Null-guard asymmetry:** Good call. Fixed in cbdc125a2 — now bailing out at the compilation-start action when `TestWebApplicationFactory<T>` can't be resolved. That means the analyzer only fires when the code fix can actually apply. `testWebApplicationFactory` is now non-nullable in the symbol action.

**3. File-scoped namespace test:** Added `Does_Not_Duplicate_Using_In_File_Scoped_Namespace`. Verifier gained a `stubsSource` overload so the stubs can live in a separate compilation source (required because file-scoped namespace has to precede every member).

**4. Partial class test:** Added `Warning_Fires_Once_For_Partial_Class`. Diagnostic fires exactly once; `GetBaseTypeLocation` returns the first syntax reference with a base list.

25/25 tests pass on net9.0.

PR #5602, opened 2026-04-17
feat: TUnit.OpenTelemetry zero-config tracing package

### LINKED ISSUE #5591: Ship opt-in OpenTelemetry ActivityProcessor that tags spans with tunit.test.id
## Problem

Third-party libraries that use static `ActivitySource` instances (DotPulsar, some EF Core providers, several message-broker clients, connection pools) emit spans from background threads whose `Activity.Current` was captured at some arbitrary earlier point — often during a different test, or during host startup. The parent-chain is effectively broken from the engine's perspective:

- `Activity.ParentSpanId` points at a span the engine did not create and may never have seen.
- Re-parenting is impossible after the fact; `Activity` does not expose a writeable parent.
- These spans appear as orphan roots in Seq/Jaeger, or worse, under the wrong test's `TraceId`.

There is no way to automate a fix at the parent-chain level. But users typically only need to *query* or *filter* by test — hierarchy is secondary.

## Proposal

Ship an optional OpenTelemetry `BaseProcessor<Activity>` in `TUnit.AspNetCore` (or a new `TUnit.OpenTelemetry` package). On `OnStart`:

1. Check `Activity.Current?.GetBaggageItem(""tunit.test.id"")` — baggage propagates through async/await and captured `ExecutionContext`, so it is reliable even when parent chain is broken.
2. If present, copy it onto the new activity as a tag (`tunit.test.id`) and optionally `tunit.test.node_uid`.
3. Leave parent/trace IDs untouched — preserve whatever the emitting library set.

```csharp
public sealed class TUnitTestCorrelationProcessor : BaseProcessor<Activity>
{
    public override void OnStart(Activity activity)
    {
        if (activity.GetTagItem(TUnitActivitySource.TagTestId) is not null)
            return; // already tagged (TUnit-emitted)

        var testId = Activity.Current?.GetBaggageItem(TUnitActivitySource.TagTestId);
        if (testId is not null)
            activity.SetTag(TUnitActivitySource.TagTestId, testId);
    }
}
```

User wires it up explicitly in their OTel setup:

```csharp
services.AddOpenTelemetry().WithTracing(t => t
    .AddProcessor<TUnitTestCorrelationProcessor>()
    .AddAspNetCoreInstrumentation()
    // ...
    .AddOtlpExporter());
```

## Considerations

- Baggage only propagates if the ambient `Activity.Current` was set by the engine at some ancestor point — works for in-process WAF scenarios. Does not help if the library detached from the async context entirely (e.g. dedicated long-running thread started before any test).
- Opt-in, not automatic: users must register the processor themselves. Automation would require a module initializer hooking into OTel, which conflicts with user-controlled configuration.
- Document querying: in Seq, `@trace_id` grouping fails cross-test; users should filter `tunit.test.id = ""<id>""`.
- Should live in its own package (`TUnit.OpenTelemetry`?) to avoid forcing an OTel dependency on all `TUnit.AspNetCore` consumers.

## Related

- Discussion #5479 — "activities being mixed between tests" in Seq/Jaeger.
- Complements the `IHostedService` flow-suppression approach (separate issue) — that issue stops the leak at the source; this issue makes spans queryable even when the leak happens in a library we cannot modify.

## Documentation

When this lands, update:
- `docs/docs/examples/opentelemetry.md` — replace the manual `TUnitTagProcessor` definition in the "Spans from test A are showing up under test B" section with the package's one-liner registration.
- `docs/docs/guides/distributed-tracing.md` — update the "Static `ActivitySource`" limitation entry to point at the shipped processor.

### LINKED ISSUE #5593: New TUnit.OpenTelemetry package: auto-start TracerProvider in test process
## Problem

Today, every user has to wire OpenTelemetry by hand — even for a basic console exporter:

```csharp
public class TraceSetup
{
    private static TracerProvider? _provider;

    [Before(TestDiscovery)]
    public static void Setup()
    {
        _provider = Sdk.CreateTracerProviderBuilder()
            .AddSource(""TUnit"")
            .AddSource(""TUnit.Lifecycle"")
            .AddSource(""TUnit.AspNetCore.Http"")
            .AddOtlpExporter(opts => opts.Endpoint = new Uri(""..."")) // hand-typed
            .Build();
    }

    [After(TestSession)]
    public static void Teardown() => _provider?.Dispose();
}
```

This is boilerplate. The source list is fixed. The OTLP endpoint is conventionally configured via `OTEL_EXPORTER_OTLP_ENDPOINT` already.

## Proposal

Ship a new package `TUnit.OpenTelemetry` that registers a `TracerProvider` automatically when:

- The user has the `OpenTelemetry.Exporter.OpenTelemetryProtocol` package referenced (detect via `Type.GetType`), AND
- `OTEL_EXPORTER_OTLP_ENDPOINT` is set, OR `OTEL_SDK_DISABLED` is not `true`.

Implementation sketch:

```csharp
[ModuleInitializer]
internal static void AutoStart()
{
    if (Environment.GetEnvironmentVariable(""OTEL_SDK_DISABLED"") == ""true"") return;
    if (Type.GetType(""OpenTelemetry.Sdk, OpenTelemetry"") is null) return;

    _provider = Sdk.CreateTracerProviderBuilder()
        .AddSource(""TUnit"", ""TUnit.Lifecycle"", ""TUnit.AspNetCore.Http"")
        .AddOtlpExporter() // honors OTEL_EXPORTER_OTLP_* env vars
        .Build();

    AppDomain.CurrentDomain.ProcessExit += (_, _) => _provider?.Dispose();
}
```

User installs the package and sets `OTEL_EXPORTER_OTLP_ENDPOINT`. No code.

## Considerations

- **Hard dependency or soft?** Prefer soft (reflective) so the package doesn't pin OTel SDK versions.
- **Lifecycle.** `[After(TestSession)]` is more reliable than `ProcessExit` for flush — wire via internal hook event rather than `AppDomain` shutdown.
- **User wants to customize.** Provide a `TUnitOpenTelemetry.Configure(builder => ...)` static hook callable from `[Before(TestDiscovery)]` that runs before the auto-build.
- **Opt-out.** Env var `TUNIT_OTEL_AUTOSTART=0`.
- **Detection of additional sources.** Optionally auto-add common SUT sources (`Microsoft.AspNetCore.Hosting`, `System.Net.Http`, `Microsoft.EntityFrameworkCore`) when their assemblies are loaded.

## Related

- Sister issue: SUT-side auto-wire via `TestWebApplicationFactory` (separate ticket).
- #5591 — `TUnitTagProcessor` would be auto-registered into this provider as well.

## Documentation

When this lands, update:
- `docs/docs/examples/opentelemetry.md` — add a new "Zero-config setup" subsection at the top of "Setup" describing the `TUnit.OpenTelemetry` package + env-var workflow. Demote the current "Option A: OpenTelemetry SDK" content to "Manual setup" for users who want full control.
- `docs/docs/guides/distributed-tracing.md` — mention the auto-start path in "Backend setup".
- New page or section documenting `TUnitOpenTelemetry.Configure(...)` hook and the `TUNIT_OTEL_AUTOSTART=0` opt-out.

thomhurst:
## Summary

- New `TUnit.OpenTelemetry` NuGet package auto-wires a `TracerProvider` at `[Before(TestDiscovery)]` and disposes at `[After(TestSession)]`. Install the package + set `OTEL_EXPORTER_OTLP_ENDPOINT` — zero boilerplate required.
- Includes `TUnitTestCorrelationProcessor` (resolves #5591's tag-propagation workaround) so spans from broken-parent-chain libraries still carry `tunit.test.id`.
- `TUnit.Aspire` refactored to delegate provider construction via `TUnitOpenTelemetry.Configure`, removing its own `TracerProvider` singleton.
- Coexists cleanly with hand-rolled setups: `[Before(TestDiscovery, Order = int.MaxValue)]` runs last, probes `ActivitySource("TUnit").HasListeners()`, and steps aside if a listener is already attached. `TUNIT_OTEL_AUTOSTART=0` opts out; `=1` force-enables.

Resolves #5593. Bundles fix for #5591.

## Architecture notes

- Public surface: `TUnitOpenTelemetry.Configure(Action<TracerProviderBuilder>)`, `AutoStart` (hooks + `AutoStartOrder` const), `TUnitTestCorrelationProcessor`.
- Gate logic: if `TUNIT_OTEL_AUTOSTART != "1"` and `OTEL_EXPORTER_OTLP_ENDPOINT` is unset and no user `Configure` callbacks are registered, the package stays dormant (no provider built, no listener attached).
- Default `service.name` resource: `OTEL_SERVICE_NAME` > entry assembly name > `"TUnit.Tests"`.
- `builder.Build()` and user callbacks execute outside the `AutoStart` lock to avoid holding a static lock across SDK initialization.
- OTLP exporter is hard-dep'd so the zero-config path doesn't require a second `dotnet add package`.

## Docs

- `docs/docs/examples/opentelemetry.md` gains **Option A: Zero-config**; existing manual setup demoted to Option B / Option C.
- `docs/docs/guides/distributed-tracing.md` gains a zero-config callout pointing to the new option.
- Manual `TUnitTagProcessor` snippet in troubleshooting now points at the pre-registered package processor.

## Test plan

- [x] `TUnit.OpenTelemetry.Tests` — 13/13 pass (Configure, AutoStart coexistence, opt-out env var, correlation processor baggage-to-tag, default service.name, end-to-end InMemory exporter smoke test)
- [x] `TUnit.Aspire.Tests` — 101/101 pass (no regression from the refactor)
- [x] `TUnit.PublicAPI` — new OpenTelemetry snapshot tests pass across net8/net9/net10
- [x] `TUnit.Engine.Tests` hook filter — 17 pass, 16 skip, 0 fail (no regression from the new `Order = int.MaxValue` hook)
- [x] `dotnet pack TUnit.OpenTelemetry` produces a valid nupkg
- [x] `docs && npm run build` — clean build, no broken links
- [ ] Manual: end-to-end trace verification against a local Jaeger/Tempo instance with `OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317`

## Pipeline

- `TUnit.Pipeline/Modules/GetPackageProjectsModule.cs` adds the new package to the pack/publish list.
- New `RunOpenTelemetryTestsModule` mirrors the Aspire test module (no OS gate — nothing platform-specific).
- All three slnx files (`TUnit.slnx`, `TUnit.Dev.slnx`, `TUnit.CI.slnx`) updated.

## Plan

Full implementation plan: `docs/plans/2026-04-17-tunit-opentelemetry-package.md` (14 tasks, each committed separately).

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -65,6 +65,7 @@
     <PackageVersion Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="10.0.1" />
     <PackageVersion Include="NSubstitute" Version="5.3.0" />
     <PackageVersion Include="NuGet.Protocol" Version="7.3.1" />
+    <PackageVersion Include="OpenTelemetry" Version="1.15.2" />
     <PackageVersion Include="OpenTelemetry.Exporter.OpenTelemetryProtocol" Version="1.15.2" />
     <PackageVersion Include="OpenTelemetry.Extensions.Hosting" Version="1.15.2" />
     <PackageVersion Include="OpenTelemetry.Instrumentation.AspNetCore" Version="1.15.1" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -66,6 +66,7 @@
     <PackageVersion Include="NSubstitute" Version="5.3.0" />
     <PackageVersion Include="NuGet.Protocol" Version="7.3.1" />
     <PackageVersion Include="OpenTelemetry" Version="1.15.2" />
+    <PackageVersion Include="OpenTelemetry.Exporter.InMemory" Version="1.15.2" />
     <PackageVersion Include="OpenTelemetry.Exporter.OpenTelemetryProtocol" Version="1.15.2" />
     <PackageVersion Include="OpenTelemetry.Extensions.Hosting" Version="1.15.2" />
     <PackageVersion Include="OpenTelemetry.Instrumentation.AspNetCore" Version="1.15.1" />
--- a/TUnit.Aspire.Tests/TestTraceExporterTests.cs
+++ b/TUnit.Aspire.Tests/TestTraceExporterTests.cs
@@ -57,41 +57,6 @@ public class TestTraceExporterTests
     }
 
     [Test]
-    public async Task CreateTracerProvider_ExportsTracesForRegisteredSource()
-    {
-        await using var server = new OtlpTraceCaptureServer();
-        server.Start();
-
-        // Per-test ActivitySource name keeps spans isolated from any other test or production
-        // listener, so this test stays parallel-safe even though OpenTelemetry exporters are
-        // process-wide.
-        var sourceName = $"TUnit.Tests.{Guid.NewGuid():N}";
-        var endpoint = new Uri($"http://127.0.0.1:{server.Port}");
-
-        using (var activitySource = new ActivitySource(sourceName))
-        using (var provider = TestTraceExporter.CreateTracerProvider(
-            endpoint, GetCurrentSessionContext(), sourceName))
-        {
-            using var testCase = activitySource.StartActivity("test case", ActivityKind.Internal);
-            using var testBody = activitySource.StartActivity("test body", ActivityKind.Internal);
-
-            await Assert.That(testCase).IsNotNull();
-            await Assert.That(testBody).IsNotNull();
-
-            testBody?.Stop();
-            testCase?.Stop();
-        }
-        // Disposing the provider flushes pending spans to the exporter.
-
-        var request = await server.WaitForRequestAsync("/v1/traces");
-        var body = Encoding.UTF8.GetString(request.Body);
-
-        await Assert.That(body).Contains("test case");
-        await Assert.That(body).Contains("test body");
-        await Assert.That(body).Contains(typeof(TestTraceExporterTests).Assembly.GetName().Name!);
-    }
-
-    [Test]
     public async Task GetTracesEndpoint_AppendsSignalPathWithoutDroppingBasePath()
     {
         var endpoint = new Uri("http://127.0.0.1:5341/ingest/otlp");
--- a/TUnit.Aspire.Tests/TestTraceExporterTests.cs
+++ b/TUnit.Aspire.Tests/TestTraceExporterTests.cs
@@ -1,5 +1,7 @@
 using System.Diagnostics;
 using System.Text;
+using OpenTelemetry;
+using OpenTelemetry.Trace;
 using TUnit.Aspire.Telemetry;
 using TUnit.Aspire.Tests.Helpers;
 using TUnit.Assertions;
@@ -27,6 +29,31 @@ public class TestTraceExporterTests
     {
         await Assert.That(TestTraceExporter.TryParseDashboardEndpoint("http://127.0.0.1:4317"))
             .IsEqualTo(new Uri("http://127.0.0.1:4317"));
+    }
+
+    [Test]
+    public async Task AddToBuilder_ExportsTracesForRegisteredSource()
+    {
+        await using var server = new OtlpTraceCaptureServer();
+        server.Start();
+
+        var sourceName = $"TUnit.Tests.{Guid.NewGuid():N}";
+        var endpoint = new Uri($"http://127.0.0.1:{server.Port}");
+
+        var builder = Sdk.CreateTracerProviderBuilder().AddSource(sourceName);
+        TestTraceExporter.AddToBuilder(builder, GetCurrentSessionContext(), endpoint);
+
+        using (var activitySource = new ActivitySource(sourceName))
+        using (var provider = builder.Build())
+        {
+            using var testCase = activitySource.StartActivity("add-to-builder case", ActivityKind.Internal);
+            testCase?.Stop();
+        }
+
+        var request = await server.WaitForRequestAsync("/v1/traces");
+        var body = Encoding.UTF8.GetString(request.Body);
+
+        await Assert.That(body).Contains("add-to-builder case");
     }
 
     [Test]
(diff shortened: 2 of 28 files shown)


thomhurst:
Thanks for the review. Addressed issues 2–4 in 532a583a7; pushing back on issue 1 with rationale below.

**Issue 2 (TOCTOU)** — moved the `HasListeners`/endpoint/`HasConfiguration` checks inside the existing `_lock` block so they sit adjacent to the `_provider is null` guard. `Build()` still runs outside the lock; the post-Build re-check disposes the loser on a race. The race the reviewer flagged (listener attached between check and Build completion) is narrowed but not eliminated — holding the lock across `Build()` would serialize SDK init, which the original design intentionally avoids.

**Issue 3 (null-conditional)** — `provider?.Dispose()` → `provider.Dispose()` after `Build()`. `Build()` throwing is propagated out of `Start` (no half-initialized provider to leak).

**Issue 4 (test wrapper in production)** — `TestTraceExporter.CreateTracerProvider` deleted. The sole consumer, `CreateTracerProvider_ExportsTracesForRegisteredSource`, was a near-duplicate of `AddToBuilder_ExportsTracesForRegisteredSource` — kept the latter since it exercises the production wiring path directly.

**Issue 1 (hard-coded `TUnit.AspNetCore.Http` source)** — not acting on this. The proposed fix (self-registration via `TUnitOpenTelemetry.RegisterActivitySource(...)` from `TUnit.AspNetCore`) would force `TUnit.AspNetCore` to take a dep on `TUnit.OpenTelemetry`, which the implementation plan explicitly rules out:

> `TUnit.AspNetCore` **not** modified — it must remain OTel-agnostic.

The three default sources are an intentional "known TUnit sources" list, not an extensibility hook. Users who want to add or swap sources can already do so via `TUnitOpenTelemetry.Configure(b => b.AddSource("..."))`. If the list grows past ~3–4 entries or a third-party package wants its own auto-wired source, the self-registration pattern is a sensible follow-up — just not worth the API surface expansion for one current integration.

Rerun results:
- `TUnit.OpenTelemetry.Tests` — 13/13
- `TUnit.Aspire.Tests` — 100/100 (was 101; `CreateTracerProvider_*` removed)
- `TUnit.PublicAPI` OpenTelemetry snapshots — pass across net8/net9/net10

PR #5606, opened 2026-04-17
feat: generalize OTLP receiver for use outside TUnit.Aspire

### LINKED ISSUE #5595: Generalize TUnit.Aspire's OTLP receiver for use outside Aspire
## Problem

`TUnit.Aspire` already runs an in-process OTLP receiver (`OtlpReceiver`) that accepts spans and logs from Aspire-hosted services and routes them into TUnit's HTML report and per-test output. Outside of Aspire — plain `WebApplicationFactory` integration tests, or out-of-process SUTs — there is no equivalent. SUT spans only reach the user's external backend (if configured) and never appear in TUnit's HTML report.

## Proposal

Extract the OTLP receiver logic from `TUnit.Aspire` into a shared internal component (or a new `TUnit.OpenTelemetry.Receiver` package) and have it auto-start in any test process that has `TUnit.OpenTelemetry` referenced.

When started:

1. Bind to `localhost:<random-port>`.
2. Set `OTEL_EXPORTER_OTLP_ENDPOINT` for child processes the test spawns (so a SUT launched via `Process.Start` auto-points back at the test).
3. For in-process WAF scenarios, wire the SUT's `OpenTelemetry` SDK at it via `TestWebApplicationFactory.ConfigureWebHost`.
4. Forward received spans into:
   - `ActivityCollector` so they appear in the HTML report.
   - The user's configured `TracerProvider` (if any) so they reach external backends too.

## Considerations

- **Process-wide HTTP server.** Need a lightweight Kestrel + grpc handler. `TUnit.Aspire`'s implementation is the reference; extract carefully so it doesn't pull Aspire dependencies.
- **Port conflicts.** Use port `0` and broadcast the chosen port via env var.
- **Security.** Bind to loopback only; reject non-loopback origins explicitly.
- **TF support.** Receiver requires .NET 8+ (matches the rest of the OTel work).
- **Memory.** Already-good caps in `ActivityCollector` (per-test and per-trace) should apply equally.

## Result

User installs `TUnit.OpenTelemetry`, inherits from `TestWebApplicationFactory<T>`. Every span the SUT emits — including from instrumented libraries inside the SUT — appears in the HTML report under the originating test, with no exporter wiring or env var setup.

## Related

- Aspire receiver code: `TUnit.Aspire/Telemetry/OtlpReceiver.cs`.
- Sister tickets: `TUnit.OpenTelemetry` package, `TestWebApplicationFactory` SUT-side auto-wire.

## Documentation

When this lands, update:
- `docs/docs/guides/distributed-tracing.md` — add a new section "Capturing SUT spans in the HTML report" explaining the in-proc OTLP receiver and the env vars it sets for child processes.
- `docs/docs/guides/html-report.md` — note that out-of-process SUT spans now appear in the report when using `TUnit.OpenTelemetry`.
- `docs/docs/examples/aspnet.md` — example for out-of-process SUT scenarios that previously couldn't show in the HTML report.

thomhurst:
## Summary

- Extracts `OtlpReceiver` + `OtlpLogParser` from `TUnit.Aspire` into `TUnit.OpenTelemetry`, adds `OtlpTraceParser`, and auto-starts the receiver at test discovery so non-Aspire tests (plain `WebApplicationFactory`, spawned processes, testcontainers, external services) can export spans into TUnit's HTML report over OTLP/HTTP.
- `ActivityCollector` gains `RegisterExternalTrace` / `IngestExternalSpan` + a process-wide `Current` pointer so the receiver routes spans without explicit wiring. Unknown trace IDs are dropped; registered traces capped at 100 external spans. Opt out with `TUNIT_OTEL_RECEIVER=0`.
- Docs: new "Capturing spans from out-of-process SUTs" section in `distributed-tracing.md`; comparison table updated.

Closes #5595. Builds on #5602.

## Test plan

- [x] `TUnit.OpenTelemetry.Tests` — 27 tests, 3 consecutive runs green (covers ingestion, caps, auto-start hook, OTLP parser, receiver round-trip, upstream forwarding).
- [x] `TUnit.Aspire` + `TUnit.Aspire.Tests` still build — Aspire consumes the extracted receiver via the new namespace.
- [x] Full solution build clean.
- [ ] CI green on PR.


thomhurst:
Thanks for the review. Addressed two concrete items in 3e73e075a; responding inline to the rest.

**Fixed**
- **Moderate (protobuf docs)** — added proto spec URLs + field-name comments at the top of `OtlpTraceParser` / `OtlpLogParser` so magic field numbers can be cross-referenced.
- **Minor (configurable cap)** — external span cap now overridable via `TUNIT_OTEL_MAX_EXTERNAL_SPANS`, with a one-time stderr warning when first hit so users can discover the knob.

**Discussing**
- **Major 1 (IVT + static singleton)** — keeping as-is. `TUnit.OpenTelemetry` and `TUnit.Engine` are sibling assemblies shipped together; IVT is the established pattern here (same relationship as `TUnit` ↔ `TUnit.Engine`). An interface + service-locator would add public API surface for a single implementation, and the whole point of this PR is zero-wiring auto-ingestion. `Interlocked.CompareExchange(ref _current, this, null)` already gives first-wins semantics.
- **Major 2 #1 (singleton race)** — I think there's a misread: `Start()` uses `Interlocked.CompareExchange(ref _current, this, null)` (line 44) and `Stop()` uses `Interlocked.CompareExchange(ref _current, null, this)` (line 125). `CompareExchange` is already there.
- **Major 2 #2 (multi-session swap vs. in-flight spans)** — acknowledged but theoretical in practice: `HtmlReporter` creates one collector per session before any test runs, so within a single process the slot is claimed and held. Cross-session swapping would require multiple concurrent `HtmlReporter` instances in one process, which isn't a supported scenario today. Happy to revisit if someone hits it.
- **Minor #2 (helper duplication)** — the two `OtlpTraceCaptureServer` copies already share the extracted `LoopbackHttpListenerFactory` for the tricky port-binding logic. Full consolidation would require a new shared test helper assembly; the remaining duplication is thin glue (HttpListener lifecycle + waiter plumbing).

thomhurst:
Thanks — both bugs are real. Fixed in d76bfeeba.

**Case-sensitivity** — You're right that `Convert.ToHexString` returns uppercase and `Activity.TraceId.ToString()` returns lowercase. The root fix is in the parser: it now emits lowercase (via a small `HexLower` helper that uses `Convert.ToHexStringLower` on net9+ and falls back to `Convert.ToHexString(…).ToLowerInvariant()` on net8). That way external spans naturally share dictionary keys with in-process spans without caller ceremony.

As defense-in-depth I also flipped the remaining trace/span-ID dicts in `ActivityCollector` (`_spansByTrace`, `_externalSpanCountsByTest/ByTrace`, `_testCaseSpanIds`) to `OrdinalIgnoreCase` — `_knownTraceIds` was already that way. Added a regression test `IngestExternalSpan_TraceIdCaseMismatch_StillCorrelates` (register lowercase, ingest uppercase, verify correlation). Docs and existing tests no longer ask callers to `.ToUpperInvariant()`.

**`SpanType = span.Name`** — You're right, it's redundant. `SpanType` is the TUnit-only classifier (`test_case`, `test_suite`, `test_assembly`) used by the HTML template to apply category styling to first-party spans. No OTLP analogue exists, so external spans now set `SpanType = null`.

thomhurst:
All three addressed in cd5c6b0a3.

- **Silent catch** — `ProcessTraces` *and* `ProcessLogs` now emit `Trace.WriteLine` with exception type + message, matching the convention used elsewhere in `OtlpReceiver`.
- **`[EditorBrowsable]` on `internal`** — dropped (and the now-unused `System.ComponentModel` using). You're right, it's a no-op.
- **Hardcoded cap** — `ActivityCollector.MaxExternalSpans` is now `internal` (IVT already covers `TUnit.OpenTelemetry.Tests`), and the test reads from it so CI-overridden caps don't produce a confusing mismatch.

Issue #5528, opened 2026-04-12
perf: runtime engine optimizations - eliminate per-test allocation waste and redundant synchronization


thomhurst:
## Summary

Performance analysis of the TUnit runtime engine identified several substantial optimization opportunities. These target the test execution hot path — code that runs once per test — and discovery/init paths. The focus is on eliminating redundant work, unnecessary allocations, and architectural overhead rather than micro-optimizations.

#### 1. Duplicate execution deduplication (2 redundant TCS dictionaries)

**Files:** `TUnit.Engine/Scheduling/TestRunner.cs:42`, `TUnit.Engine/Services/TestExecution/TestExecutionGuard.cs:11`

Two independent `ConcurrentDictionary<string, TaskCompletionSource<bool>>` implement the exact same deduplication pattern. Every test allocates **two** TCS objects, performs **two** dictionary lookups, and maintains **two** dictionary entries.

- `TestRunner.ExecuteTestAsync` — TCS + ConcurrentDictionary guard for DependsOn recursion
- `TestExecutionGuard.TryStartExecutionAsync` — identical TCS + ConcurrentDictionary guard inside TestCoordinator

For 10,000 tests: 20,000 TCS allocations and 40,000 dictionary operations where half are redundant.

**Fix:** Consolidate to a single deduplication layer. `TestRunner` already owns the entry point — remove `TestExecutionGuard` and call `ExecuteTestInternalAsync` directly from `TestCoordinator.ExecuteTestAsync`.

---

#### 2. Per-test CancellationToken.Register accumulation (~99% redundant)

**File:** `TUnit.Engine/Services/AfterHookPairTracker.cs:63-96`

`RegisterAfterAssemblyHook` and `RegisterAfterClassHook` are called for **every single test** but only need to register once per assembly/class. For 10,000 tests across 200 classes and 5 assemblies: 20,000 `CancellationTokenRegistration` heap allocations in an unbounded `ConcurrentBag`, when only ~205 are needed.

The session-level hook already has a guard (`_sessionHookRegistered` at line 26). Assembly and class hooks don't.

**Fix:** Add `ConcurrentHashSet<Assembly>` and `ConcurrentHashSet<Type>` guards:

```csharp
private readonly ConcurrentHashSet<Assembly> _assemblyHookRegistered = new();

public void RegisterAfterAssemblyHook(Assembly assembly, ...)
{
    if (!_assemblyHookRegistered.Add(assembly)) return;
    // existing registration logic
}
```

[...]

PR #5611, opened 2026-04-17
perf: engine discovery/init path cleanup (#5528 C)

thomhurst:
Part C of three-PR split for issue #5528.

## Scope
Items **7-10** from the perf checklist — discovery/init path cleanup.

### Item 7 — TestGroupingService sync hot path
`GroupTestsByConstraintsAsync` previously built an async state machine per call even though every `await` was behind an `IsTraceEnabled` guard. The core grouping work is now synchronous; trace messages are collected into a list and flushed through a thin `async` wrapper only when tracing is enabled. Hot path: zero state-machine overhead, no list allocation for trace messages.

### Item 8 — LINQ chains → direct loops
- `TestSessionCoordinator.InitializeEventReceivers` — dropped `Select(t => t.Context)` and passed the test list through directly.
- `EventReceiverOrchestrator.InitializeTestCounts` — accepts `IReadOnlyList<AbstractExecutableTest>` directly; no more `allTestContexts as IList<TestContext> ?? ToList()` fallback. Internal-only API, no public surface change.
- `EventReceiverOrchestrator.InvokeHookRegistrationEventReceiversAsync` — replaced `OfType<T>().OrderBy(...)` with an explicit filtering loop + in-place `List<T>.Sort`.
- `HookDelegateBuilder` — three copies of `.OrderBy().ThenBy().Select().ToList()` consolidated into a single `SortAndProject<TContext>` helper using in-place `List.Sort` + index loop.
- `TestDiscoveryContext.Assemblies` / `.TestClasses` — `Select(...).Distinct().ToArray()` replaced with inlined `HashSet<T>.Add` loops + `CopyTo`.
- `TestDiscoveryService.GetCachedTestContexts` — materializes a `List<TestContext>` from the cached queue instead of returning a `Select` enumerable that allocates a lambda-backed iterator per foreach.

### Item 9 — `TestSessionContext.HasFailures` via atomic counter
Added `public bool HasFailures` backed by a lock-free `int _failureCount` incremented via `Interlocked.Increment` in `TestCoordinator`'s terminal switch for `Failed`/`Timeout`/`Cancelled` states. `HookExecutor.ExecuteAfterTestSessionHooksAsync` now reads the flag in O(1) instead of materializing `AllTests` (which itself eagerly flattens `SelectMany` across every class) and LINQ-scanning it.

### Item 10 — `TestContext.Id` lazy string
`TestContext` now stores a `Guid` field and materializes the string form lazily via `field ??= _idGuid.ToString()`. The internal registry is keyed by `Guid` (`ConcurrentDictionary<Guid, TestContext>`); `GetById(string)` parses once on the lookup path. `TestCoordinator` cleanup calls the new `RemoveFromRegistry()` instance method so a test's string Id is never materialized unless something external asks for it. Public surface unchanged.

## Verification
- Build clean across all TFMs.
- TUnit.Engine.Tests (net10.0): 148/148 passed, 99 skipped (platform/CI-only).
- TUnit.UnitTests (net10.0): 180/180 passed.
- TUnit.PublicAPI (net10.0): 4/4 passed. One expected snapshot update (`HasFailures` property added to `TestSessionContext`) committed in `Tests.Core_Library_Has_No_API_Changes.DotNet10_0.verified.txt`.

## Related PRs
- #5610 — Part B (hot-path allocation wins).
- #5612 — Part A (consolidate dedup + hook register guards).

### DIFF
--- a/TUnit.Core/Models/TestDiscoveryContext.cs
+++ b/TUnit.Core/Models/TestDiscoveryContext.cs
@@ -35,7 +35,8 @@ public class TestDiscoveryContext : Context
 
     private AssemblyHookContext[] BuildUniqueAssemblies()
     {
-        // List+HashSet preserves first-occurrence insertion order (matches prior LINQ Distinct() contract).
+        // HashSet<T> iteration order is not contractually guaranteed; the parallel list
+        // preserves first-occurrence order so hook execution sequence stays deterministic.
         var seen = new HashSet<AssemblyHookContext>();
         var ordered = new List<AssemblyHookContext>();
         foreach (var cls in TestClasses)
--- a/TUnit.Core/Models/TestDiscoveryContext.cs
+++ b/TUnit.Core/Models/TestDiscoveryContext.cs
@@ -28,10 +28,38 @@ public class TestDiscoveryContext : Context
     public required string? TestFilter { get; init; }
 
     [field: AllowNull, MaybeNull]
-    public IEnumerable<AssemblyHookContext> Assemblies => field ??= TestClasses.Select(x => x.AssemblyContext).Distinct().ToArray();
+    public IEnumerable<AssemblyHookContext> Assemblies => field ??= BuildUniqueAssemblies();
 
     [field: AllowNull, MaybeNull]
-    public IEnumerable<ClassHookContext> TestClasses => field ??= AllTests.Where(x => x.ClassContext != null).Select(x => x.ClassContext!).Distinct().ToArray();
+    public IEnumerable<ClassHookContext> TestClasses => field ??= BuildUniqueTestClasses();
+
+    private AssemblyHookContext[] BuildUniqueAssemblies()
+    {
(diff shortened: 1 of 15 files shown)


thomhurst:
Thanks for the follow-up review. Addressing both points:

**Re: `HasFailures` + retries** — not an issue. `RetryHelper.ExecuteWithRetry` catches intermediate exceptions inside its own loop and only rethrows on final failure. The `MarkFailure()` call in `TestCoordinator` lives in the post-lifecycle `switch (test.State)` block, which runs once per test after all retry attempts complete. So `test.State` reflects the final outcome and `MarkFailure()` is never called for intermediate retry failures. Semantics match the original `AllTests.Any(...)` scan exactly.

**Re: `TestContext.Id` lazy materialization** — fair point. Switched to `Interlocked.CompareExchange` so we publish a single string even if the Id format ever becomes non-deterministic. Removes the documentation burden too.

PR #5652, opened 2026-04-21
fix(asp-net): forward disposal in FlowSuppressingHostedService (#5651)

### LINKED ISSUE #5651: [Bug]: FlowSuppressingHostedService does not forward IDisposable / IAsyncDisposable to inner service
### Description

TUnit.AspNetCore.TestWebApplicationFactory<T> wraps every registered IHostedService in a FlowSuppressingHostedService to run StartAsync on a clean ExecutionContext. The wrapper implements IHostedLifecycleService and forwards its methods, but it does not implement IDisposable or IAsyncDisposable. When the DI container disposes the host, it only sees the non-disposable wrapper, so the inner service's Dispose / DisposeAsync is never called.

For hosted services that own unmanaged resources (gRPC channels, database connections, process handles, file streams), this is a silent resource leak.

Introduced in commit 76c4d4ed0 / PR #5589 / #5598 (merged 2026-04-17) which added FlowSuppressingHostedService to solve a span-correlation issue during StartAsync. The disposal forwarding was not part of that change.

The sibling class TracedWebApplicationFactory in the same package already implements the correct dual-interface forwarding pattern, so this is a gap, not a novel design question.

### Expected Behavior

When the DI container disposes the host (via either sync Dispose or async DisposeAsync on the WebApplicationFactory), the inner IHostedService's Dispose / DisposeAsync should be called so any owned resources are released.

### Actual Behavior

The wrapper does not implement IDisposable or IAsyncDisposable, so the DI container skips it when iterating disposables. The inner service's Dispose / DisposeAsync is never called. Resources leak silently.

### Steps to Reproduce

Construct the wrapper directly and check whether disposal is forwarded. This repro does not require a WebApplicationFactory or a host:

```csharp
using Microsoft.Extensions.Hosting;
using TUnit.AspNetCore;

public class DisposalRepro
{
    [Test]
    public async Task Inner_DisposeAsync_Is_Called_By_Wrapper()
    {
        var inner = new AsyncProbe();
        var wrapper = new FlowSuppressingHostedService(inner);

        await ((IAsyncDisposable)wrapper).DisposeAsync();

        await Assert.That(inner.DisposeAsyncCalled).IsTrue();
        // FAILS on 1.37.10: cast to IAsyncDisposable is invalid at runtime
        // because the wrapper does not declare the interface.
    }
}

internal sealed class AsyncProbe : IHostedService, IAsyncDisposable
{
    public bool DisposeAsyncCalled { get; private set; }
    public Task StartAsync(CancellationToken ct) => Task.CompletedTask;
    public Task StopAsync(CancellationToken ct) => Task.CompletedTask;
    public ValueTask DisposeAsync() { DisposeAsyncCalled = true; return ValueTask.CompletedTask; }
}

### TUnit Version

1.37.10

### .NET Version

.NET 10.0 (also affects net8.0 and net9.0 targets)

### Operating System

Other (please specify in additional context)

### IDE / Test Runner

```shell

```

### Additional Context

- AOT / trimming: not AOT-specific. Reproduces identically under JIT and Native AOT publish.
- Custom configuration: none. Standard TestWebApplicationFactory usage with the default SuppressHostedServiceExecutionContextFlow = true.
- Specific test data: none. Reproduces with any IHostedService that implements IDisposable or IAsyncDisposable.
- Target framework coverage: affects all published TFMs of TUnit.AspNetCore.Core (net8.0, net9.0, net10.0).
- Workaround: override SuppressHostedServiceExecutionContextFlow => false on the factory subclass. Restores disposal but disables flow suppression (trading one desirable behaviour for another).
- Have a patch with regression tests on a local branch; PR to follow.

### IDE-Specific Issue?

JohnVerheij:
## Description

FlowSuppressingHostedService implements IHostedLifecycleService but did not
implement IDisposable or IAsyncDisposable. When the DI container disposed the
host, the container only saw the non-disposable wrapper, so the inner service's
Dispose / DisposeAsync was never called, leaking unmanaged resources silently.

Add both IDisposable and IAsyncDisposable to the wrapper, forwarding to the
inner service. DisposeAsync prefers the async path on inner and falls back to
sync Dispose when inner is IDisposable only. Dispose forwards only when inner
implements IDisposable; async-only inner services are not disposed on the
synchronous path, matching the "never block on async" rule in CLAUDE.md.
Callers with async-only resources should use DisposeAsync (or await using on
the owning factory).

Follows the dual-interface forwarding precedent already in TracedWebApplicationFactory. That class can call `_inner.Dispose()`/`DisposeAsync()` directly because its inner type always implements both interfaces; this wrapper cannot make that assumption about IHostedService implementations, so it uses an is-check.

## Related Issue

Fixes #5651

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

### Required

- [ ] **Dual-Mode Implementation**: N/A. The fix is in `TUnit.AspNetCore.Core`, not `TUnit.Core.SourceGenerator` or `TUnit.Engine`. Disposal forwarding is not test-discovery/execution metadata.
- [ ] **Snapshot Tests**: N/A. No source generator output or public API change.
- [x] **Performance**: change is disposal-time branching. Zero new allocations, no LINQ, no reflection. Not a hot path.
- [x] **AOT Compatibility**: reflection-free. Verified under `PublishAot=true`.

### Testing

- [x] All existing tests pass
- [ ] I have tested both source-generated and reflection modes (N/A — not a dual-mode concern)

## Additional Notes

Regression tests: 7 pure-unit-tests in `HostedServiceDisposalForwardingTests.cs` covering:

| Test | Scenario |
|---|---|
| `DisposeAsync_Forwards_To_IAsyncDisposable_Inner` | Async path on async-only inner |
| `Dispose_Forwards_To_IDisposable_Inner` | Sync path on sync-only inner |
| `DisposeAsync_On_SyncOnly_Inner_Falls_Back_To_Dispose` | Async wrapper method, sync inner |
| `Dispose_On_AsyncOnly_Inner_Is_No_Op` | Sync wrapper, async-only inner — documented no-op |
| `DisposeAsync_On_DualInterface_Inner_Prefers_Async_Path` | Preference ordering |
| `Dispose_On_DualInterface_Inner_Prefers_Sync_Path` | Preference ordering |
| `Wrapper_Does_Not_Throw_When_Inner_Is_Not_Disposable` | Non-disposable inner is safe |

Tests construct the wrapper directly (no `TestWebApplicationFactory`, no host) to isolate the forwarding contract from `TestWebApplicationFactory`'s disposal-ordering behaviour, which is outside this fix's scope. Verified stable over 10 consecutive runs across net8.0, net9.0, and net10.0 (70/70 assertion successes per run, 210/210 across all three TFMs).

The "never block on async" rule in `CLAUDE.md` is the reason the sync `Dispose` path does not attempt an async-to-sync bridge for async-only inners. `Microsoft.Extensions.Hosting.Host.Dispose` internally uses `DisposeAsync().AsTask().GetAwaiter().GetResult()`, but that pattern is deliberately avoided here per the repo's stated convention — the documented no-op is the tradeoff.

### DIFF
--- a/TUnit.AspNetCore.Core/FlowSuppressingHostedService.cs
+++ b/TUnit.AspNetCore.Core/FlowSuppressingHostedService.cs
@@ -11,12 +11,20 @@ namespace TUnit.AspNetCore;
 /// as their parent.
 /// </summary>
 /// <remarks>
+/// <para>
 /// Implements <see cref="IHostedLifecycleService"/> so the Host's lifecycle hooks keep
 /// firing for inner services that implement it — the Host uses an <c>is</c> check
 /// against the registered instance, so without passthrough wrapping would silently
 /// drop those hooks.
+/// </para>
+/// <para>
+/// Also implements <see cref="IAsyncDisposable"/> and <see cref="IDisposable"/> so the
+/// DI container forwards disposal to the inner service when the host is disposed.
+/// Without this, wrapped services that own unmanaged resources leak silently because
+/// the container only sees the non-disposable wrapper.
+/// </para>
 /// </remarks>
-internal sealed class FlowSuppressingHostedService(IHostedService inner) : IHostedLifecycleService
+internal sealed class FlowSuppressingHostedService(IHostedService inner) : IHostedLifecycleService, IAsyncDisposable, IDisposable
 {
     public Task StartAsync(CancellationToken cancellationToken) =>
         RunOnCleanContext(inner.StartAsync, cancellationToken);
@@ -46,6 +54,34 @@ internal sealed class FlowSuppressingHostedService(IHostedService inner) : IHost
         inner is IHostedLifecycleService lifecycle
             ? lifecycle.StoppedAsync(cancellationToken)
             : Task.CompletedTask;
+
+    /// <inheritdoc />
+    public async ValueTask DisposeAsync()
+    {
+        if (inner is IAsyncDisposable asyncDisposable)
+        {
+            await asyncDisposable.DisposeAsync().ConfigureAwait(false);
+        }
+        else if (inner is IDisposable disposable)
+        {
+            disposable.Dispose();
+        }
+    }
+
+    /// <inheritdoc />
+    /// <remarks>
+    /// Forwards to the inner service's <see cref="IDisposable.Dispose"/> when implemented.
+    /// Inner services that only implement <see cref="IAsyncDisposable"/> are not disposed on
+    /// this synchronous path. Callers should use <see cref="DisposeAsync"/> (or
+    /// <c>await using</c> on the owning factory) to release async-only resources.
+    /// </remarks>
+    public void Dispose()
+    {
+        if (inner is IDisposable disposable)
+        {
+            disposable.Dispose();
+        }
+    }
 
     // Dispatch onto a thread-pool worker with a clean captured ExecutionContext by
     // combining SuppressFlow + Task.Run. Unlike wrapping `using (SuppressFlow()) return op(ct);`
--- a/TUnit.AspNetCore.Tests/HostedServiceDisposalForwardingTests.cs
+++ b/TUnit.AspNetCore.Tests/HostedServiceDisposalForwardingTests.cs
@@ -1,0 +1,145 @@
+using Microsoft.Extensions.Hosting;
+using TUnit.AspNetCore;
+
+namespace TUnit.AspNetCore.Tests;
+
+/// <summary>
+/// Tests that <see cref="FlowSuppressingHostedService"/> forwards <see cref="IDisposable.Dispose"/>
+/// and <see cref="IAsyncDisposable.DisposeAsync"/> calls to the wrapped inner service.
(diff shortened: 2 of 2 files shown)


JohnVerheij:
Thanks for the careful review.

**Obs 1 (indirect assertion):** addressed in the force-push to `dc08af34a`. Replaced the `StartedCount == 0` check with explicit `ThrowsNothing()` assertions on the disposal calls and dropped the now-unused `StartedCount` tracking from the probe. Test name and test body now match semantically, and failure messages will point at the specific disposal call that threw.

**Obs 2 (double-disposal guard):** keeping the current shape to stay aligned with `TracedWebApplicationFactory`, which also omits the idempotency guard. The PR description cites that as precedent. If @thomhurst prefers the explicit `Interlocked.Exchange` guard, happy to add it here and push a matching update to `TracedWebApplicationFactory` in a follow-up. Otherwise staying consistent seems right.

thomhurst:
Thanks @JohnVerheij ! Nice catch

Issue #4818, opened 2026-02-15
feat: Per-test correlated logging for TUnit.Aspire via hosting startup injection


thomhurst:
## Context

Raised during the design of `TUnit.Aspire` (#4768). Aspire resources run out-of-process, so per-test log correlation (like `TUnit.AspNetCore` achieves via `IHttpContextAccessor` middleware) isn't straightforward.

## Current State (v1)

`TUnit.Aspire` will ship with `WatchResourceLogs("resourceName")` — an opt-in method that subscribes to ALL logs from a named resource and routes them to the subscribing test's output. This works for all resource types but doesn't correlate logs to the specific test that caused them.

## Proposed Enhancement

Use ASP.NET Core's `IHostingStartup` mechanism to inject correlation middleware into project resources without modifying user code:

1. **Ship a `TUnit.Aspire.HostingStartup` assembly** containing middleware that reads a correlation header (`X-TUnit-TestId`) and attaches it to the logging scope
2. **Auto-configure project resources** via `WithEnvironment("ASPNETCORE_HOSTINGSTARTUPASSEMBLIES", "TUnit.Aspire.HostingStartup")`
3. **Wrap `CreateHttpClient`** with a `DelegatingHandler` that injects the test ID header from `TestContext.Current`
4. **Correlated logger** reads the test ID from the log scope and routes to the correct test's output

## Challenges

- **Only works for ASP.NET Core project resources** — containers (Postgres, Redis, RabbitMQ) and executables can't have middleware injected
- **Assembly availability** — the hosting startup assembly must be loadable by the resource's process, which has packaging implications since Aspire launches projects from their own build output
- **Complexity** — requires a separate assembly, environment variable wiring, a delegating handler, middleware, and a correlated logger

## Alternatives Considered

- Moving the AsyncLocal-based correlated logger from `TUnit.AspNetCore` to `TUnit.Logging.Microsoft` as a shared primitive — helps with host-level logs but doesn't solve resource-level correlation
- Reverse proxy / sidecar pattern — too complex

## Related

- #4768 — TUnit.Aspire integration package
- `TUnit.AspNetCore` correlated logging (`CorrelatedTUnitLoggerProvider`, `TUnitTestContextMiddleware`) as reference implementation

thomhurst:
@claude Any alternative designs for this?

redoz:
[...]

PR #5666, opened 2026-04-22
feat(aspire): emit client spans for HTTP

thomhurst:
## Summary

Follow-up to #5665 focused on trace topology rather than OTLP forwarding.

- create a real `TUnit.Aspire.Http` client span around `AspireFixture.CreateHttpClient` requests
- propagate `traceparent` and `baggage` from that client span, with fallback to the ambient activity when no listener is attached
- export the new source from `TUnit.OpenTelemetry.AutoStart`
- update `BaggagePropagationHandlerTests` to assert client span creation, parent-child shape, response tagging, and fallback behavior

## Why

#5665 fixes forwarding to user dashboards, but the Aspire handler still propagated the ambient test span directly. That kept correlation working while flattening the distributed trace tree in backends like the Aspire dashboard.

This change restores the normal OTel client/server shape:
`test case` -> `test body` -> `TUnit.Aspire.Http` client span -> SUT server span

Refs #4818

## Test plan

- [x] `dotnet test TUnit.Aspire.Tests/TUnit.Aspire.Tests.csproj -c Release -- --treenode-filter "/*/*/BaggagePropagationHandlerTests/*"`
- [x] `dotnet test TUnit.OpenTelemetry.Tests/TUnit.OpenTelemetry.Tests.csproj -c Release -- --treenode-filter "/*/*/AutoStartTests/*"`
- [x] `dotnet build TUnit.Aspire.Tests/TUnit.Aspire.Tests.csproj -c Release --no-restore`
- [x] `dotnet build TUnit.OpenTelemetry.Tests/TUnit.OpenTelemetry.Tests.csproj -c Release --no-restore`

### DIFF
--- a/TUnit.AspNetCore.Core/Http/ActivityPropagationHandler.cs
+++ b/TUnit.AspNetCore.Core/Http/ActivityPropagationHandler.cs
@@ -52,14 +52,25 @@ internal sealed class ActivityPropagationHandler : DelegatingHandler
         InjectTraceContext(propagationActivity, request.Headers);
         InjectBaggage(propagationActivity, request.Headers);
 
-        var response = await base.SendAsync(request, cancellationToken);
+        HttpResponseMessage response;
+        try
+        {
+            response = await base.SendAsync(request, cancellationToken).ConfigureAwait(false);
+        }
+        catch (Exception ex)
+        {
+            RecordException(activity, ex);
+            throw;
+        }
 
         if (activity is not null)
         {
-            activity.SetTag("http.response.status_code", (int)response.StatusCode);
-            if (!response.IsSuccessStatusCode)
+            var statusCode = (int)response.StatusCode;
+            activity.SetTag("http.response.status_code", statusCode);
+            if (statusCode >= 400)
             {
                 activity.SetStatus(ActivityStatusCode.Error);
+                activity.SetTag("error.type", statusCode.ToString());
             }
         }
 
@@ -126,5 +137,24 @@ internal sealed class ActivityPropagationHandler : DelegatingHandler
         {
             headers.TryAddWithoutValidation(TUnit.Core.TUnitActivitySource.BaggageHeader, baggage);
         }
+    }
+
+    private static void RecordException(Activity? activity, Exception exception)
+    {
+        if (activity is null)
+        {
+            return;
+        }
+
+        var tags = new ActivityTagsCollection
+        {
+            { "exception.type", exception.GetType().FullName },
+            { "exception.message", exception.Message },
+            { "exception.stacktrace", exception.ToString() }
+        };
+
+        activity.AddEvent(new ActivityEvent("exception", tags: tags));
+        activity.SetTag("error.type", exception.GetType().FullName);
+        activity.SetStatus(ActivityStatusCode.Error, exception.Message);
     }
 }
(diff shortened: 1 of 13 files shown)


thomhurst:
Addressed the review with a follow-up push.

Changes made:
- record thrown transport exceptions on synthesized HTTP client spans in both handlers
- add explicit coverage for thrown inner-handler failures, `traceparent` format, and multi-item baggage
- align HTTP status handling in both handlers to the current OTel HTTP semconv: 3xx leaves status unset, 4xx/5xx set `Error` with `error.type`

On the 4xx point specifically: the current OpenTelemetry HTTP span spec says 4xx SHOULD be `Error` for `SpanKind.CLIENT`, while 1xx/2xx/3xx stay unset unless there was another error:
https://opentelemetry.io/docs/specs/semconv/http/http-spans/

thomhurst:
Folded in the cleanup suggestions in 4ae5f04:\n\n- reused TUnitActivitySource.RecordException(...) from both HTTP handlers instead of keeping three near-identical copies\n- added a FullName ?? Name fallback for the recorded exception type\n- updated the AspireFixture.CreateHttpClient comment to call out that the shared handler creates/disposes a per-request client span\n\nI left CopyBaggage as-is for now since that felt like a lower-priority refactor than the exception path cleanup.

thomhurst:
Took the small cleanup in 6631e07:\n\n- clarified in both handlers that exception recording only applies when a synthesized client span exists\n- tightened SendAsync_MultipleBaggageItems_CommaSeparated so the listener scope is now asserted instead of incidental\n\nOn the 4xx point, I re-checked the current OpenTelemetry HTTP span status guidance and it still says 4xx should remain unset for server spans but should be marked Error for client spans when the instrumentation has no extra request context, so I’m keeping the >= 400 client behavior in both handlers:\nhttps://opentelemetry.io/docs/specs/semconv/http/http-spans/#status

thomhurst:
Took the span-name cleanup in b188f81.\n\nI ended up going a little further than just dropping the HTTP  prefix: both handlers now emit method-only client span names (GET, POST, etc.), and both test projects assert that explicitly.\n\nReason: the current HTTP semconv says span names should be {method} {target} only when a low-cardinality target is available, otherwise just {method}, and instrumentation must not default the target to the raw URI path. Since these helpers don't have a route template/url template, method-only is the closest spec-aligned name here:\nhttps://opentelemetry.io/docs/specs/semconv/http/http-spans/#name

thomhurst:
Addressed the latest review in e8d2d10:

- restored explicit W3C traceparent format assertions in both HTTP handler test suites
- expanded the AspNetCore recorded-span helper so it now asserts the same trace, parent, and kind invariants as the Aspire tests
- clarified the Aspire baggage-copy comment to match the cross-framework rationale already called out in the sibling handler, plus a short note that we only skip baggage already attached to the synthesized span

Verification passed with:
- dotnet test TUnit.Aspire.Tests/TUnit.Aspire.Tests.csproj -c Release -- --treenode-filter "/*/*/BaggagePropagationHandlerTests/*"
- dotnet test TUnit.AspNetCore.Tests/TUnit.AspNetCore.Tests.csproj -f net10.0 -c Release -- --treenode-filter "/*/*/ActivityPropagationHandlerTests/*"

PR #5668, opened 2026-04-22
fix(telemetry): remove duplicate HTTP client spans

thomhurst:
## Summary

- HTTP propagation handlers (`TUnitBaggagePropagationHandler` for Aspire, `ActivityPropagationHandler` for ASP.NET Core) are now pure header propagators — no synthesized client span. Both delegate to a shared `HttpActivityPropagator.Inject(Activity?, HttpRequestHeaders)` in `TUnit.Core`.
- `TUnitTestCorrelationProcessor` tags the `tunit.test.id` on both `OnStart` and `OnEnd` so ASP.NET Core server spans (remote-context parent, baggage populated by the propagator *after* `StartActivity` returns) still get tagged.
- `AspireHttpSourceName` (unreleased, added in #5666) removed; `AspNetCoreHttpSourceName` marked `[Obsolete]` — kept for binary compatibility since it shipped in v1.36.0.

## Context

Originally reported via a trace timeline showing two identical `POST` spans per HTTP call under Aspire:

```
POST 324ms              ← TUnit.Aspire.Http   (synthesized in #5666)
  POST 324ms            ← System.Net.Http     (runtime, always there)
    HTTP wait_for_connection
  POST /api/products/   ← ASP.NET Core server
```

`AspireFixture.CreateHttpClient` wraps a real `SocketsHttpHandler`, so the runtime already emits a client span via `System.Net.Http`. The synthesized span was a second one.

The same cleanup applies to the ASP.NET Core side even though it didn't have a visible duplicate: `ActivityPropagationHandler` is also injected into every SUT-side `IHttpClientFactory` pipeline via `TUnitHttpClientFilter`, and that path *does* end in `SocketsHttpHandler` — so a SUT with HTTP instrumentation would also have seen duplicates there.

Could we have reused the runtime's `DiagnosticsHandler`? No — it's `internal sealed`, can't instantiate without reflection (violates AOT rule), and its tag shape drifts across .NET versions. Since the WAF in-memory case has no client span at all (neither runtime nor ours), the cleanest answer is to let the ASP.NET Core server span be a direct child of the ambient test Activity. Trace topology becomes:

```
test body
  POST /api/products/  ← ASP.NET Core server span, parent = test activity
```

## Trace-correlation tag fix

Moving tagging to OnEnd fixes a latent issue: the processor previously read `Activity.Current.GetBaggageItem(TagTestId)` at OnStart, but ASP.NET Core Hosting populates baggage via `Activity.AddBaggage(...)` *after* `ActivitySource.StartActivity` returns. The synthesized client span used to mask this by being tagged itself and passing the test's "some span has the tag" assertion. With the synthesized span gone, the server span must be the tagged one — hence OnEnd.

## Test plan

- [x] `TUnit.AspNetCore.Tests` — 35/35 pass (including `AutoWires_TagsAspNetCoreSpans_WithTestId` which now polls briefly for server-activity end)
- [x] `TUnit.Aspire.Tests/BaggagePropagationHandlerTests` — 10/10 pass (rewritten to assert on outgoing headers rather than synthesized spans)
- [x] `TUnit.OpenTelemetry.Tests` — 29/29 pass (new `Processor_TagsOnEnd_WhenBaggageAddedAfterStart` covers the new path)
- [x] `TUnit.PublicAPI` snapshots — 4/4 pass (reflects removed const + new `OnEnd` override + `[Obsolete]` attribute)
- [ ] Verify in a real Aspire dashboard trace that the duplicate `POST` rows disappear


thomhurst:
Thanks for the review — addressed in a516d9e78.

**`TryTag` fallback concurrency concern:** Good catch. Switched the fallback from `Activity.Current?.GetBaggageItem` to `TraceRegistry.GetContextId(activity.TraceId.ToString())`. `TraceRegistry` is already populated in `TestExecutor.cs:155` with `traceId → contextId` when each test activity starts, so the lookup is bound to the span's own trace ID rather than whatever ambient context happens to be current at OnEnd. That eliminates the cross-attribution risk under concurrent tests when a span is stopped on a thread whose `Activity.Current` has swung to another test.

Added `Processor_FallsBackToTraceRegistry_WhenActivityHasNoBaggage` to cover the new path.

**Polling loop `CancellationToken` race:** Switched to a monotonic `Environment.TickCount64` deadline so `Task.Delay` no longer takes the token. A timeout now produces the clean "no tagged span found" assertion failure rather than a `TaskCanceledException`.

**`AspireHttpSourceName` release history:** Confirmed via `git log -S`. Added in commit `ecb01f89b` (the PR feature branch for #5666) and merged to main as `572118965` — no intermediate release tag. First appearance in the public-API verified snapshots is also only in that commit. Safe to hard-remove.

thomhurst:
Thanks for the follow-up review. Picked up the import-style nit in da533fb4d — `System.Diagnostics` is now imported in `ActivityPropagationHandler.cs` to match the Aspire handler.

Skipping the other two observations as non-blocking:

- **Empty `baggage` header edge case**: no propagator in the BCL (W3C, Legacy, Pass-through) emits an empty `baggage` header — W3C simply omits it when there's nothing to emit. Guarding against a hypothetical custom propagator that emits `baggage:` (empty) would add complexity for a contract violation on the propagator's side. Skipped.
- **`HttpActivityPropagator` `#if NET` gating**: consistent with `TUnitActivitySource` and the handlers themselves; netstandard2.0 doesn't carry the `DistributedContextPropagator` API anyway.
- **`TraceRegistry.Register` leak in test**: reviewer already flagged as acceptable given process-global design.

thomhurst:
Addressed the two actionable observations in 84b195ef2:

1. **`HttpActivityPropagator.Inject` method visibility** → `internal` to match the enclosing class and the rest of TUnit's internal helper style.
2. **OnEnd mutation visibility reasoning** → added a remarks paragraph to `TUnitTestCorrelationProcessor` noting that tag writes during OnEnd are visible to deferred-serialization exporters (BatchExportProcessor, InMemoryExporter) and that synchronous exporters need this processor registered before them.

Not taking:
- **Polling loop exporter-flush caveat**: the test uses `AddInMemoryExporter` which captures by reference, so no flush-window fragility. The comment already captures *why* polling is needed.
- **`AspireHttpSourceName` release confirmation**: confirmed earlier via `git log -S "AspireHttpSourceName"` — only commits are `ecb01f89b` (feature branch) and `572118965` (merge to main, unreleased). Also verified first appearance in public-API snapshots is only on those commits.
- **Exception-annotation behavior changelog note**: the PR body already states "span synthesis is removed" which implicitly removes the span-level error annotations. The errors are still recorded on the runtime's `System.Net.Http` client span (Aspire) and ASP.NET Core server span (WAF) where they naturally belong — no one should be filtering on TUnit-synthesized spans specifically.

thomhurst:
Both review rounds clarified below.

**SimpleExportProcessor ordering note added** (e39814e46) — `docs/docs/examples/opentelemetry.md` now calls out the ordering requirement for synchronous exporters. `BatchExportProcessor` (the default for OTLP/Jaeger/Zipkin) defers serialization so order doesn't matter there; `Simple`-wrapped exporters need TUnit's processor first.

**On the Aspire propagation concern (review #5, item 1):** empirically verified that correlation still works without the `System.Net.Http` subscriber — I temporarily removed `AddSource(\"System.Net.Http\")` from \`AutoStart.cs\` and reran the Aspire trace-demo test. Result: SUT server spans share the test activity's \`TraceId\`, baggage (including \`tunit.test.id\`) is still extracted on the SUT side, and OTLP log correlation still routes to the right test. The only thing missing without the subscriber is the visual client span in the dashboard waterfall — the server span ends up with an orphan-parent reference but is otherwise correctly correlated.

Runtime rationale: .NET 8+'s \`DiagnosticsHandler.SendAsyncCore\` injects \`traceparent\`/\`baggage\` based on \`Activity.Current\` as long as one exists, regardless of whether \`System.Net.Http\` has a listener. The listener only gates whether the client span itself is recorded/exported. This was symmetric pre-fix: the old handler also required \`Activity.Current\` to be non-null (guarded via early return in \`InjectTraceContext\`/\`InjectBaggage\`). So there's no correlation regression for users with a test activity active — which is the only case where the scenario is meaningful.

**On exception annotation removal (reviews #4 and #5, item 2):** the WAF in-memory path's exception surface area is narrow — WAF doesn't have a real transport layer, so \`HttpRequestException\` thrown by the inner handler is essentially limited to a host crashing mid-response, which is vanishingly rare in integration tests. Exceptions that propagate from user middleware are captured on the ASP.NET Core server span by that pipeline's own instrumentation. I think this is not worth a dedicated CHANGELOG entry — the PR body already states \"span synthesis removed\" which implicitly removes all span-level annotations including error recording. Happy to add a note if you disagree.

thomhurst:
Applied two of the four items in e7ca305be:

- **Ordering note near custom pipeline example** — `docs/docs/examples/opentelemetry.md` now carries the \`SimpleExportProcessor\` ordering caveat next to the custom pipeline code sample (Option B), in addition to the correlation-processor section later on.
- **Polling window 2s → 5s** — zero cost on success, headroom for slower CI runners.

Skipping the other two:

- **Silent baggage drop when caller pre-sets \`baggage\` header**: the behavior is intentional parity with \`DistributedContextPropagator.Current.Inject\`, which also skips when the header is already set. The \`SendAsync_ExistingBaggageHeader_IsPreserved\` test pins this contract — a caller who explicitly sets a \`baggage\` header takes responsibility for its content. A \`string.Contains(TagTestId)\` heuristic would false-positive on keys like \`my.tunit.test.id.custom=...\`, and proper merge requires parsing the multi-valued W3C baggage format. Not worth the complexity for a niche scenario (caller manually pre-sets baggage without including \`tunit.test.id\` while expecting TUnit correlation to still work).

- **\`System.Net.Http\` OTLP feedback loop**: empirically verified against Jaeger — the Aspire trace-demo demo produced 11 clean spans, not a runaway. The OpenTelemetry SDK wraps export HTTP calls in \`SuppressInstrumentationScope.Begin()\`, and \`TracerProviderSdk\`'s sampler returns \`None\` when \`Sdk.SuppressInstrumentation\` is set. Adding a comment here to explain why this *isn't* a problem felt backwards.

thomhurst:
Thanks for the additional pass. Responding to each:

**1. Aspire propagation without \`TUnit.OpenTelemetry\` — the premise is incorrect**

> The runtime's DiagnosticsHandler only injects headers when an ActivityListener is subscribed to \`System.Net.Http\`.

Empirically verified that this isn't the case. I already ran this experiment earlier in the conversation: removed \`AddSource(\"System.Net.Http\")\` from \`AutoStart.cs\`, reran the Aspire trace-demo test. Result: SUT server spans still shared the test activity's \`TraceId\`, baggage (including \`tunit.test.id\`) still propagated, OTLP log correlation still routed to the right test. The only thing missing without the subscriber is the client span's own export — server span ends up with an orphan-parent reference in the dashboard but is otherwise correctly correlated.

.NET 8+'s \`DiagnosticsHandler.SendAsyncCore\` injects \`traceparent\`/\`baggage\` based on \`Activity.Current\` as long as one exists, regardless of whether \`System.Net.Http\` has a listener. The listener only gates whether the *client span itself* is recorded/exported. Symmetric pre-fix: the old handler also required \`Activity.Current\` to be non-null (guarded via early return in \`InjectTraceContext\`). Users with no TracerProvider at all don't have test activities emitted in either version, so there's nothing for either implementation to propagate from.

So no correlation regression for users with a test activity active — which is the only meaningful case.

**2. Polling loop comment** — already present:
\`\`\`csharp
// ASP.NET Core stops its server activity on a continuation that may outlive the
// client response, so poll briefly instead of reading _exported synchronously.
\`\`\`

**3. \`HttpRequestHeaders\` carrier cast** — reviewer themselves noted \"harmless, standard OTel SDK boilerplate\". The pattern matches every other \`DistributedContextPropagator.Current.Inject\` callsite in the .NET ecosystem; a comment here would be redundant with the SDK's own conventions.

PR #5680, opened 2026-04-23
fix(mocks): disambiguate IEquatable<T>.Equals from object.Equals (#5675)

### LINKED ISSUE #5675: [Bug] TUnit.Mocks: cannot mock IEquatable<T>.Equals — setup extension shadowed by object.Equals
## Description

Mocking a class that implements `IEquatable<T>` (with `T` = the mocked type itself) produces unusable setup: `mock.Equals(other).Returns(...)` fails to compile because the compiler resolves `mock.Equals(...)` to `object.Equals(object?)` (inherited by `Mock<T>`) instead of to the generator-emitted setup extension method.

## Minimal repro

\`\`\`csharp
public class SelfEquatable : IEquatable<SelfEquatable>
{
    public virtual bool Equals(SelfEquatable? other) => ReferenceEquals(this, other);
    public override int GetHashCode() => 0;
}

var mock = SelfEquatable.Mock();
var other = new SelfEquatable();
mock.Equals(other).Returns(true);  // CS1061 — bool has no Returns
\`\`\`

Error:
\`\`\`
error CS1061: 'bool' does not contain a definition for 'Returns' ...
\`\`\`

## Expected

`mock.Equals(other).Returns(...)` should bind to the generated setup extension and let the user configure and verify the virtual Equals implementation just like any other virtual method.

## Actual

`mock.Equals(other)` binds to the instance method `bool object.Equals(object?)`, which returns `bool`. The setup extension is never reached, so Equals can't be configured or verified.

## Suggested fix

Two options for the generator:

1. Emit a disambiguating helper name for members that collide with `object`'s virtual surface (`Equals`, `GetHashCode`, `ToString`) — e.g. `mock.EqualsOf(...)`.
2. Generate an instance method on `Mock<T>` (not an extension) with a signature more specific than `object.Equals(object?)`, taking `Arg<SelfEquatable?>` so overload resolution picks it.

1. Add a dedicated test that reproduces this failure. The entry point already exists: unskip and restore the `T8` block in `TUnit.Mocks.Tests/KitchenSinkEdgeCasesTests.cs` (both the `SelfEquatable` type and the `T8_Self_Referential_IEquatable_Mockable` test) and confirm it fails with the exact error above in the current codebase.
   - `Returns(...)` actually intercepts the Equals call (setup works).
   - `WasCalled`/`WasNeverCalled` tracks the call (verification works).
   - Both the direct `mock.Object.Equals(other)` path and the `IEquatable<T>.Equals` interface-cast path route to the same setup.

## Context

Found while adding KitchenSink edge-case coverage in #5674. Tracked in `KitchenSinkEdgeCasesTests.cs` as T8 SKIPPED.

thomhurst:
## Summary
- Renames colliding setup-extension members (`Equals`, `GetHashCode`, `ToString`, `GetType`) to `EqualsOf`/`GetHashCodeOf`/`ToStringOf`/`GetTypeOf` so they are no longer shadowed by `object`'s instance methods on `Mock<T>`.
- Restores the SKIPPED `T8_Self_Referential_IEquatable_Mockable` kitchen-sink test, covering setup, direct invocation, interface-cast invocation, and `Was(Never)Called` tracking.

## Breaking change
Yes — minor. Prior to this PR, the four `object`-method names were already in `MockMemberNames` and emitted with a trailing underscore (`Equals_`, `GetHashCode_`, `ToString_`, `GetType_`). Those underscore-suffixed names are NOT shadowed by `object` (different identifier), so they were technically reachable at the call site. Anyone relying on `mock.Equals_(...)` / `mock.GetHashCode_(...)` / `mock.ToString_(...)` / `mock.GetType_(...)` will need to migrate to the new `*Of` names. The expected blast radius is small (the underscore form is undocumented and the only realistic use case — `Equals` — was the broken one users hit and reported as #5675), but flagging it explicitly here.

The `Object` -> `Object_` rename (Mock<T>.Object property collision) is unchanged.

## Why
`Mock<T>` inherits `object.Equals(object?)` as an instance method. C# overload resolution always prefers an instance method over an extension method, so `mock.Equals(other).Returns(...)` bound to `bool object.Equals(object?)` and produced CS1061. Renaming at the emission site is the only fix that works uniformly for both interface and class mocks (class mocks go through `BuildForPartialMock` and return runtime `Mock<T>` from another assembly, so a typed-instance-method approach is not viable).

## Test plan
- [x] T8 (`TUnit.Mocks.Tests/KitchenSinkEdgeCasesTests.cs`) passes — now also covers `GetHashCodeOf` and `ToStringOf` (review feedback)

Closes #5675


thomhurst:
Review feedback addressed in 6e68366ec:

**Applied**
1. **Naming-strategy comment (item 1)** — expanded comments above `MockMemberNames` and `ObjectMemberDisambiguations` to explicitly contrast the two distinct shadowing problems: `Object` is a property/method member-kind collision (underscore suffices), whereas `Equals/GetHashCode/ToString/GetType` are inherited `object` instance-method collisions that overload resolution would still hide regardless of suffix style.
2. **Test coverage (item 2)** — extended T8 to also cover `GetHashCodeOf` and `ToStringOf` end-to-end (setup -> call -> verify). `GetTypeOf` is documented inline as not exercisable: `object.GetType` is non-virtual, so no derived type can override it and the helper is unreachable in practice. Added an `override string ToString()` to `SelfEquatable` so the typed `ToStringOf` setup is generated.
3. **Breaking-change determination (item 3)** — researched the prior emission via `git show 932d2538c -- ...MockMembersBuilder.cs`. Before this PR, `MockMemberNames` already contained `Equals/GetHashCode/ToString/GetType` and emitted them as `Equals_`/`GetHashCode_`/`ToString_`/`GetType_`. Underscore-suffixed names are NOT shadowed by `object` (different identifier), so they WERE reachable. The rename to `*Of` is therefore a (minor) breaking change for anyone using the undocumented underscore form. Noted explicitly in the updated PR body's new \"Breaking change\" section.

**Deferred / not applicable**
- **Codacy issues (item 4)** — ran `dotnet format --verify-no-changes` against both `TUnit.Mocks.SourceGenerator` and `TUnit.Mocks.Tests`. The whitespace/import warnings it surfaced are all in pre-existing files and pre-existing lines (none on the lines this PR touches in `MockMembersBuilder.cs` or in the T8 block). Per the review's explicit \"changed files only\" scoping I left those alone to avoid muddying the diff. The 3 Codacy findings (1 BestPractice + 2 CodeStyle) likely reflect Codacy's own analyzer ruleset rather than `dotnet format` rules and have no detail in the GitHub check — viewing the Codacy dashboard at https://app.codacy.com/gh/thomhurst/TUnit/pull-requests/5680/issues is required to see them.
- **Suggestion #4 (expression-tree `Setup(x => x.Equals(...))` design)** — out of scope per the reviewer's own framing; tracked for a future ticket.

Test status: T8 passes; full `TUnit.Mocks.Tests` net10.0 = 953/953; `TUnit.Mocks.SourceGenerator.Tests` net10.0 = 45/45.

thomhurst:
Round-2 review feedback applied in f4a63be22:

**1. Snapshot coverage for `*Of` helpers — applied.**
Added `SelfEquatable_Generates_EqualsOf_GetHashCodeOf_ToStringOf` snapshot test in `TUnit.Mocks.SourceGenerator.Tests` using a `SelfEquatableSnapshot` type that implements `IEquatable<T>` and overrides `GetHashCode`/`ToString`. The new `.verified.txt` records the generated `EqualsOf` (4 overloads — Arg+Func, both signatures), `GetHashCodeOf`, and `ToStringOf` extensions. The runtime T8 test no longer carries the burden alone; future regressions will surface as snapshot diffs.

**2. Drop unreachable `GetTypeOf` — applied.**
Removed the `{ "GetType", "GetTypeOf" }` entry from `ObjectMemberDisambiguations` in `TUnit.Mocks.SourceGenerator/Builders/MockMembersBuilder.cs`. `object.GetType` is non-virtual so the generated helper could never be useful. Updated the class-level comment to drop the `GetType` reference and added a sentence explaining the omission. Also tightened the related comment block in `KitchenSinkEdgeCasesTests.cs` (T8) so it no longer claims `GetTypeOf` is generated.

**Verification:**
- `TUnit.Mocks.SourceGenerator.Tests` (net10.0): 46/46 pass — confirms no other snapshot referenced `GetTypeOf`.
- `TUnit.Mocks.Tests` (net10.0): 953/953 pass — including T8.
- No `.received.txt` files left behind.

PR #5681, opened 2026-04-23
fix(mocks): skip MockBridge for class targets with static-abstract interfaces (#5677)

### LINKED ISSUE #5677: [Bug] TUnit.Mocks: mocking a class that implements a static-abstract interface emits invalid MockBridge
## Description

When mocking a class that implements an interface with `static abstract` members, the generator emits a `MockBridge` that treats the *class* as if it were an interface, producing CS0527 and CS0540.

## Minimal repro

\`\`\`csharp
public interface IStaticAbstractFactory
{
    static abstract IStaticAbstractFactory Create();
    int InstanceValue { get; }
}

public class StaticAbstractImpl : IStaticAbstractFactory
{
    public static IStaticAbstractFactory Create() => new StaticAbstractImpl();
    public virtual int InstanceValue => 99;
}

var mock = StaticAbstractImpl.Mock();
\`\`\`

Errors (in generated `*_MockBridge.g.cs`):
\`\`\`
error CS0527: Type 'StaticAbstractImpl' in interface list is not an interface
error CS0540: '...IStaticAbstractFactory.Create()': containing type does not implement interface 'IStaticAbstractFactory'
\`\`\`

## Expected

Mocking a class that implements a static-abstract interface should work — the mock should either inherit the class's static impl or skip bridge generation since the class (not an interface) is the target.

## Actual

`MockBridgeBuilder` adds the mocked target to an interface list and emits explicit interface implementations, but for a class target neither operation is legal.

## Suggested fix

`MockBridgeBuilder` should guard on `TypeKind`: the bridge shape is only needed when the mock target is itself an interface with static-abstract members that needs a runtime dispatch hook. For a class target, the class already provides the concrete static members — no bridge needed.

1. Add a dedicated test that reproduces this failure. The entry point already exists: unskip and restore the `T15` block in `TUnit.Mocks.Tests/KitchenSinkEdgeCasesTests.cs` (both the `IStaticAbstractFactory` / `StaticAbstractImpl` types and the `T15_Class_Implementing_Static_Abstract_Interface_Mockable` test) and confirm the build fails with the exact CS0527 and CS0540 above in the current codebase.
   - The class mock compiles despite the static-abstract interface.
   - Instance virtual members are mockable via `Returns(...)` and verifiable via `WasCalled`.
   - The class's concrete static impl still works (call `StaticAbstractImpl.Create()` and verify it returns a real instance — not routed through the mock engine).
4. Run the full `TUnit.Mocks.Tests` and `TUnit.Mocks.SourceGenerator.Tests` suites to confirm no regression — in particular, ensure existing static-abstract-on-interface mocks (`StaticAbstractMemberTests`) still work.

## Context

Found while adding KitchenSink edge-case coverage in #5674. Tracked in `KitchenSinkEdgeCasesTests.cs` as T15 SKIPPED.

thomhurst:
## Summary
- Stops generating `_MockBridge.g.cs` when the mock target is a class. Bridge shape is only required for interface targets that need a runtime dispatch hook for static-abstract members; a class already provides the concrete static implementation directly.
- Adds a TypeKind guard in `MemberDiscovery` so static-abstract interface members are not collected for class targets, plus a defensive `model.IsInterface` check in `MockGenerator.GenerateSingleTypeMock`.
- Restores the SKIPPED `T15_Class_Implementing_Static_Abstract_Interface_Mockable` kitchen-sink test.

## Why
Previously `MemberDiscovery` walked a class's implemented interfaces and pulled in their static-abstract members, which set `HasStaticAbstractMembers = true` on the model. `MockBridgeBuilder` then emitted a bridge that put the class in its own interface base list (CS0527) and produced explicit interface impls without listing the interface (CS0540), making any class that implements a static-abstract interface unmockable.

## Test plan
- [x] T15 (`TUnit.Mocks.Tests/KitchenSinkEdgeCasesTests.cs`) passes
- [x] `StaticAbstractMemberTests` (interface-target path): 13/13

Closes #5677


thomhurst:
Applied review feedback in 46f589409:

1. **Consolidated the static-abstract guard.** Extracted `ShouldCollectStaticAbstractFromInterfaces(ITypeSymbol)` in `MemberDiscovery.cs`. Both `DiscoverMembers` and `DiscoverMembersFromMultipleTypes` now route through the same helper, and its doc comment spells out that any new collection loop MUST gate static-abstract collection through it.
2. **Trimmed the `MockGenerator.GenerateSingleTypeMock` comment** to a one-liner pointing at the MemberDiscovery contract: `// Defence-in-depth: MemberDiscovery already filters static-abstract collection on TypeKind, so HasStaticAbstractMembers is always false here for class targets.` Kept the `&& model.IsInterface` guard.
3. **Stripped `(#5677)` issue references** from source comments in `MemberDiscovery.cs` and `MockGenerator.cs`. The WHY (CS0527 / CS0540 / class-vs-interface semantics) is preserved on the new helper's doc comment.

**Tests (net10.0):**
- `KitchenSinkEdgeCasesTests/T15_*` -> 1/1 pass
- `StaticAbstractMemberTests/*` -> 13/13 pass
- `TUnit.Mocks.Tests` (full) -> 953/953 pass
- `TUnit.Mocks.SourceGenerator.Tests` (full) -> 45/45 pass

No snapshot diffs produced; nothing to commit beyond the source edits.

thomhurst:
Round 2 review applied in 8efe4cf17.

**Item 1 (structural enforcement):** Funnelled both `DiscoverMembers` and `DiscoverMembersFromMultipleTypes` static-member branches through a single new private helper `TryCollectStaticAbstractFromInterface(member, typeSymbol, interfaceFqn, ...)`. The helper derives the static-abstract collection flag from the `typeSymbol` argument internally (calling `ShouldCollectStaticAbstractFromInterfaces`). The two per-loop `var collectStaticAbstractFromInterfaces = ...` locals are removed, so the only code path that reaches `CollectStaticAbstractMember` now goes through the gated helper. A future third loop physically cannot skip the gate without re-implementing the helper.

Chose this shape over a `MemberCollectionContext` struct because the call sites already had `typeSymbol` in scope and only needed one piece of derived state — a 12-line helper makes the bypass impossible with strictly less surface area than a context type.

**Item 2 (defence-in-depth `&& model.IsInterface`):** Left as-is per follow-up review.

**Tests (net10.0):**
- T15 `KitchenSinkEdgeCasesTests`: 1/1 pass
- `StaticAbstractMemberTests`: 13/13 pass
- `TUnit.Mocks.Tests`: 953/953 pass
- `TUnit.Mocks.SourceGenerator.Tests`: 45/45 pass (no snapshot diffs)

thomhurst:
3rd-round review feedback applied in 5b24c6081.

**Per item:**

1. **(M) Guard tightened** — `ShouldCollectStaticAbstractFromInterfaces` now uses `typeSymbol.TypeKind == TypeKind.Interface` (was `!= TypeKind.Class`). Encodes intent and stays robust against future TypeKinds (struct/enum/delegate would have slipped past the old check).
2. **(M) Snapshot added** — `Class_Implementing_Static_Abstract_Interface` mirrors the T15 type shape. The new `Snapshots/Class_Implementing_Static_Abstract_Interface.verified.txt` contains zero `MockBridge` occurrences (confirmed via `grep -c MockBridge` → `0`).
3. **(L) Stray `#5677`** — removed from the `TryCollectStaticAbstractFromInterface` XML doc. Kept the CS0527/CS0540 + class-vs-interface rationale.
4. **(L) Oversized comment** — `MockGenerator.cs` invariant comment reformatted to the 3-line block recommended in the review.
5. **(L) Misleading XML doc** — `DiscoverMembersFromMultipleTypes` summary now notes that the first element may be a class (via `MockTypeDiscovery.TransformToModels` with `isPartialMock == true`), so the `TypeKind` guard in `TryCollectStaticAbstractFromInterface` is genuinely required.
6. Skipped per reviewer.

**Test status (all net10.0):**
- T15 (`KitchenSinkEdgeCasesTests.T15_Class_Implementing_Static_Abstract_Interface_Mockable`): 1/1 passed.
- `StaticAbstractMemberTests`: 13/13 passed.
- `TUnit.Mocks.Tests`: 953/953 passed.
- `TUnit.Mocks.SourceGenerator.Tests` (`MockGeneratorTests/*`): 46/46 passed (including the new snapshot).

No `*.received.txt` files committed.

thomhurst:
Final-round review feedback addressed in 8cdd519cb (doc-only follow-up to 5b24c6081):

**Applied (both items 2 and 3):**
- Removed the stray `#5677` reference from the `Class_Implementing_Static_Abstract_Interface` test comment in `MockGeneratorTests.cs` — the WHY (CS0527 / CS0540 explanation) is preserved; only the now-redundant issue link was stripped.
- Added an explicit one-line note that the verified snapshot for that test intentionally OMITS a `_MockBridge.g.cs` file section, so the implicit "no bridge" assertion is now stated intent for future debuggers.

**Deferred (per reviewer guidance):**
- Item 1 (`seenFullMethods` not threaded to `CollectStaticAbstractMember`) — pre-existing inconsistency, not introduced by this PR.
- Item 4 (`MemberCollectionContext` ref struct refactor) — out of scope for a targeted bug fix.
- Item 5 (inline `ShouldCollectStaticAbstractFromInterfaces`) — reviewer marked either form acceptable; current shape kept for the structural-guard intent already documented in the helper's XML.

**Tests (net10.0, all green):**
- T15 (`KitchenSinkEdgeCasesTests`): 1/1
- `StaticAbstractMemberTests` (`TUnit.Mocks.Tests`): 13/13
- `TUnit.Mocks.Tests` full suite: 953/953
- `TUnit.Mocks.SourceGenerator.Tests` full suite: 46/46 (incl. `Class_Implementing_Static_Abstract_Interface` snapshot)
- No new `.received.txt`.

**Codacy:** ran `dotnet format --verify-no-changes` on the changed projects. The whitespace/imports findings that surfaced are all in unchanged code (pre-existing in `MemberDiscovery.cs` lines ~1159-1181 and `KitchenSinkEdgeCasesTests.cs` lines 358-360 / 446) — none touch lines this PR added or modified, so no in-diff format fixes to apply.

thomhurst:
Resolved merge conflicts with `origin/main` after sister PRs #5680 (T8 SelfEquatable) and #5683 (T14 indexer / T14b in-indexer) merged.

**Conflict scope:** Two regions in `TUnit.Mocks.Tests/KitchenSinkEdgeCasesTests.cs` — the type-declaration block and the test-method block. Both sides were purely additive:
- main: T14 (`IHasIndexer`, `IHasInIndexer`) types + 3 test methods, plus a "T15 SKIPPED" comment.
- this branch: T15 (`IStaticAbstractFactory`, `StaticAbstractImpl`) types + test, plus a "T14 SKIPPED" comment.

**Resolution:** kept all real types and tests from both sides (T14, T14b, T15) and dropped both "SKIPPED" placeholder comments since each side now has a working implementation. No type or method names collide.

**Verification (net10.0):**
- `TUnit.Mocks.SourceGenerator.Tests`: 47/47 pass.
- `TUnit.Mocks.Tests` full suite: 957/957 pass.
- `KitchenSinkEdgeCasesTests` filtered: 20/20 pass — confirmed `T14_*`, `T14b_*`, and `T15_Class_Implementing_Static_Abstract_Interface_Mockable` all green.

No `.received.txt` snapshot files produced. Merge commit: `e39e5e03d`.

PR #5682, opened 2026-04-23
fix(mocks): emit [SetsRequiredMembers] on generated mock ctor (#5678)

### LINKED ISSUE #5678: [Bug] TUnit.Mocks: mocking a type with `required` members fails with CS9035
## Description

Mocking a class that declares a `required` member fails to compile because the generated factory uses an object initializer that does not set the required member.

## Minimal repro

\`\`\`csharp
public abstract class RequiredShape
{
    public required string Name { get; init; }
    public abstract int Compute();
}

var mock = RequiredShape.Mock();
\`\`\`

Error (in generated `*_MockImplFactory.g.cs`):
\`\`\`
error CS9035: Required member 'RequiredShape.Name' must be set in the object initializer or attribute constructor.
\`\`\`

## Expected

Mock construction should succeed without the user having to set required members — the mock is a test double, it does not need the compile-time enforcement of `required`.

## Actual

Factory can't be instantiated at all, so the type can't be mocked.

## Suggested fix

Emit `[System.Diagnostics.CodeAnalysis.SetsRequiredMembers]` on the generated mock constructor (the private one inside the factory). This tells the compiler that the constructor itself takes responsibility for the required members, bypassing the initializer requirement.

1. Add a dedicated test that reproduces this failure. The entry point already exists: unskip and restore the `T17` block in `TUnit.Mocks.Tests/KitchenSinkEdgeCasesTests.cs` (both the `RequiredShape` type and the `T17_Required_Property_Does_Not_Block_Mock_Instantiation` test) and confirm the build fails with the exact CS9035 above in the current codebase.
   - Mock instantiation succeeds without the user providing a value for the required member.
   - The abstract/virtual members on the required-carrying type are mockable via `Returns(...)` and verifiable via `WasCalled`.
   - Extend the coverage with at least one extra shape: a class that has multiple required members (including a required reference type and a required value type) so the initializer-skip logic is exercised broadly.

## Context

Found while adding KitchenSink edge-case coverage in #5674. Tracked in `KitchenSinkEdgeCasesTests.cs` as T17 SKIPPED.

thomhurst:
## Summary
- Emits `[SetsRequiredMembers]` on every generated `XxxMockImpl` / `XxxWrapMockImpl` constructor so the factory can `new` them without the user having to satisfy `required` members at the call site.
- Restores the SKIPPED `T17_Required_Property_Does_Not_Block_Mock_Instantiation` and adds a multi-required shape (ref + value type) so the initializer-skip logic is exercised broadly.

## Why
A class with `required` members (e.g. `public required string Name { get; init; }`) couldn't be mocked: the generated factory's `new XxxMockImpl(engine)` produced CS9035 because the constructor didn't declare it took responsibility for the required members. Annotating the generated ctor with `[SetsRequiredMembers]` tells the compiler the constructor satisfies the contract — which is appropriate for a mock that is itself the test double.

No polyfill was needed: net8.0+ ships the attribute, and the repo's `Polyfill` package supplies it for net472.

## Test plan
- [x] T17 single-required test passes
- [x] T17 multi-required (ref + value) test passes
- [x] `TUnit.Mocks.Tests` full suite (net10.0): 954/954

Closes #5678


thomhurst:
## Review feedback applied

**Applied (commit `1b0b89cc2`):**
- Added explicit default-state assertions to both T17 tests in `TUnit.Mocks.Tests/KitchenSinkEdgeCasesTests.cs`:
  - `T17_Required_Property_Does_Not_Block_Mock_Instantiation`: asserts `mock.Object.Name` is `null`.
  - `T17_Multiple_Required_Members_Reference_And_Value_Types`: asserts `Title` is `null`, `Count` is `0`, `Id` is `Guid.Empty`.
- Added the requested one-line comment above each block: `// Required members intentionally remain at default in mocks; [SetsRequiredMembers] suppresses CS9035 only.`

**Deferred:**
- The `WriteConstructorPreamble` helper extraction in `MockImplBuilder.cs` — per reviewer's own low-priority/YAGNI note, and given the prior `b6edcbc15` cleanup already hoisted the attribute string to a `const`.

**Verification (net10.0):**
- `TUnit.Mocks.Tests` T17 filter: 2/2 passed.
- `TUnit.Mocks.Tests` full suite: 954/954 passed.
- `TUnit.Mocks.SourceGenerator.Tests`: 45/45 passed.
- No `.received.txt` snapshot diffs produced (test-only change).

thomhurst:
2nd-round review applied in b80c3c5ab:

1. **Comment tightened** — `MockImplBuilder.cs` const-decl comment collapsed from 2 lines to a single one-liner per project style guideline:
   ```csharp
   // Suppresses CS9035: ctor claims responsibility for required members so factory can `new XxxMockImpl(engine)` without initializers.
   ```

2. **Required-members snapshot added** — new `Class_With_Required_Members` test in `MockGeneratorTests.cs` covers the exact `required`-member scenario fixed by #5678 (one ref-typed `string` + one ref-typed `object` required property on a `ConfigBase` partial-mock target). Verified `.verified.txt` shows `[SetsRequiredMembers]` correctly emitted on the generated impl ctor; locks in generator output as a regression guard.

3. **Skipped** — `WriteConstructorPreamble` helper (reviewer marked advisory; const-hoist already covered the duplication concern in round 1).

**Tests** — all green on net10.0:
- T17 required-member runtime tests in `KitchenSinkEdgeCasesTests`: 2/2 pass
- `TUnit.Mocks.Tests`: 954/954 pass
- `TUnit.Mocks.SourceGenerator.Tests`: 46/46 pass (45 prior + 1 new)

No `.received.txt` committed.

thomhurst:
Merged latest `origin/main` (which now contains the merged sister PRs #5680 and #5683).

**Conflict resolution** (single hunk in `TUnit.Mocks.SourceGenerator.Tests/MockGeneratorTests.cs`):
- Both sides added a new `[Test]`-attributed snapshot test method right after `Class_Implementing_Forwarded_Interface_Hierarchy`. Resolution kept both methods: this branch's `Class_With_Required_Members` followed by main's `SelfEquatable_Generates_EqualsOf_GetHashCodeOf_ToStringOf`.
- No duplicates: 46 distinct `public Task` test methods after merge.

**Snapshot reconciliation:**
- `SelfEquatable_Generates_EqualsOf_GetHashCodeOf_ToStringOf.verified.txt` (added on main) needed one line — `[global::System.Diagnostics.CodeAnalysis.SetsRequiredMembers]` on the generated mock impl ctor — because that attribute is exactly what this PR emits for every generated impl. Verified the diff is the intentional fix from #5678 and accepted the regenerated snapshot.

**Verification (net10.0):**
- `TUnit.Mocks.SourceGenerator.Tests`: 47 / 47 passed.
- `TUnit.Mocks.Tests` filtered to `KitchenSinkEdgeCasesTests/*`: 21 / 21 passed (includes T17 `T17_Required_Property_Does_Not_Block_Mock_Instantiation` and `T17_Multiple_Required_Members_Reference_And_Value_Types`).

Merge commit: `cb93299da`. Snapshot follow-up: `0e51359b9`.

PR #5683, opened 2026-04-23
fix(mocks): implement indexer in generated mock (#5676)

### LINKED ISSUE #5676: [Bug] TUnit.Mocks: interface with indexer fails to compile — CS0535 on generated mock impl
## Description

Mocking an interface that declares an indexer produces a compile error because the generator skips indexer emission in `MockImplBuilder` (all three builders contain `if (prop.IsIndexer) continue;`) but the generated class still has the interface in its base list, leaving the indexer unimplemented.

## Minimal repro

\`\`\`csharp
public interface IHasIndexer
{
    string this[int index] { get; set; }
    int Regular { get; set; }
}

var mock = IHasIndexer.Mock();
\`\`\`

Error:
\`\`\`
error CS0535: 'IHasIndexerMockImpl' does not implement interface member 'IHasIndexer.this[int]'
\`\`\`

## Expected

Either:
- the generator implements the indexer (routed through the mock engine like properties), or
- it emits a stub that throws `NotSupportedException` so the interface contract is satisfied and the rest of the interface stays mockable.

## Actual

Indexer is silently dropped. Any interface containing an indexer can't be mocked at all, even if the consumer only needs the other members.

## Suggested fix

In the three sites that do `if (prop.IsIndexer) continue;` (`MockImplBuilder.cs` around lines 68, 116, 454), emit a throwing stub that satisfies the contract. Indexer get/set are still routable through the engine with member-id dispatch the same as regular properties — full support is also possible.

1. Add a dedicated test that reproduces this failure. The entry point already exists: unskip and restore the `T14` block in `TUnit.Mocks.Tests/KitchenSinkEdgeCasesTests.cs` (both the `IHasIndexer` type and the `T14_Interface_With_Indexer_Compiles_Regular_Property_Works` test) and confirm the build fails with the exact CS0535 above in the current codebase.
   - The interface compiles with the mock (contract satisfied).
   - The non-indexer members of the same interface still work (`Returns`, `WasCalled`).
   - If full indexer support is added: indexer `get`/`set` are both configurable via `Returns(...)` and verifiable via `WasCalled`; different index values produce independent setups.
   - If stub-only support is added: calling the indexer throws a clear exception.

## Context

Found while adding KitchenSink edge-case coverage in #5674. Tracked in `KitchenSinkEdgeCasesTests.cs` as T14 SKIPPED.

thomhurst:
## Summary
- Implements interface/partial/wrap indexer emission in `MockImplBuilder`, replacing the three `if (prop.IsIndexer) continue;` skip sites that were leaving `IFoo.this[int]` unimplemented and producing CS0535.
- Adds `Item(args)` and `SetItem(args, value)` extension methods in `MockMembersBuilder` so per-index setups (`Returns`) and verifications (`WasCalled`) work via the existing `MockMethodCall<T>` / `VoidMockMethodCall` plumbing.
- Restores the SKIPPED `T14` kitchen-sink tests covering contract satisfaction, the regular-property co-resident path, per-index Returns/WasCalled, and setter verification with mixed `Any<T>()` / literal matchers.

## Why
Any interface with an indexer was previously unmockable — even if the consumer only needed the other members. The full-support route was small (the indexer model was already discovered) and avoids shipping a stub that would have made `IList`-style mocks throw at runtime.

## Known limitations
- `MockWrapperTypeBuilder.CanGenerateWrapper` still returns `false` when an indexer is present, so the typed `IFoo_Mock` wrapper is not emitted for indexer-bearing interfaces. The base `Mock<IFoo>` API works fully; only the typed-wrapper sugar is unavailable. Out of scope for this PR.
- Static-abstract indexers in `MockBridgeBuilder` are still skipped (separate code path, no interface contract failure).

## Test plan
- [x] T14 (2 tests in `TUnit.Mocks.Tests/KitchenSinkEdgeCasesTests.cs`) pass
- [x] `TUnit.Mocks.Tests` full suite (net10.0): 954/954
- [x] `TUnit.Mocks.Analyzers.Tests` (net10.0): 30/30

Closes #5676

### DIFF
--- a/TUnit.Mocks.SourceGenerator/Builders/MockImplBuilder.cs
+++ b/TUnit.Mocks.SourceGenerator/Builders/MockImplBuilder.cs
@@ -1064,67 +1064,21 @@ internal static class MockImplBuilder
     /// to the base implementation. See <see cref="GenerateInterfaceIndexer"/> for the dispatch shape.
     /// </summary>
     private static void GeneratePartialIndexer(CodeWriter writer, MockMemberModel prop)
-    {
-        var accessModifier = prop.IsProtected ? "protected" : "public";
-        var paramList = FormatIndexerParameterList(prop);
-        var argPassList = string.Join(", ", prop.Parameters.Select(p => p.Name));
-        var argsArray = GetIndexerGetterArgsArray(prop);
-        writer.AppendLineIfNotEmpty(prop.ObsoleteAttribute);
-        writer.AppendLine($"{accessModifier} override {prop.ReturnType} this[{paramList}]");
-        writer.OpenBrace();
-
-        if (prop.HasGetter)
-        {
-            writer.AppendLineIfNotEmpty(prop.GetterObsoleteAttribute);
-            if (prop.IsAbstractMember)
-            {
-                writer.AppendLine($"get => _engine.HandleCallWithReturn<{prop.ReturnType}>({prop.MemberId}, \"get_Item\", {argsArray}, {prop.SmartDefault});");
-            }
-            else
-            {
-                writer.AppendLine("get");
-                writer.OpenBrace();
-                writer.AppendLine($"if (_engine.TryHandleCallWithReturn<{prop.ReturnType}>({prop.MemberId}, \"get_Item\", {argsArray}, {prop.SmartDefault}, out var __result))");
-                writer.AppendLine("{");
-                writer.IncreaseIndent();
-                writer.AppendLine("return __result;");
-                writer.DecreaseIndent();
-                writer.AppendLine("}");
-                writer.AppendLine($"return base[{argPassList}];");
-                writer.CloseBrace();
-            }
-        }
-
-        if (prop.HasSetter)
-        {
-            var setterArgs = GetIndexerSetterArgsArray(prop);
-            writer.AppendLineIfNotEmpty(prop.SetterObsoleteAttribute);
-            if (prop.IsAbstractMember)
-            {
-                writer.AppendLine($"set => _engine.HandleCall({prop.SetterMemberId}, \"set_Item\", {setterArgs});");
-            }
-            else
-            {
-                writer.AppendLine("set");
-                writer.OpenBrace();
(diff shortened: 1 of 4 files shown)


thomhurst:
Applied review feedback in 88f777698.

**Applied**
1. Removed `<summary>` doc blocks from `GenerateInterfaceIndexer`, `GeneratePartialIndexer`, `GenerateWrapIndexer`, and `GenerateOverrideIndexer`. Names are self-describing per project standards.
2. Fixed the modifier-drop bug. Two changes were needed:
   - `MemberDiscovery.CreateIndexerModel` was hardcoding `Direction = ParameterDirection.In` for indexer parameters; now uses `p.GetParameterDirection()` (the same helper method/event discovery uses).
   - `FormatIndexerParameterList` now delegates to the existing `FormatParameterList`, so the modifier prefix (`in `, etc.) is forwarded.
   - Added T14b: `IHasInIndexer` with `string this[in int key]` exercising the path. Generated impl now correctly emits `public string this[in int key]`.
3. Moved `argsArray` inside `if (HasGetter)` in `GenerateInterfaceIndexer` and `GenerateOverrideIndexer`; setters compute their own args separately.
4. `dotnet format --verify-no-changes` reports several pre-existing whitespace/import issues in unrelated files (`MemberDiscovery.cs` line 78, `MockTypeDiscovery.cs`, `EquatableArray.cs`, `MockParameterModel.cs`, plus older lines of `KitchenSinkEdgeCasesTests.cs`). None are in the lines this PR touches, so left alone to keep the diff scoped. Codacy's specific findings still need to be viewed on their dashboard.

**Test status (net10.0)**
- TUnit.Mocks.Tests: 955/955 passing
- TUnit.Mocks.SourceGenerator.Tests: 45/45 passing
- TUnit.Mocks.Analyzers.Tests: 30/30 passing

**Deferred / out-of-scope**
- `MockWrapperTypeBuilder.CanGenerateWrapper` returning `false` for indexer interfaces remains a known limitation per the original PR body. Happy to file a follow-up issue if you'd like — let me know.

PR #5684, opened 2026-04-23
fix(mocks): escape C# keyword identifiers at all emit sites (#5679)

### LINKED ISSUE #5679: [Bug] TUnit.Mocks: member names matching C# keywords (class, event, params) break generated mock
## Description

When a mocked interface or class declares a member whose name matches a C# reserved keyword (`class`, `event`, `record`, `params`, etc.), the generator emits the name unescaped in the generated impl/members/extensions code, producing a cascade of compile errors.

## Minimal repro

\`\`\`csharp
public interface IEscapedNames
{
    int @class { get; }
    string @record();
    void @event(int @params);
}

var mock = IEscapedNames.Mock();
\`\`\`

Errors (in generated `*_Mock.g.cs`, `*_MockMembers.g.cs`, `*_MockImplFactory.g.cs`):
\`\`\`
error CS0539: '...IEscapedNamesMock.()' in explicit interface declaration is not found
error CS0106: The modifier 'new' is not valid for this item
error CS0066: 'IEscapedNamesMock.': event must be of a delegate type
error CS0708: '...mock': cannot declare instance members in a static class
error CS0542: '': member names cannot be the same as their enclosing type
... (many more)
\`\`\`

## Expected

Member names are accepted regardless of keyword collision — the generator escapes them with `@` when emitting the identifier in generated C#.

## Actual

The `EscapeIdentifier` helper (in `MemberDiscovery.cs`, line 1040) already exists and already escapes **parameter** names. It is not applied to method, property, or event names when they feed into the emitted identifier positions, so any user-declared member whose name is a keyword produces malformed code.

## Suggested fix

Apply `EscapeIdentifier` at every emission site where a member name becomes a C# identifier:
- method/property/event declarations in `MockImplBuilder`
- extension method names in `MockMembersBuilder`
- explicit interface member references in `MockBridgeBuilder` and elsewhere

Keep the stored model `Name` unescaped (used for dispatch/log/keys) and escape only at emit.

1. Add a dedicated test that reproduces this failure. The entry point already exists: unskip and restore the `T18` block in `TUnit.Mocks.Tests/KitchenSinkEdgeCasesTests.cs` (both the `IEscapedNames` interface and the `T18_Member_Names_That_Are_Contextual_Keywords` test) and confirm the build fails with the cascade of errors above in the current codebase.
   - Property named after a keyword: `Returns(...)` intercepts the getter, `WasCalled` tracks it.
   - Method named after a keyword: `Returns(...)` intercepts, `WasCalled` tracks.
   - Void method named after a keyword with a keyword-named parameter: `WasCalled` tracks.
   - Extend with at least one escaped parameter name collision check — e.g. a method that has two parameters both named after keywords — so the parameter-escape path is jointly verified with the member-name escape path.
4. Run the full `TUnit.Mocks.Tests` and `TUnit.Mocks.SourceGenerator.Tests` suites to confirm no regression — in particular the member-name is used as a dispatch key, so dispatch keys must not accidentally include the `@`.

## Context

Found while adding KitchenSink edge-case coverage in #5674. Tracked in `KitchenSinkEdgeCasesTests.cs` as T18 SKIPPED.

thomhurst:
## Summary
- Adds `IdentifierEscaping.EscapeIdentifier` and applies it at every emission site in `MockImplBuilder`, `MockMembersBuilder`, `MockBridgeBuilder`, `MockEventsBuilder`, and `MockWrapperTypeBuilder` where a member name becomes a C# identifier.
- Stored model `Name` values stay UNESCAPED, so engine dispatch keys, logging, and identity remain `"event"` / `"get_class"` / `"namespace"` rather than `"@event"` etc.
- Restores the SKIPPED `T18_Member_Names_That_Are_Contextual_Keywords` kitchen-sink test, with an `IEscapedNames` interface covering keyword-named property (`@class`), function (`@record()`), void method with keyword parameter (`@event(@params)`), and a method with two keyword-named parameters and a keyword-named method name (`@namespace(@new, @static)`) to exercise the joint escape paths.
- Dedupe pass: collapses the pre-existing private `EscapeIdentifier` helper in `MemberDiscovery` into the new shared utility so there is one keyword-escape implementation across the generator.

## Why
Any interface or class that declared a member whose name happened to match a C# keyword (`class`, `event`, `record`, `namespace`, etc.) emitted unescaped identifiers at member declaration sites — producing a cascade of CS0539 / CS0106 / CS0066 / CS0708 / CS0542 errors and making the type unmockable. The `EscapeIdentifier` helper already existed at the parameter-discovery layer; this PR threads it through every member-name emission site.

## Test plan
- [x] T18 (`TUnit.Mocks.Tests/KitchenSinkEdgeCasesTests.cs`) passes
- [x] All 4 source-generator project variants build (base + Roslyn 4.14 / 4.4 / 4.7)
- [x] Dispatch keys in generated factories remain unescaped (`"event"`, `"get_class"`, `"namespace"`)

Closes #5679


thomhurst:
Addressed review feedback in f0cf2b441 (new commit on top, no amends):

1. **Added snapshot test** — `Interface_With_Keyword_Member_Names` in `TUnit.Mocks.SourceGenerator.Tests/MockGeneratorTests.cs`. Mixed coverage: keyword-named property (`@class`), keyword-named methods (`@event`, `@namespace`, plus contextual `record`), and keyword-named parameters (`@params`, `@new`, `@static`). New `.verified.txt` snapshot accepted; no `.received.txt` committed.

2. **Renamed T18** to `T18_Member_Names_That_Are_Reserved_Keywords` in `TUnit.Mocks.Tests/KitchenSinkEdgeCasesTests.cs`. Kept `record` on `IEscapedNames` with a one-line comment noting it's a contextual keyword and is there to confirm contextual keywords pass through `EscapeIdentifier` unchanged.

3. **Documented `IdentifierEscaping`** — added a WHY-only comment above `EscapeIdentifier` explaining why only `GetKeywordKind` (reserved) is used, and that escaping contextual keywords like `record`/`async`/`var`/`get`/`set` would change identifier text and break explicit interface impl matching.

**Test status (net10.0):**
- `TUnit.Mocks.SourceGenerator.Tests`: 46/46 passed
- `TUnit.Mocks.Tests` (incl. T18 renamed): 953/953 passed

thomhurst:
## Round-2 review feedback applied (commit `9fd285d3`)

**1. Visibility mismatch — applied.**
`IdentifierEscaping.EscapeIdentifier` is now `internal` (was misleadingly `public` inside an `internal static class`).

**2. `GetWrapperName` deliberate raw-name embedding — documented.**
Added a WHY comment above `MockMembersBuilder.GetWrapperName` clarifying that `method.Name` is intentionally embedded raw: the result is a compound identifier like `IFoo_event_M2_MockCall`, where a keyword as a non-terminal substring of a larger token is valid C#. `EscapeIdentifier` is only required when the name would stand alone as a complete identifier in the emitted source.

**3. `record` future-keyword risk — no action** (flagged by reviewer as informational only).

### Verification (net10.0)
- `T18_Member_Names_That_Are_Reserved_Keywords` — 1/1 passed
- `TUnit.Mocks.Tests` — 953/953 passed
- `TUnit.Mocks.SourceGenerator.Tests` — 46/46 passed
- No snapshot drift (no `.received.txt` produced).

thomhurst:
Doc clarification applied in `b296be2cd`: extended `GetWrapperName` comment to note that `safeName` may legitimately start with `@` (when the member name is a reserved keyword), and that this is valid C# — the `@` prefix applies to the entire compound identifier. Documentation-only; no logic change.

thomhurst:
Merged latest `origin/main` (commit 1b63f8821).

**Conflict:** single hunk in `TUnit.Mocks.SourceGenerator/Builders/MockMembersBuilder.cs` at `GetSafeMemberName`.
- main (#5680) added an `ObjectMemberDisambiguations` lookup mapping `Equals`/`GetHashCode`/`ToString` to their `*Of` aliases.
- this branch added an `EscapeIdentifier(...)` wrapper so keyword-named members get the `@` prefix.

**Resolution:** combined both — the disambiguation lookup runs first (its renamed values like `EqualsOf` are not C# keywords, so escaping them is unnecessary), and the fallback path keeps `EscapeIdentifier(...)` around the `MockMemberNames`/raw-name branch:

```csharp
private static string GetSafeMemberName(string name)
{
    if (ObjectMemberDisambiguations.TryGetValue(name, out var renamed))
        return renamed;
    return EscapeIdentifier(MockMemberNames.Contains(name) ? name + "_" : name);
}
```

**EscapeIdentifier audit of main's new emit sites:**
- `GenerateIndexerExtensionMethods` (#5683): emits `Item`/`SetItem` (fixed identifiers, no escaping needed) and weaves in `indexer.Parameters[i].Name`. Those parameter names are already escaped at the model layer in `MemberDiscovery.cs` (lines 594/737/777/872), so no extra escaping required at the emit site.
- `*Of` extension method names from `ObjectMemberDisambiguations`: hardcoded constants, never collide with C# keywords.
- `using static TUnit.Mocks.SourceGenerator.IdentifierEscaping;` import preserved.

**Tests (net10.0):**
- `TUnit.Mocks.SourceGenerator.Tests`: 47/47 passed (all snapshots, including `Interface_With_Keyword_Member_Names` and the new `SelfEquatable_Generates_EqualsOf_GetHashCodeOf_ToStringOf`).
- `TUnit.Mocks.Tests` `KitchenSinkEdgeCasesTests/*` filter: 20/20 passed (T18 included).
- No `.received.txt` files generated.

PR #5690, opened 2026-04-24
perf(engine): collapse async state-machine layers on hot test path (#5687)

### LINKED ISSUE #5687: perf(engine): collapse async state-machine layers on hot test path
## Summary

Every test traverses 8+ stacked async state machines (TestRunner → TestCoordinator → RetryHelper → TimeoutHelper → TestExecutor → span wrapper → …). Sum of `AsyncMethodBuilder.Start` variants = **~50% inclusive CPU** in the profile. Several layers are pure overhead wrappers with no-op fast paths that should short-circuit the state machine entirely.

Baseline trace: `.profile/TUnit.Profile-20260424-011500/`.

### 1. Fast-path `RetryHelper` when `retryLimit == 0`

- Location: `TUnit.Engine/Services/TestExecution/TestCoordinator.cs:105–118` (the routing condition) and `TUnit.Engine/Services/TestExecution/RetryHelper.cs:10–59`
- `RetryHelper.ExecuteWithRetry` shows **8.31% inclusive CPU** even for tests with no retry attribute, because the current condition only skips `RetryHelper` when both `retryLimit == 0` AND `!testTimeout.HasValue`. If only a timeout is set, retry helper still runs.
- Fix: route directly past `RetryHelper` whenever `retryLimit == 0`, regardless of timeout.

### 2. Inline `TimeoutHelper.ExecuteWithTimeoutAsync` wrapper

- Location: `TUnit.Engine/Helpers/TimeoutHelper.cs:26–41`
- 4.65% inclusive CPU. The method exists only to adapt `Func<CancellationToken, Task>` → `Func<CancellationToken, Task<bool>>` and is `async` purely for the signature change — allocates a state machine + closure lambda per call.
- Fix: add a non-generic `Task`-returning overload of `ExecuteWithTimeoutCoreAsync` and call it directly; drop the outer `async` wrapper + `return true` lambda.

### 3. Collapse `ExecuteTestWithCompletionAsync` TCS wrapper

- Location: `TUnit.Engine/Scheduling/TestRunner.cs:65–77`
- **11.87% inclusive / ~0 exclusive** = 100% overhead. The method only awaits `ExecuteTestInternalAsync` then sets a `TaskCompletionSource`.
- Fix: attach a synchronous continuation to the returned `ValueTask` (or `AsTask()`) instead of making `ExecuteTestWithCompletionAsync` itself `async`.

### 4. Gate `DisposeTestInstanceWithSpanAsync` on `HasListeners()`

- Location: `TUnit.Engine/Services/TestExecution/TestCoordinator.cs:330–346`
- Always runs the OTEL span wrapper even when no listener is attached — allocates a `Func<Task>` delegate and state machine per test. The `TUnitActivitySource.Source.HasListeners()` guard is already used elsewhere (`TestExecutor.ExecuteAsync`).
- Fix: check `HasListeners()` before `RunWithSpanAsync`; call `DisposeTestInstanceCoreAsync` directly in the else branch.

## Acceptance criteria

- [ ] All 4 sub-tasks implemented
- [ ] Verify with `scripts/profile-tunit.ps1 -Top 50`:
  - `AsyncMethodBuilder.Start` inclusive drops from 19.96% → ~12%
  - `RetryHelper.ExecuteWithRetry` inclusive drops from 8.31% to <1%
  - `TimeoutHelper.ExecuteWithTimeoutAsync` frame gone (merged into core)
- [ ] Functional tests all pass

## Files

- `TUnit.Engine/Services/TestExecution/TestCoordinator.cs`
- `TUnit.Engine/Services/TestExecution/RetryHelper.cs`
- `TUnit.Engine/Helpers/TimeoutHelper.cs`
- `TUnit.Engine/Scheduling/TestRunner.cs`

thomhurst:
## Summary

Closes #5687.

Every test traverses 8+ stacked async state machines; four of those layers are pure overhead wrappers with no-op fast paths. This PR short-circuits them on the hot path.

- **RetryHelper fast path (TestCoordinator.cs)** — skip `RetryHelper.ExecuteWithRetry` whenever `retryLimit == 0`, regardless of whether a timeout is configured. Timeout is already applied per-attempt inside `TestExecutor.ExecuteAsync`, so the retry wrapper was pure overhead for any test with only a timeout attribute. Previously reported at ~8.3% inclusive CPU.
- **TimeoutHelper wrapper inlined (TimeoutHelper.cs)** — drop the outer `async` method + `return true` lambda whose only job was to adapt `Func<CT, Task>` to `Func<CT, Task<bool>>`. Fold the core straight into the non-generic `Task`-returning entry point; eliminates one state machine + closure allocation per timed test. Previously ~4.7% inclusive CPU.
- **TCS wrapper collapsed (TestRunner.cs)** — stop wrapping `ExecuteTestInternalAsync` in an outer `async` method whose only job was fulfilling a `TaskCompletionSource`. Start the inner `ValueTask`, fast-path synchronous completion into the TCS with no `Task` allocation, otherwise attach a synchronous continuation that mirrors `Faulted`/`Canceled`/`Completed` onto the TCS. Previously ~11.9% inclusive / ~0% exclusive — 100% wrapper overhead.
- **Dispose span gated (TestCoordinator.cs)** — check `TUnitActivitySource.Source.HasListeners()` before calling `RunWithSpanAsync` for test disposal, mirroring the gate already used in `TestExecutor.ExecuteAsync`. When no OTEL listener is attached, call `DisposeTestInstanceCoreAsync` directly and skip the per-test `Func<Task>` closure + state machine allocation.

## Test plan

- [x] `dotnet build TUnit.Engine/TUnit.Engine.csproj` — builds clean across `netstandard2.0;net8.0;net9.0;net10.0`, 0 warnings/errors.
- [x] `dotnet test TUnit.UnitTests/TUnit.UnitTests.csproj --framework net10.0` — 180/180 passed.
- [x] `dotnet test TUnit.Engine.Tests/TUnit.Engine.Tests.csproj --framework net10.0` — 146/146 non-subprocess tests passed, 99 skipped. The 2 failures (`VB.Test`, `FSharp.Test`) reproduce on `origin/main` and are pre-existing environmental issues with the VB/F# subprocess test harness — not caused by this change.
- [x] Targeted retry/timeout engine tests pass.

## Correctness notes

- **Timeout + retry**: timeout is still applied per attempt inside `TestExecutor.ExecuteAsync`; `RetryHelper` still wraps when `retryLimit > 0`, so tests configured with both timeout and retry retain prior semantics.
- **Exception propagation**: `ContinueWith` in `TestRunner` runs with `ExecuteSynchronously` and forwards `Faulted`/`Canceled`/`Completed` states onto the TCS; the returned `ValueTask(asTask)` surfaces the exception to the immediate awaiter. `DisposeTestInstanceCoreAsync` swallows exceptions internally, so the span-gate change does not alter observable behavior.

### DIFF
--- a/TUnit.Engine/Helpers/TimeoutHelper.cs
+++ b/TUnit.Engine/Helpers/TimeoutHelper.cs
@@ -29,35 +29,14 @@ internal static class TimeoutHelper
         CancellationToken cancellationToken,
         string? timeoutMessage = null)
     {
-        await ExecuteWithTimeoutCoreAsync<bool>(
-            async ct =>
-            {
-                await taskFactory(ct).ConfigureAwait(false);
-                return true;
-            },
-            timeout,
-            cancellationToken,
-            timeoutMessage).ConfigureAwait(false);
-    }
-
-    /// <summary>
-    /// Core timeout implementation. Only called when a timeout is actually configured.
-    /// Allocates CancellationTokenSource, TaskCompletionSource, and uses Task.WhenAny.
-    /// </summary>
-    private static async Task<T> ExecuteWithTimeoutCoreAsync<T>(
-        Func<CancellationToken, Task<T>> taskFactory,
-        TimeSpan timeout,
-        CancellationToken cancellationToken,
-        string? timeoutMessage)
-    {
         // Timeout path: create linked token so task can observe both timeout and external cancellation.
         using var timeoutCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
 
         // Set up cancellation detection BEFORE scheduling timeout to avoid race condition
         // where timeout fires before registration completes (with very small timeouts)
-        var cancelledTcs = new TaskCompletionSource<T>(TaskCreationOptions.RunContinuationsAsynchronously);
+        var cancelledTcs = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
         using var registration = timeoutCts.Token.Register(
-            static state => ((TaskCompletionSource<T>)state!).TrySetCanceled(),
+            static state => ((TaskCompletionSource<bool>)state!).TrySetCanceled(),
             cancelledTcs);
 
         // Now schedule the timeout - registration is guaranteed to catch it
@@ -102,6 +81,6 @@ internal static class TimeoutHelper
             throw new TimeoutException(diagnosticMessage);
         }
 
-        return await executionTask.ConfigureAwait(false);
+        await executionTask.ConfigureAwait(false);
     }
 }
--- a/TUnit.Engine/Scheduling/TestRunner.cs
+++ b/TUnit.Engine/Scheduling/TestRunner.cs
@@ -60,8 +60,9 @@ public sealed class TestRunner
         }
 
         // Skip the extra async state machine that a wrapper method would create. Start the
-        // inner ValueTask, fast-path synchronous completion into the TCS, and otherwise attach
-        // a synchronous continuation that mirrors the outcome onto the TCS.
+        // inner ValueTask, fast-path synchronous completion into the TCS, and otherwise fall
+        // back to a minimal async helper that mirrors the outcome onto the TCS without
+        // allocating a Task via AsTask().
         var innerTask = ExecuteTestInternalAsync(test, cancellationToken);
 
         if (innerTask.IsCompletedSuccessfully)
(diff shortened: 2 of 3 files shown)


thomhurst:
Addressed review feedback:

**WrapAsync refactor** (1ae4da7da)
- Replaced `AsTask()` + `ContinueWith` plumbing with a minimal `WrapAsync` helper.
- No `.AsTask()` allocation on the async-completion path.
- Restores original Faulted-vs-Canceled exception semantics (`SetException` for all non-success outcomes).
- Sync fast path remains zero-alloc.
- Outer `ValueTask` still observes failure because we rethrow after `SetException`.

**CI `modularpipeline` (ubuntu/macos) failure**
Not caused by this PR. Root cause is a snapshot mismatch in `TUnit.Mocks.SourceGenerator.Tests` (`Class_Implementing_Static_Abstract_Interface`) that reproduces on `main` at the same commit (see run 24864304955 on `main`, which fails on all three OSes with the identical `Snapshot mismatch` error). This is pre-existing, related to the recent MockBridge/static-abstract work, and independent of the async engine changes here.

Local `TUnit.Engine.Tests` (net10.0) and `TUnit.UnitTests` (net10.0) both green after the refactor (247 + 180 tests respectively).

PR #5691, opened 2026-04-24
chore(mocks): regenerate source generator snapshots

thomhurst:
## Summary

Regenerate the TUnit.Mocks source-generator snapshot baselines to match the current generator output after the recent mocks fixes (#5677 / #5678 / #5679 / #5680 / #5681 / #5682 / #5683 / #5684).

## Changes

- 1 snapshot file updated: `Class_Implementing_Static_Abstract_Interface.verified.txt`
- The drift is a single-line addition of `[global::System.Diagnostics.CodeAnalysis.SetsRequiredMembers]` on the generated mock constructor — directly matching the intent of #5678 ("fix(mocks): emit [SetsRequiredMembers] on generated mock ctor").

## Verification

- Spot-checked the diff: legitimate generator output change, not a bug. Generated C# is well-formed.
- Before regen: 48 passed / 1 failed (on net10.0).
- After regen: 49 passed / 0 failed.

## Test plan

- [x] `dotnet build TUnit.Mocks.SourceGenerator.Tests/TUnit.Mocks.SourceGenerator.Tests.csproj`
- [x] `dotnet run --framework net10.0 --no-build` in `TUnit.Mocks.SourceGenerator.Tests/` — all 49 tests pass
- [x] Only `.verified.txt` files touched; no `.received.txt` committed

### DIFF
--- a/TUnit.Mocks.SourceGenerator.Tests/Snapshots/Class_Implementing_Static_Abstract_Interface.verified.txt
+++ b/TUnit.Mocks.SourceGenerator.Tests/Snapshots/Class_Implementing_Static_Abstract_Interface.verified.txt
@@ -10,6 +10,7 @@ namespace TUnit.Mocks.Generated
         [global::System.ComponentModel.EditorBrowsable(global::System.ComponentModel.EditorBrowsableState.Never)]
         global::TUnit.Mocks.IMock? global::TUnit.Mocks.IMockObject.MockWrapper { get; set; }
 
+        [global::System.Diagnostics.CodeAnalysis.SetsRequiredMembers]
         internal StaticAbstractImplMockImpl(global::TUnit.Mocks.MockEngine<global::StaticAbstractImpl> engine) : base()
         {
             _engine = engine;


thomhurst:
The windows-latest modularpipeline failure on this PR is **unrelated to this PR's changes** (which only regenerate one `.verified.txt` snapshot file).

### Root cause
`TUnit.Assertions.Tests.ExecutionTimeTests.Completes_Within_Unhappy` (net8.0) failed with a flaky timing assertion:

```
AssertionException: Expected to be less than 00:00:01
but received 00:00:02.3657737
  at Assert.That(duration).IsLessThan(TimeSpan.FromSeconds(1))
  at TUnit.Assertions.Tests.ExecutionTimeTests.Completes_Within_Unhappy() in ExecutionTimeTests.cs:30
```

The test launches an action that calls `Thread.Sleep(10000)`, asserts a 250ms timeout fires as expected, then asserts the total stopwatch elapsed is `< 1 second`. On a slow/contended Windows runner, cleanup from the sleeping thread (or CI scheduling jitter) can exceed 1s. This is a pre-existing flaky test, not introduced by this PR.

### Evidence it is pre-existing
The most recent main run (24864304955) also failed its modularpipeline jobs. On main, all three OS runners fail `RunMockSourceGeneratorTestsModule` on the `Class_Implementing_Static_Abstract_Interface` snapshot — **which is exactly what this PR fixes**. On this PR, the `RunMockSourceGeneratorTestsModule` step passes on every OS, and only windows-latest hit the unrelated flake.

### Recommendation
This PR can be merged as-is. The `Completes_Within_Unhappy` test should be loosened (or re-run on failure) in a follow-up.

PR #5723, opened 2026-04-24
perf(engine): cache non-final TestNode PropertyBag across state transitions

### LINKED ISSUE #5717: perf: cache non-final TestNode PropertyBag instead of rebuilding per state transition
## Summary
`ToTestNode` is called **3 times per test** (Discovered, InProgress, terminal) and builds a full `PropertyBag` every time. Discovered and InProgress states are immutable for a given `TestId`, so the TestNode can be cached. Estimated **~1% CPU, thousands of PropertyBag allocations removed**.

## Evidence
- `TUnit.Engine/TUnitMessageBus.cs:37, 45, 56, 77` — three non-final calls per test lifecycle.
- `TUnit.Engine/Extensions/TestExtensions.cs:115-217` — builds `PropertyBag` with FileLocation, MethodIdentifier, categories, custom props every call.
- Trace: `TUnitMessageBus.InProgress` 1.11% inclusive / 0.15% self, plus `TestExtensions+<>c.<GetOrCreateCachedProperties>b__6_0` 0.19% self.

## Proposed fix
- Cache the `TestNode` per `(TestId, state)` for the two non-final states, where the state property is already a cached singleton.
- Store the cached TestNode on `TestContext` (one allocation per test instead of three).
- Leave terminal-state TestNode creation as-is (result data varies).

## Expected impact
- ~1% CPU.
- ~2 PropertyBag allocations per test avoided (~2K on the profile workload).
- Less pressure on MTP message bus publish path.

thomhurst:
## Summary
- `ToTestNode` was called three times per test lifecycle (Discovered, InProgress, terminal) and rebuilt a fresh `PropertyBag` every time. The Discovered and InProgress nodes carry immutable data for a given `TestId`, so two of those allocations were pure waste.
- Cache the built `TestNode` on `TestContext` keyed by the two non-final singleton state properties (`DiscoveredTestNodeStateProperty.CachedInstance`, `InProgressTestNodeStateProperty.CachedInstance`). Subsequent reporter calls with the same singleton state return the cached instance instead of reconstructing the `PropertyBag`.
- Terminal-state `TestNode`s still build fresh because timing, output, TRX, and exception data vary by result.
- The cache fields live on `TestContext` typed as `object?` to avoid pulling a `Microsoft.Testing.Platform` type reference into `TUnit.Core`, matching the existing `AbstractExecutableTest.CachedPropertyBag` pattern.

Expected impact: ~1% CPU, ~2 PropertyBag allocations removed per test (~2K on the profile workload).

## Test plan
- [x] `dotnet run -c Release --framework net10.0 -- --treenode-filter "/*/*/BasicTests/*"` in `TUnit.TestProject` — 3/3 passed
- [x] `dotnet run -c Release --framework net10.0` in `TUnit.Core.SourceGenerator.Tests` — 116/117 passed (1 pre-existing skip)

Closes #5717

### DIFF
--- a/TUnit.Core/TestContext.cs
+++ b/TUnit.Core/TestContext.cs
@@ -425,6 +425,14 @@ public partial class TestContext : Context,
 
     internal ConcurrentDictionary<string, object?> ObjectBag => _testBuilderContext.StateBag;
 
+    // Cached TestNode instances for the two non-final lifecycle states (Discovered, InProgress).
+    // Both states carry immutable data for a given TestId, so the built TestNode (including
+    // its PropertyBag) can be reused across the two reporter calls per test. Terminal-state
+    // TestNodes vary by result and are built fresh each time.
+    // Using object? to avoid a type dependency on Microsoft.Testing.Platform from TUnit.Core.
+    internal object? CachedDiscoveredTestNode { get; set; }
+    internal object? CachedInProgressTestNode { get; set; }
+
     internal AbstractExecutableTest InternalExecutableTest { get; set; } = null!;
 
     internal SortedList<int, HashSet<object>> TrackedObjects { get; } = new();
--- a/TUnit.Core/TestContext.cs
+++ b/TUnit.Core/TestContext.cs
@@ -425,14 +425,6 @@ public partial class TestContext : Context,
 
     internal ConcurrentDictionary<string, object?> ObjectBag => _testBuilderContext.StateBag;
 
-    // Cached TestNode instances for the two non-final lifecycle states (Discovered, InProgress).
-    // Both states carry immutable data for a given TestId, so the built TestNode (including
-    // its PropertyBag) can be reused across the two reporter calls per test. Terminal-state
-    // TestNodes vary by result and are built fresh each time.
-    // Using object? to avoid a type dependency on Microsoft.Testing.Platform from TUnit.Core.
-    internal object? CachedDiscoveredTestNode { get; set; }
-    internal object? CachedInProgressTestNode { get; set; }
-
     internal AbstractExecutableTest InternalExecutableTest { get; set; } = null!;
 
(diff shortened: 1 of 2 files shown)


thomhurst:
You're right — I traced every `ToTestNode` call site and the cache never hits:

- `Discovered`: called once per test in `TestRequestHandler.HandleDiscoveryRequestAsync` and explicitly skipped in `HandleRunRequestAsync` (line 68-69: "Skip sending Discovered messages during execution").
- `InProgress`: called once per test in `TestCoordinator.ExecuteTestInternalAsync` **before** the retry loop. `RetryHelper.ExecuteWithRetry` wraps only `ExecuteTestLifecycleAsync`, not the InProgress publish.
- `RepeatAttribute`: `TestBuilder` constructs a separate `AbstractExecutableTest` + `TestContext` per repeat index — no shared context.
- IDE discover-then-run: `TestDiscoveryService.DiscoverTests` runs the builder pipeline on every call, producing fresh `TestContext` instances. The `_cachedTests` queue is only used by `TestFinder` for lookup, not reused by subsequent `DiscoverTests` calls.

So every `(TestContext, state-property)` pair is seen exactly once. The cache was pure overhead — two object fields per TestContext plus two `ReferenceEquals` checks and a store on every `ToTestNode` call with no payoff.

Pre-building at discovery time is equivalent work (one `PropertyBag` + one `TestNode` per state), so there's no alternative win to salvage here either. Reverted the change in 94096508d.

Thanks for the careful review — should have traced the call graph before claiming savings.

thomhurst:
Closing — investigation (see review thread) confirmed the cache never hits in any real lifecycle path (Discovered called once per test in discovery flow; InProgress called once per test pre-retry-loop; RepeatAttribute builds distinct TestContext instances). Pre-build alternative offers no dedup opportunity. Reverted; branch now zero-diff vs main. Closing #5717 resolved-by-investigation.

PR #5724, opened 2026-04-24
perf(engine): skip Console.Out/Err FlushAsync when no output captured (#5712)

### LINKED ISSUE #5712: perf: skip Console.Out/Err FlushAsync when no output captured
## Summary
Every test unconditionally awaits two `FlushAsync` calls in the `finally` block, producing two extra async state machines per test for the common case where nothing was written to stdout/stderr. Estimated **~1.5-2% CPU reduction, -2 state machines per test**.

## Evidence
`TUnit.Engine/Services/TestExecution/TestCoordinator.cs:141-151`:
```csharp
finally
{
    await Console.Out.FlushAsync();
    await Console.Error.FlushAsync();
    ...
}
```

CPU-sampling trace shows `AsyncMethodBuilderCore.Start` at 16.39% inclusive / 2.93% self — test execution dispatches dozens of state machines per test, and these two are unconditionally created even when output redirection captured nothing.

## Proposed fix
- Have the output interceptor set a `HasCapturedOutput` flag on the context when it first writes.
- Only flush when the flag is set.

## Expected impact
- ~1.5-2% CPU.
- Removes 2 async state machines per test.

thomhurst:
Closes #5712

## Summary
- Adds a volatile `HasCapturedConsoleOutput` flag on `Context`, set by the console interceptor on any write path.
- `TestCoordinator` checks the flag before awaiting `Console.Out.FlushAsync()` / `Console.Error.FlushAsync()` in the per-test finally block, skipping both calls when nothing was written.
- Removes 2 async state machines per passing test with no output; CPU-sampling trace showed `AsyncMethodBuilderCore.Start` at 16.39% inclusive / 2.93% self — expected ~1.5-2% CPU reduction.

## Correctness
- The flag is monotonic (false -> true) and set *before* the buffer append, so any write that lands triggers the flush.
- The fire-and-forget race (`Task.Run(() => Console.WriteLine(...))` without awaiting) already exists under the old unconditional flush — the flag preserves identical semantics: when the Task.Run actually writes it flips the flag and is observed by the next test's finally or the test session cleanup path.
- Partial writes (Console.Write without newline) are covered — they hit `GetLineBufferForWrite` which marks the flag before appending to the buffer.
- Volatile read/write (no lock, no allocation) on the hot path.

## Test plan
- [x] `dotnet build TUnit.slnx -c Release` succeeds (0 errors, pre-existing warnings only).
- [x] `BasicTests` smoke — 3/3 passed.
- [x] `CaptureOutputTests` (Console.WriteLine + GetStandardOutput assertions, `[Repeat(15)]`) — 80/80 passed.
- [x] `TestBuildContextOutputCaptureTests` — 7/7 passed.
- [x] `ParallelConsoleOutputTests` — 44/44 passed.
- [x] `OutputTruncationTests` — 3/3 passed.

## Risks
- If a test writes only during an async continuation that resumes *after* the coordinator's finally runs, output is lost — but this is already the case under the old code (`FlushAsync` is still a point-in-time call), so no regression.
- The flag lives on `Context` (base class), not `TestContext`; writes from non-test contexts (hooks) also set it, which is harmless since the flag is per-context.

### DIFF
--- a/TUnit.Core/Context.cs
+++ b/TUnit.Core/Context.cs
@@ -35,6 +35,11 @@ public abstract class Context : IContext, IDisposable
     private ConsoleLineBuffer? _consoleStdOutLineBuffer;
     private ConsoleLineBuffer? _consoleStdErrLineBuffer;
 
+    // Set by the console interceptor on first write so TestCoordinator can skip the
+    // two Console.Out/Err FlushAsync state machines per test when nothing was written.
+    // Volatile read/write is cheap and sufficient — the flag only ever transitions false -> true.
+    private int _consoleOutputCaptured;
+
     // Thread-safe: console interceptors may access from multiple threads.
     private StringBuilder GetOutputBuilder() =>
         LazyInitializer.EnsureInitialized(ref _outputBuilder)!;
@@ -62,6 +67,10 @@ public abstract class Context : IContext, IDisposable
 
     internal ConsoleLineBuffer ConsoleStdErrLineBuffer =>
         LazyInitializer.EnsureInitialized(ref _consoleStdErrLineBuffer)!;
+
+    internal bool HasCapturedConsoleOutput => Volatile.Read(ref _consoleOutputCaptured) != 0;
+
+    internal void MarkConsoleOutputCaptured() => Volatile.Write(ref _consoleOutputCaptured, 1);
 
     internal Context(Context? parent)
     {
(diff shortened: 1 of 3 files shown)


thomhurst:
Addressed the three review suggestions:

1. Replaced `int _consoleOutputCaptured` + `Volatile.Read/Write` with `private volatile bool _consoleOutputCaptured` for clarity (no CAS was needed).
2. Added a WHY comment on `GetLineBuffer()` noting it's for Flush/Close only — write paths must use `GetLineBufferForWrite()` so draining the buffer doesn't spuriously set `HasCapturedConsoleOutput`.
3. Added `ConsoleOutputCaptureFlagTests` in `TUnit.UnitTests` covering the three requested scenarios (fresh context false; after `Write("x")` true; after empty `WriteLine()` true) plus the direct `MarkConsoleOutputCaptured` path. All 4 tests pass on net10.0.

PR #5725, opened 2026-04-24
perf(engine): collapse async forwarding wrappers in test execution (#5714)

### LINKED ISSUE #5714: perf: collapse 4-layer async forwarding chain in test execution
## Summary
The per-test entry point chains through 4 async layers before any real work happens, each materializing its own state machine. Estimated **2 state machines per test removed**.

## Evidence
Trace (inclusive %):
```
11.54% TestRunner.ExecuteTestAsync
11.28% TestRunner.ExecuteTestInternalAsync
11.10% TestCoordinator.ExecuteTestAsync           ← pure forward
11.10% TestCoordinator.ExecuteTestInternalAsync
10.13% TestCoordinator.ExecuteTestLifecycleAsync
 9.94% TestExecutor.ExecuteAsync
```

`TUnit.Engine/Services/TestExecution/TestCoordinator.cs:50-51` is a pure forwarding async method:
```csharp
public ValueTask ExecuteTestAsync(AbstractExecutableTest test, CancellationToken cancellationToken)
    => ExecuteTestInternalAsync(test, cancellationToken);
```

`ExecuteTestLifecycleAsync` is an inline-able wrapper on the no-retry branch.

## Proposed fix
- Inline pure-forward `TestCoordinator.ExecuteTestAsync` into its caller.
- Inline `ExecuteTestLifecycleAsync` into `ExecuteTestInternalAsync` for the no-retry path.
- Audit `TestRunner.ExecuteTestAsync` → `ExecuteTestInternalAsync` for the same pattern.

## Expected impact
- 2-3 async state machines per test removed.
- Directly reduces `AsyncMethodBuilderCore.Start` 2.93% self time observed in the trace.

## Files
- `TUnit.Engine/Services/TestExecution/TestCoordinator.cs:50-51, 53, 279`
- `TUnit.Engine/Scheduling/TestRunner.cs:48, 91, 125`
- `TUnit.Engine/TestExecutor.cs:69`

thomhurst:
## Summary
- Collapses the pure-forward `TestCoordinator.ExecuteTestAsync` → `ExecuteTestInternalAsync` wrapper into a single method, eliminating one async state machine per test.
- Inlines the no-retry `ExecuteTestLifecycleAsync` body directly into `ExecuteTestAsync`. The retry branch still dispatches through `ExecuteTestLifecycleAsync` because `RetryHelper.ExecuteWithRetry` requires a `Func<ValueTask>`.
- Net effect on the hot no-retry path: ~2 async state machines removed per test (consistent with the profiler evidence in #5714).

`TestRunner.ExecuteTestAsync` was also audited. It is not a pure forward — it owns the session-wide dedup ledger (`ConcurrentDictionary<string, TCS<bool>>`) and already uses a minimal `WrapAsync` helper that only spins a state machine on the non-synchronous path, so it is already optimized and left alone.

`TestExecutor.ExecuteAsync` was audited and performs substantial real work (hook orchestration, activity lifecycle, timeout wiring, exception aggregation) — not a forwarder; left alone.

## Risks
- Intentional duplication: the no-retry inlined body is a verbatim copy of `ExecuteTestLifecycleAsync`. Both locations carry cross-reference comments so future edits stay in sync. If the lifecycle gets any more complex, consider reverting to a single method and accepting the state-machine cost, or refactoring into AOT-safe non-async helpers.
- Retry, cancellation, and exception semantics are preserved exactly (skip checks run in the same order as before; dispose-with-span still fires in `finally`).

## Test plan
- [x] `dotnet build TUnit.slnx -c Release` — succeeds (0 errors).
- [x] `TUnit.Core.SourceGenerator.Tests` snapshot tests on net10.0 — 116 passed, 1 skipped (pre-existing).
- [x] `TUnit.TestProject` `BasicTests` smoke — 3/3 passed.
- [x] `TUnit.TestProject` `RetryTests` — retry counts exactly match expectations (`AssertCounts` passes; 3 intentional failure tests fail after the expected attempt counts: 2, 3, 4 — confirming retry wrapper path still works).
- [x] `TUnit.TestProject` `DependsOnTests` — 2/2 passed (dependency-chain path via `TestRunner`).

Closes #5714

### DIFF
--- a/TUnit.Engine/Services/TestExecution/TestCoordinator.cs
+++ b/TUnit.Engine/Services/TestExecution/TestCoordinator.cs
@@ -47,10 +47,10 @@ internal sealed class TestCoordinator : ITestCoordinator
     // Dedup happens in TestRunner via its own ConcurrentDictionary<string, TCS<bool>> —
     // it's the single entry point for both scheduler and dependency recursion, so a second
     // guard here would just double the TCS/dict allocations per test.
-    public ValueTask ExecuteTestAsync(AbstractExecutableTest test, CancellationToken cancellationToken)
-        => ExecuteTestInternalAsync(test, cancellationToken);
-
-    private async ValueTask ExecuteTestInternalAsync(AbstractExecutableTest test, CancellationToken cancellationToken)
+    //
+    // Note: the previous pure-forward ExecuteTestAsync → ExecuteTestInternalAsync wrapper
+    // was collapsed to eliminate one async state machine per test (#5714).
+    public async ValueTask ExecuteTestAsync(AbstractExecutableTest test, CancellationToken cancellationToken)
     {
         try
         {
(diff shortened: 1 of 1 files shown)


thomhurst:
Applied both non-blocking review suggestions in ea534fa5b:

- Extracted `HandleSkipAsync` private helper (mark state + skipped receivers + end receivers) and routed all three skip sites through it (pre-registration, fast-path pre-instance, fast-path post-instance + the two matching sites in `ExecuteTestLifecycleAsync`).
- Converted the inlined no-retry fast path to early-return style so its body now mirrors `ExecuteTestLifecycleAsync` shape for shape, making the "must stay in sync" maintenance contract trivial to verify by diff.

Semantics preserved: `MarkCompleted` is idempotent on an already-skipped test (`MarkSkipped` set `test.Result`, so `Result ??=` is a no-op), so the early `return` in the fast-path skip branches is observationally equivalent to the previous fall-through via `MarkCompleted`. Retry, cancellation, and exception handling paths are unchanged.

Validation:
- `dotnet build TUnit.slnx -c Release` - 0 errors
- `TUnit.Core.SourceGenerator.Tests` snapshots - 116/117 passed, 1 skipped (pre-existing flaky)
- `BasicTests/*` smoke - 3/3 passed
- Skip-related suites (`SkipTests`, `SimpleSkipTest`, `SkipConstructorTest`, `SkipExceptionFixTest`, `DynamicSkipReasonTests`, `RunOnSkipTests`, `SkippedEventReceiverTests`, `RuntimeSkipEventReceiverTests`) - 24 passed, 1 intentionally skipped
- Retry suites (`CustomRetryTests`, `ScopedRetryAttributePriorityTests`, `SimpleRetryPriorityTest`) - failures match baseline on the pre-refactor commit (designed-to-fail tests that throw intentionally; retry-count `AssertCounts` passes)

PR #5727, opened 2026-04-24
perf: eliminate per-test closure + GetOrAdd factory alloc (#5710)

### LINKED ISSUE #5710: perf: eliminate per-test closure + GetOrAdd factory alloc in ToTestMetadata
## Summary
Top user-code CPU hit in CPU-sampling trace: **2.62% inclusive / 2.11% self** on `<ToTestMetadata>b__80_0`. Runs once per test.

## Evidence
From `dotnet-trace` CPU sampling of `TUnit.Profile` (1013 tests, net10.0, Release):

```
8. __Canon].<ToTestMetadata>b__80_0(...)          2.62%   2.11%   (top user-code self time)
9. __Canon].GetOrAdd(!0, class Func`2<!0,!1>)     1.36%   1.08%
```

## Root cause
`TUnit.Core/TestEntry.cs:110-113`:
```csharp
Interlocked.CompareExchange(ref _cachedInvokeTypedTest,
    (instance, args, ct) => InvokeBody(instance, MethodIndex, args, ct), null);
```

- Lambda captures `this` → compiler emits `<>c__DisplayClass80_0` closure allocated every time ToTestMetadata runs.
- The closure is then invoked once per test, doing nothing but forwarding to a static switch.
- Related: `ConcurrentDictionary.GetOrAdd(key, Func<K,V>)` also allocates a non-static `Func` closure on every call — even on cache hits — contributing the 1.08% in entry 9.

## Proposed fix
- Store `InvokeBody` (already a static delegate) and `MethodIndex` directly on `TestMetadata<T>` / `ExecutableTest<T>`. Invoke as `_metadata.InvokeBody!((T)instance, _metadata.MethodIndex, args, ct)` — no closure.
- Replace `dict.GetOrAdd(key, Func<K,V>)` with `dict.GetOrAdd(key, static (k, state) => ..., state)` across the hot paths so cache-hit path stops allocating factories.

## Expected impact
~2-3% CPU reduction per test plus significant per-test heap allocation removal.

## Files
- `TUnit.Core/TestEntry.cs:107-115`
- `TUnit.Core/TestMetadata\`1.cs`
- `TUnit.Core/ExecutableTest\`1.cs`

thomhurst:
Closes #5710

## Summary

CPU sampling (`dotnet-trace`, 1013 tests, net10.0, Release) flagged `<ToTestMetadata>b__80_0` as the top user-code self-time hit at 2.62% inclusive / 2.11% self. The lambda inside `TestEntry<T>.ToTestMetadata`

```csharp
(instance, args, ct) => InvokeBody(instance, MethodIndex, args, ct)
```

captures `this`, so the compiler emits a `<>c__DisplayClass80_0` closure object on the first call for every entry, and the cached delegate is then invoked once per test. The matching `() => CreateAttributes(AttributeGroupIndex)` closure adds a second per-entry allocation and hot-path indirection.

This change forwards the class-shared static delegates (`InvokeBody`, `CreateAttributes`) and their per-test indexes (`MethodIndex`, `AttributeGroupIndex`) directly onto `TestMetadata<T>` via new fields:

- `TestMetadata<T>.IndexedInvokeBody` + `MethodIndex`
- `TestMetadata.IndexedAttributeFactory` + `AttributeGroupIndex`

`ExecutableTest<T>.InvokeTestAsync` and `TestMetadata.GetOrCreateAttributes` prefer the indexed form, falling back to the legacy `InvokeTypedTest` / `AttributeFactory` delegates used by the per-method source-gen path and reflection mode. `AttributeFactory` loses `required` so each TestEntry-backed metadata only needs to populate one form; existing call sites that set `AttributeFactory` (reflection discovery, unit tests, `TestMetadataFactory.Create`) continue to work unchanged.

## Impact

- Zero closure/delegate allocations per `ToTestMetadata` call on the TestEntry path — the `<>c__DisplayClass` is gone and the static class-level delegates flow through by reference.
- One call-site dispatch in `ExecutableTest<T>.InvokeTestAsync` (`indexed((T)instance, MethodIndex, args, ct)`) with no captured state.
- Reflection mode, generic-type resolution (`GenericTestMetadata`), and the per-method source-gen path (`TestMetadataFactory.Create`) are untouched and continue to set `InvokeTypedTest`/`AttributeFactory`.

## Verification

- `dotnet build TUnit.slnx -c Release` — 0 errors, pre-existing warnings only.
- `TUnit.Core.SourceGenerator.Tests` — 116 passed, 1 skipped (pre-existing).
- `TUnit.TestProject --treenode-filter "/*/*/BasicTests/*"` — 3/3 passed (source-gen mode).
- `TUnit.PublicAPI` — 4/4 passed on net10; verified files for net9 / net8 / net472 updated consistently.

## Test plan

- [ ] CI runs public-API tests on net9 / net8 / net472 to confirm those verified files are correct
- [ ] CI runs full test suite in both source-generated and reflection modes
- [ ] Repeat CPU sampling on TUnit.Profile to confirm `ToTestMetadata` is no longer in the top self-time hits

### DIFF
--- a/TUnit.Core/ExecutableTest`1.cs
+++ b/TUnit.Core/ExecutableTest`1.cs
@@ -66,6 +66,13 @@ internal sealed class ExecutableTest<
 
     public override async Task InvokeTestAsync(object instance, CancellationToken cancellationToken)
     {
+        var indexed = _metadata.IndexedInvokeBody;
+        if (indexed is not null)
+        {
+            await indexed((T)instance, _metadata.MethodIndex, Arguments, cancellationToken).ConfigureAwait(false);
+            return;
+        }
+
         await _metadata.InvokeTypedTest!((T)instance, Arguments, cancellationToken).ConfigureAwait(false);
     }
 }
--- a/TUnit.Core/TestEntry.cs
+++ b/TUnit.Core/TestEntry.cs
@@ -107,13 +107,6 @@ public sealed class TestEntry<
     private PropertyDataSource[]? _cachedPropertyDataSources;
     private PropertyInjectionData[]? _cachedPropertyInjections;
 
-    // Satisfies TestMetadata's `required` AttributeFactory without a per-test allocation:
-    // TestEntry-sourced metadata always takes the IndexedAttributeFactory path, so this throws
-    // if ever invoked — reaching it indicates an engine bug.
-    private static readonly Func<Attribute[]> s_indexedAttributeFactoryPlaceholder =
-        static () => throw new InvalidOperationException(
-            "TestEntry metadata must resolve attributes via IndexedAttributeFactory.");
-
     internal TestMetadata<T> ToTestMetadata(string testSessionId)
     {
         return new TestMetadata<T>
@@ -129,7 +122,7 @@ public sealed class TestEntry<
             InstanceFactory = CreateInstance,
             IndexedInvokeBody = InvokeBody,
             MethodIndex = MethodIndex,
-            AttributeFactory = s_indexedAttributeFactoryPlaceholder,
+            AttributeFactory = TestEntrySentinel.IndexedAttributeFactoryPlaceholder,
             IndexedAttributeFactory = CreateAttributes,
             AttributeGroupIndex = AttributeGroupIndex,
             FilePath = FilePath,
@@ -176,4 +169,16 @@ public sealed class TestEntry<
         }
         return result;
     }
+}
+
+// Non-generic holder so the placeholder delegate is shared across every closed TestEntry<T> —
+// a static field on the generic type would otherwise be duplicated per closed type (Sonar S2743).
+internal static class TestEntrySentinel
+{
+    // Satisfies TestMetadata's `required` AttributeFactory without a per-test allocation:
+    // TestEntry-sourced metadata always takes the IndexedAttributeFactory path, so this throws
+    // if ever invoked — reaching it indicates an engine bug.
+    internal static readonly Func<Attribute[]> IndexedAttributeFactoryPlaceholder =
+        static () => throw new InvalidOperationException(
+            "TestEntry metadata must resolve attributes via IndexedAttributeFactory.");
 }
--- a/TUnit.Core/TestEntry.cs
+++ b/TUnit.Core/TestEntry.cs
@@ -107,6 +107,13 @@ public sealed class TestEntry<
     private PropertyDataSource[]? _cachedPropertyDataSources;
     private PropertyInjectionData[]? _cachedPropertyInjections;
 
+    // Satisfies TestMetadata's `required` AttributeFactory without a per-test allocation:
+    // TestEntry-sourced metadata always takes the IndexedAttributeFactory path, so this throws
+    // if ever invoked — reaching it indicates an engine bug.
+    private static readonly Func<Attribute[]> s_indexedAttributeFactoryPlaceholder =
+        static () => throw new InvalidOperationException(
+            "TestEntry metadata must resolve attributes via IndexedAttributeFactory.");
+
     internal TestMetadata<T> ToTestMetadata(string testSessionId)
     {
         return new TestMetadata<T>
@@ -122,6 +129,7 @@ public sealed class TestEntry<
             InstanceFactory = CreateInstance,
             IndexedInvokeBody = InvokeBody,
             MethodIndex = MethodIndex,
+            AttributeFactory = s_indexedAttributeFactoryPlaceholder,
             IndexedAttributeFactory = CreateAttributes,
             AttributeGroupIndex = AttributeGroupIndex,
             FilePath = FilePath,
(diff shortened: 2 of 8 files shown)


thomhurst:
Pushed 726910b addressing review feedback:

- **Public API growth**: `IndexedAttributeFactory`, `AttributeGroupIndex`, `IndexedInvokeBody`, and `MethodIndex` are now `internal` — only set by `TestEntry.ToTestMetadata()` and read from `ExecutableTest<T>`/`GetOrCreateAttributes`, all inside `TUnit.Core`. PublicAPI verified snapshots shrunk back to the pre-PR shape (save for the negligible `Field -> MethodIndex` removal).
- **Lost `required` guarantee**: `AttributeFactory` reverted to `required`. `TestEntry` now passes a single static throwing sentinel (`s_indexedAttributeFactoryPlaceholder`) to satisfy the compile-time constraint without a per-test allocation; the indexed path always supersedes it, so the sentinel is never actually invoked at runtime.
- **Thread safety**: `GetOrCreateAttributes` now uses `Interlocked.CompareExchange` with a short WHY comment. The benign race (factories are idempotent; a contended double-invoke is harmless) is now safe-by-construction rather than unexplained.
- **Trailing blank lines** trimmed from `TestMetadata\`1.cs`.
- Source-gen snapshot tests (`TUnit.Core.SourceGenerator.Tests`), `BasicTests` smoke run, and `TUnit.PublicAPI` snapshot tests all pass; full `TUnit.slnx` Release build clean (0 errors).

Source generator emits `MethodIndex`/`AttributeGroupIndex` only on the public `TestEntry<T>` — never on `TestMetadata` directly — so the internal-ification is safe without `InternalsVisibleTo` changes to user assemblies.

thomhurst:
Fixed the 2 Codacy issues flagged on `726910b3`:

**1. `TUnit.Core/TestEntry.cs:113` — SonarCSharp_S2743 (ErrorProne)**
> A static field in a generic type is not shared among instances of different close constructed types.

The `s_indexedAttributeFactoryPlaceholder` delegate had no dependency on `T`, so hosting it on `TestEntry<T>` duplicated the field per closed generic type. Moved it to a new non-generic `TestEntrySentinel` holder class so a single instance is shared across every closed `TestEntry<T>`. Behavior is unchanged — it still throws if ever invoked, and the comment explaining the WHY (IndexedAttributeFactory is the real path) is preserved.

**2. ``TUnit.Core/TestMetadata`1.cs:94`` — SonarCSharp_S2372 (BestPractice)**
> Remove the exception throwing from this property getter, or refactor the property into a method.

The abstract base declares `CreateExecutableTestFactory` as a property, so it has to stay a property — refactoring to a method would be a public API break. Extracted the throw into a `[DoesNotReturn]` helper `ThrowMissingInvoker()`; the getter now contains no lexical `throw`. The T-qualified error message is preserved because the helper is itself a static member of `TestMetadata<T>`.

Neither fix is a suppression — both address the underlying complaint structurally. Verified green:
- `dotnet build TUnit.slnx -c Release` — 0 errors
- `TUnit.Core.SourceGenerator.Tests` net10.0 — 116 passed / 1 skipped
- `TUnit.TestProject` net10.0 `BasicTests/*` smoke filter — 3/3 passed

Commit: 883e5bad4

thomhurst:
Third-pass review suggestions applied in 15f74795c:

1. **Sentinel error now enriched at call site.** `GetOrCreateAttributes()` in `TUnit.Core/TestMetadata.cs` checks `ReferenceEquals(AttributeFactory, TestEntrySentinel.IndexedAttributeFactoryPlaceholder)` on the no-indexed-factory branch and throws `InvalidOperationException` carrying `TestName` and `TestClassType.FullName` — the sentinel delegate itself has no access to that context. The `Interlocked.CompareExchange` publication pattern is preserved (the throw happens before the CAS). The sentinel's own message is relaxed to a generic fallback since the call-site guard should normally prevent it from executing.
2. **XML docs on `MethodIndex` / `AttributeGroupIndex`** now note that each is only read when its paired indexed delegate (`IndexedInvokeBody` / `IndexedAttributeFactory`) is non-null.

Build: OK. Snapshot tests (`TUnit.Core.SourceGenerator.Tests`, net10.0): 116 passed / 1 skipped / 0 failed. Smoke (`BasicTests`): 3/3 passed.

thomhurst:
Diagnosed the Windows CI failure on run 24889949190.

**Failing test:** `Completes_Within_Unhappy2` in `TUnit.Assertions.Tests` (net472|x64) — `AssertionException: Expected to be less than 00:00:01` (observed 2s 551ms).

**Root cause:** Pre-existing timing flake. The test asserts total stopwatch duration < 1s after exercising `CompletesWithin(250ms)` against a 10s Task.Delay. On contended Windows CI, especially net472, blowing past the 1s bound is a known issue (see prior commit `128e1e3d0 Reduce flakiness on ExecutionTimeTests.cs`).

**Why unrelated to this PR:** The diff touches only `TestEntry.cs`, `TestMetadata.cs`, `TestMetadata\`1.cs`, `ExecutableTest\`1.cs`, and PublicAPI snapshots — no assertion, timing, or stopwatch code paths.

Re-running failed jobs.

PR #5728, opened 2026-04-24
perf: skip TimeoutHelper wrap when no explicit [Timeout] is set (#5711)

### LINKED ISSUE #5711: perf: skip TimeoutHelper wrap when no explicit [Timeout] is set
## Summary
Every test is wrapped in `TimeoutHelper.ExecuteWithTimeoutAsync` because `TestDetails.Timeout` is always set to the default (30 min), even when the user never applied `[Timeout]`. Estimated **~5% CPU + 4 extra allocations per test** (linked `CancellationTokenSource`, `TaskCompletionSource`, `CancellationTokenRegistration`, `Task.WhenAny` array).

This is likely the single biggest ROI fix from the profiling session.

## Evidence
Trace shows `TimeoutHelper.ExecuteWithTimeoutAsync` as ~5% inclusive / ~3% direct in the per-test path.

## Root cause
`TUnitSettings.Default.Timeouts.DefaultTestTimeout` (30 min) is copied into `TestDetails.Timeout` during build, so `testTimeout.HasValue` is always true:

- `TUnit.Engine/Building/TestBuilderPipeline.cs:260, 387, 466, 519`
- `TUnit.Engine/Building/TestBuilder.cs:1089, 1182`
- Check-site: `TUnit.Engine/TestExecutor.cs:203`

Every test then goes through `TUnit.Engine/Helpers/TimeoutHelper.cs:26` with the full allocation set.

## Proposed fix
- Leave `TestDetails.Timeout = null` when the user did not specify `[Timeout]`.
- Only run through `TimeoutHelper.ExecuteWithTimeoutAsync` when a timeout is actually requested.
- If a global default is still desirable, apply it as a coarser session-level guard rather than per-test.

## Expected impact
- ~5% CPU reduction per test.
- Removes 4 allocations per test — cuts GC pressure dramatically on large suites.

## Files
- `TUnit.Engine/Building/TestBuilderPipeline.cs`
- `TUnit.Engine/Building/TestBuilder.cs`
- `TUnit.Engine/TestExecutor.cs`
- `TUnit.Engine/Helpers/TimeoutHelper.cs`

thomhurst:
Closes #5711

## Summary

Every test was running through `TimeoutHelper.ExecuteWithTimeoutAsync` using the built-in 30-minute default, paying a linked `CancellationTokenSource` + `TaskCompletionSource` + `Task.WhenAny` cost per test for a ceiling almost no one relies on.

- `TestBuilder` and `TestBuilderPipeline` no longer seed `TestDetails.Timeout` from the default.
- `TestCoordinator` only feeds `TestExecutor` a non-null timeout when either a `[Timeout]` attribute set `TestDetails.Timeout`, or the user opted into a session-wide default via `TUnitSettings.Timeouts.DefaultTestTimeout`.
- `TestExecutor`'s existing null-timeout fast path then invokes the test directly with no wrapping overhead.

User-visible behavior is preserved:
- `[Timeout(...)]` still enforces its duration (verified against `TimeoutCancellationTokenTests`).
- Programmatic `DefaultTestTimeout` overrides are tracked via a new internal `DefaultTestTimeoutExplicitlySet` flag so they continue to apply to every test.

## Test plan

- [x] `dotnet build TUnit.Engine/TUnit.Engine.csproj -c Release -f net10.0` — clean
- [x] `TUnit.Core.SourceGenerator.Tests` — 116 passed, 0 failed (1 pre-existing skip)
- [x] `TUnit.TestProject` `BasicTests` filter — 3 passed
- [x] `TimeoutDoesNotFireTests` (positive path) — passes
- [x] `TimeoutCancellationTokenTests/BasicTest` with explicit `[Timeout(5000)]` — still correctly times out via `TimeoutHelper`


thomhurst:
Pushed `8ef26c9b0` addressing review feedback:

- **Doc gap (Concern 2):** `DefaultTestTimeout` XML doc now explicitly states that leaving the property unset means tests without `[Timeout]` run with no timeout wrapper at all, and the 30-minute fallback only kicks in when the property is explicitly assigned.
- **Minor (issue refs in comments):** Dropped `(#5711)` suffixes from both `TimeoutSettings.cs` and `TestCoordinator.cs`; kept the WHY text.
- **Micro-tidy in `TestCoordinator`:** Moved the `TUnitSettings.Default.Timeouts` lookup inside the `testTimeout is null` branch so the static singleton access is skipped on the common `[Timeout]`-set path.

**Concern 1 (DI of `TimeoutSettings`) deferred as a follow-up.** `TUnitSettings.Default` is the established access pattern used consistently by `HookTimeoutHelper`, `TestScheduler`, `TestRunner`, and `TUnitMessageBus` on the engine side — threading `TimeoutSettings` through constructor DI just for `TestCoordinator` would leave the architecture half-converted. Converting the whole engine to inject `TUnitSettings` is worth doing, but it's a larger DI-cleanup task separate from the perf fix. I'll open a follow-up for it if you agree.

Verified locally:
- `dotnet build TUnit.slnx -c Release` — 0 errors
- `TUnit.Core.SourceGenerator.Tests` (net10.0) — 116 passed, 1 pre-existing skip
- `TUnit.TestProject --treenode-filter "/*/*/BasicTests/*"` (net10.0) — 3/3 passed

Codacy reported 1 minor CodeStyle issue; annotations aren't exposed through GitHub or public API so I couldn't pinpoint it exactly. The revised block replaces the previous mixed `var` / explicit-type variable pair with a single `var` and drops the `timeouts` local from the fast path, which should address the most likely candidates.

thomhurst:
Pushed `7195d82f2` addressing the remaining feedback:

**Non-blocking review suggestion (nullable backing field).** Replaced the `TimeSpan _defaultTestTimeout` + `bool _defaultTestTimeoutExplicitlySet` pair with a single `TimeSpan? _defaultTestTimeout`:

- Public `DefaultTestTimeout` getter now returns `_defaultTestTimeout ?? TimeSpan.FromMinutes(30)` — same observable default for any user code reading the property, no breaking change.
- Setter just assigns; no more secondary flag to keep in sync.
- New internal `ExplicitDefaultTestTimeout` surfaces the raw `TimeSpan?` for the engine path.
- `TestCoordinator` collapses to a single null-coalescing chain:
  ```csharp
  var testTimeout = test.Context.Metadata.TestDetails.Timeout
      ?? TUnitSettings.Default.Timeouts.ExplicitDefaultTestTimeout;
  ```
  Still skips the static-singleton lookup on the common `[Timeout]`-set path, and the wrap-skip semantics are unchanged (null flows through to `TestExecutor`'s existing null-timeout fast path).

**Codacy ACTION_REQUIRED (1 minor CodeStyle).** Same commit. The specific rule isn't exposed through the GitHub annotations pipeline or the public Codacy API for this account, but the rewrite removes the most plausible candidates: the mixed-intent dual-field declaration, the nested `if` + intermediate `timeouts` local, and the redundant assignment in the setter. If a separate issue remains, it'll retrigger on this push and I'll handle it specifically.

Verified locally on net10.0:
- `dotnet build TUnit.slnx -c Release` — 0 errors
- `TUnit.Core.SourceGenerator.Tests` — 116 passed, 1 pre-existing skip
- `TUnit.TestProject --treenode-filter "/*/*/BasicTests/*"` — 3/3 passed
- `TUnit.TestProject --treenode-filter "/*/*/TimeoutCancellationTokenTests/*"` — exercises both branches: `DefaultTest` (no `[Timeout]`, no default set → no wrap, completes in 5s) passes; the 7 `ExpectedResult.Failure` tests with `[Timeout(5_000)]` fire and fail as designed.

thomhurst:
Added test coverage for the `ExplicitDefaultTestTimeout` fallback in `TUnit.UnitTests/TUnitSettingsTests.cs` (commit b43f1498d):

- `ExplicitDefaultTestTimeout_Is_Null_When_Unset` — a freshly-constructed `TimeoutSettings` returns `null` from the internal getter while the public `DefaultTestTimeout` still reports the 30-minute fallback. Confirms the right-hand side of the coalesce is null when the user never assigns `DefaultTestTimeout` (the no-wrap fast path).
- `ExplicitDefaultTestTimeout_Returns_Assigned_Value` — once the setter is called, the internal getter surfaces the assignment.
- `Configured_Default_Timeout_Fires_On_Hanging_Test` — mirrors `TestCoordinator`'s call site by feeding `ExplicitDefaultTestTimeout` into the same `TimeoutHelper.ExecuteWithTimeoutAsync`. A non-cooperative 30s hang against a 200ms configured default throws `TimeoutException` as expected. Uses a private CTS in `finally` to terminate the hang task rather than leaking it to process exit.

All three tests reuse the existing `SnapshotSettings` / `RestoreSettings` hooks in `TUnitSettingsTests`, so no new global-state leakage. Full `TUnit.UnitTests` (183 tests) green on `net10.0`.

thomhurst:
Round 5 review follow-up — 7cc64e7ba.

**Primary finding (inlined no-retry fast path in TestCoordinator) — not reproducible.**
Checked `TUnit.Engine/Services/TestExecution/TestCoordinator.cs` on this branch: there is only one `_testExecutor.ExecuteAsync` call site, at line 325 inside `ExecuteTestLifecycleAsync`, and it already does
```csharp
var testTimeout = test.Context.Metadata.TestDetails.Timeout
    ?? TUnitSettings.Default.Timeouts.ExplicitDefaultTestTimeout;
```
Both branches of the `retryLimit == 0` check (lines 109–120) route through `ExecuteTestLifecycleAsync`; there is no inlined copy at line ~154. `grep -n 'TestDetails\.Timeout\|ExecuteAsync(test, _testInitializer' TUnit.Engine/Services/TestExecution/TestCoordinator.cs` confirms the single occurrence. So the observed behaviour (a user-set `DefaultTestTimeout` only applying to `[Retry]` tests) can't happen on this branch — the coalesce is on the shared path.

**Secondary finding (Snapshot/Restore leaks state) — real, fixed in 7cc64e7ba.**
`SnapshotSettings()` was reading through the public `DefaultTestTimeout` getter, which coalesces the null backing field to 30 minutes. `RestoreSettings()` then wrote that value back through the validating setter, permanently pinning `_defaultTestTimeout` after the first test ran. Fix:
- `TimeoutSettings.SetExplicitDefaultTestTimeout(TimeSpan?)` — internal, accepts null, reuses `ValidatePositive` for non-null values.
- Snapshot now captures the internal `ExplicitDefaultTestTimeout` (nullable); restore uses the new setter.
- New `SnapshotRestore_Preserves_Unset_DefaultTestTimeout` test exercises the full round-trip on `TUnitSettings.Default` to guard the regression.

**Verified**
- `dotnet build TUnit.slnx -c Release` — 0 errors.
- `TUnit.UnitTests` net10.0 — 184/184 passed (including the existing tests from b43f1498d and the new round-trip guard).
- `TUnit.Core.SourceGenerator.Tests` net10.0 — 116/116 passed (1 pre-existing skip).
- `TUnit.TestProject` `BasicTests` filter — 3/3 passed.

If the "line ~154 inlined fast path" finding was based on a local branch or a different PR, please point me at the exact commit/file and I'll take another look.

thomhurst:
### Fourth-round review follow-up

**Bug 2 (DefaultTestTimeout misclassification) — fixed via Option A.** Commit 6ff6f367a.

Went with Option A (propagate the resolved timeout back to `TestDetails.Timeout` at the coordinator) over Option B (simplify the classification guard). Rationale:

- Option A keeps the duration/`TestDetails.Timeout` guard intact, preserving its defense against misclassifying unrelated `OperationCanceledException` paths (the FailureCategorizer groups those with real timeouts).
- Option B has a wider blast radius: flipping the classification to trust the category alone means several existing tests that rely on `Counters.Failed` for timeout outcomes (TimeoutTests1, HookTimeoutTests, etc.) would need updating to `Counters.Timeout`. Not worth the collateral for this PR.
- `TestDetails.Timeout` has a public setter on this branch (not init-only), so Option A fits cleanly.

**Integration test added** under `TUnit.TestProject/Bugs/5728/DefaultTimeoutClassificationTests.cs` + matching engine test at `TUnit.Engine.Tests/DefaultTimeoutClassificationTests.cs`. The TestProject class uses a `[Before(TestDiscovery)]` hook that only applies `DefaultTestTimeout = 200ms` when `TUNIT_BUG5728_DEFAULT_TIMEOUT=1` — so the process-wide setting never leaks into other runs of TUnit.TestProject. The engine test spawns the TestProject with that env var via a new `RunOptions.WithEnvironmentVariable` helper and asserts the failure message contains "timed out".

Before the fix, the test body would complete the full 10s `Task.Delay`. After the fix, it fails in ~1.2s with `TimeoutException: Test '...' timed out after 00:00:00.2000000`.

**Bug 1 (no-retry fast path on TestCoordinator) — still not present on this branch.** Confirmed once more: `TUnit.Engine/Services/TestExecution/TestCoordinator.cs:323-325` is the single call site and already includes the `?? ExplicitDefaultTestTimeout` coalesce. The inlined fast path is introduced by #5725. Whichever of #5725 / #5728 lands second will need to propagate the coalesce (and this timeout write-back) to the inlined call site.

### Verification

- `dotnet build TUnit.slnx -c Release` — clean.
- `TUnit.Core.SourceGenerator.Tests` (net10.0) — 116 pass / 1 skipped.
- `TUnit.UnitTests` (net10.0) — 184 pass.
- `TUnit.Engine.Tests` new test `DefaultTimeoutClassificationTests/*` — pass.
- `TUnit.Engine.Tests` existing `TimeoutTests1/*` and `HookTimeoutTests/*` — still pass (no regression from the `TestDetails.Timeout` write-back).

Head SHA: 6ff6f367a.

thomhurst:
Re: Round 6 "no-retry fast path still ignores `ExplicitDefaultTestTimeout`" — I don't think this bug is real. The review describes an inlined copy of the lifecycle at `TestCoordinator.cs:154`, but that line is in the cleanup block (`cleanupExceptions = await _objectTracker.UntrackObjects(...)`), not a duplicated execute path.

The no-retry branch still calls `ExecuteTestLifecycleAsync` — it just skips the `RetryHelper` wrapper:

```csharp
// TestCoordinator.cs ~L107-120
var retryLimit = test.Context.Metadata.TestDetails.RetryLimit;

if (retryLimit == 0)
{
    test.Context.CurrentRetryAttempt = 0;
    await ExecuteTestLifecycleAsync(test, cancellationToken).ConfigureAwait(false);
}
else
{
    await RetryHelper.ExecuteWithRetry(test.Context,
        () => ExecuteTestLifecycleAsync(test, cancellationToken)).ConfigureAwait(false);
}
```

Both paths land in `ExecuteTestLifecycleAsync`, where the coalesce + write-back live (~L322-333):

```csharp
var testTimeout = test.Context.Metadata.TestDetails.Timeout
    ?? TUnitSettings.Default.Timeouts.ExplicitDefaultTestTimeout;
if (testTimeout is not null && test.Context.Metadata.TestDetails.Timeout is null)
{
    test.Context.Metadata.TestDetails.Timeout = testTimeout;
}
await _testExecutor.ExecuteAsync(test, _testInitializer, cancellationToken, testTimeout).ConfigureAwait(false);
```

So `DefaultTestTimeout` is honored for both retry and no-retry tests. The `Hanging_Test_With_DefaultTestTimeout_Should_Timeout` integration test passing is the expected outcome, not evidence of another mechanism.

Happy to re-look if I'm missing something, but from the current `perf/5711-timeout-skip` tip I can't see the duplicated execute path the review references.

thomhurst:
Round 11 follow-up — db7daf1e9.

**🟡 AOT concern: applied.** `GlobalContext.Current` now passes an explicit factory:

```csharp
get => LazyInitializer.EnsureInitialized(ref _current, static () => new GlobalContext())!;
```

Avoids the `Activator.CreateInstance<T>()` codepath whose `[DynamicallyAccessedMembers]` annotation may not preserve `GlobalContext`'s internal ctor under trimming.

**🔴 "Blocker" — verified not present in current code.** The review references the inlined no-retry fast path at `TestCoordinator.cs:152–155` with the un-coalesced `var testTimeout = test.Context.Metadata.TestDetails.Timeout;`. That inlined block was removed in commit `7cc64e7ba` (Round 5 follow-up — the same fix flagged in the Round 6 "no-retry fast path" thread). The current `ExecuteTestInternalAsync` at lines 109-122 routes **both** retry-limit=0 and retry-limit>0 through `ExecuteTestLifecycleAsync`:

```csharp
if (retryLimit == 0)
{
    test.Context.CurrentRetryAttempt = 0;
    await ExecuteTestLifecycleAsync(test, cancellationToken).ConfigureAwait(false);
}
else
{
    await RetryHelper.ExecuteWithRetry(test.Context,
        () => ExecuteTestLifecycleAsync(test, cancellationToken)).ConfigureAwait(false);
}
```

`ExecuteTestLifecycleAsync` (lines 320-330 on this branch) has the proper coalesce + write-back:

```csharp
var testTimeout = test.Context.Metadata.TestDetails.Timeout
    ?? TUnitSettings.Default.Timeouts.ExplicitDefaultTestTimeout;
if (testTimeout is not null && test.Context.Metadata.TestDetails.Timeout is null)
{
    test.Context.Metadata.TestDetails.Timeout = testTimeout;
}
await _testExecutor.ExecuteAsync(test, _testInitializer, cancellationToken, testTimeout).ConfigureAwait(false);
```

`Hanging_Test_With_DefaultTestTimeout_Should_Timeout` has no `[Retry]`, takes the `retryLimit == 0` branch, and goes straight through `ExecuteTestLifecycleAsync` with the coalesce.

The CI failure on macOS for that test is a separate issue (the `[Before(TestDiscovery)]` filter-gate in `DefaultTimeoutClassificationHooks` not taking effect under MTP test-host-controller mode); it isn't a missing coalesce in TestCoordinator.

PR #5730, opened 2026-04-24
perf(engine): batch per-test overhead cleanups (#5719)

### LINKED ISSUE #5719: perf: small per-test overhead cleanups (batch)
## Summary
Batch of five smaller per-test overhead fixes surfaced by CPU-sampling profiling. Individually modest (~0.2-1% each), collectively scale with suite size. All live-in the hot path.

Trace source: `dotnet-trace` CPU sampling of `TUnit.Profile` (1013 tests / 377ms, net10.0 Release).

---

### 1. `EnsureEventReceiversCached` called 5× per test
`TUnit.Engine/Extensions/TestContextExtensions.cs:21-141` — 5 getters (`GetTestStartReceivers`, `GetTestEndReceivers`, etc.) each guard via this method. Trace: 0.39% inclusive / 0.35% self.

**Fix:** hoist a single call into `TestCoordinator.ExecuteTestLifecycleAsync`; replace internal check with a plain `_eventReceiversBuilt` bool. Inline the getters.

Expected: ~0.3% CPU.

---

### 2. `ObjectTracker.s_trackedObjects` global dict churn
`TUnit.Core/Tracking/ObjectTracker.cs:44, 216` — every per-test object (class instance, args, attributes, injected props) hits `GetOrCreateCounter` then `TryRemove` on the shared static dict. Trace: `ConcurrentDictionary.TryRemoveInternal` 0.17% self.

**Fix:** store the `Counter` on a per-`TestContext` slot for non-shared objects; fall back to the global dict only for genuinely shared objects (already identified via `_initializedObjects`).

Expected: ~0.3% CPU, drops contention on the global dict.

---

### 3. `ObjectGraphDiscoverer.TraverseInitializerProperties` caching
`TUnit.Core/Discovery/ObjectGraphDiscoverer.cs:397-441` — for every test, for every graph object, walks `type.BaseType` up to `object` and allocates a `HashSet<string>` (line 413) just to dedup property names. `ShouldSkipType` also runs `type.Namespace?.StartsWith("System")` per call.

**Fix:** cache per-type flattened `InitializerPropertyInfo[]` in a `ConcurrentDictionary<Type, ...>`. Cache `ShouldSkipType` result per Type. Ideally emit a compile-time flag "has zero initializer properties transitively" at source-gen time.

Expected: ~0.2% CPU.

---

### 4. `RestoreExecutionContext` called ~10× per test
`TUnit.Engine/TestExecutor.cs:87, 105, 117, 168, 173, 180` + `TestCoordinator:319` + multiple sites in hook execution. Each call touches `ExecutionContext.Restore` + `SetAsyncLocalContext`.

**Fix:** add a version counter to `Context`; thread-static "last-restored-version"; skip Restore when unchanged on this logical thread.

Expected: ~0.5-1% CPU, reduces AsyncLocal churn.

---

### 5. Dead `_maxParallelismSemaphore` on modern .NET
`TUnit.Engine/Scheduling/TestScheduler.cs:60-61, 341-376` — `Lazy<SemaphoreSlim>` allocated but only used on the netstandard2.0 fallback. On .NET, `Parallel.ForEachAsync`'s `MaxDegreeOfParallelism` is authoritative.

**Fix:** drop the Lazy + semaphore parameter for the .NET path. Cosmetic.

Expected: few bytes per session, removes dead code.

---

## Combined impact
Roughly **~1.5-2% CPU** across all five. Low-risk, high-clarity cleanups.

thomhurst:
## Summary

Batch of the smaller per-test overhead fixes surfaced in #5719. Individually modest (~0.2-1% each), collectively ~1% CPU across typical suites.

Closes #5719

### Shipped (4 of 5)

- **Hoist `EnsureEventReceiversCached`** out of the five per-test receiver getters. Replaced the per-getter cache-array probe with an `EventReceiversBuilt` bool on `TestContext`; `TestCoordinator.ExecuteTestLifecycleAsync` rebuilds once after `ClassInstance` is assigned. Hot-path getters (`GetTestStartReceivers`/`GetTestEndReceivers`/`GetTestSkippedReceivers`) are now plain field reads.
- **Cache `ObjectGraphDiscoverer.TraverseInitializerProperties` output per-type.** The inheritance walk + `HashSet<string>` dedupe now runs once and is memoized in a `ConcurrentDictionary<Type, InitializerPropertyInfo[]?>`. `ShouldSkipType` is memoized too, so `Namespace.StartsWith("System")` no longer runs per call.
- **Drop `Lazy<SemaphoreSlim> _maxParallelismSemaphore`** on the .NET 8+ path. `Parallel.ForEachAsync` caps concurrency via `ParallelOptions.MaxDegreeOfParallelism`; the semaphore is now only built on the netstandard2.0 fallback.

### Skipped (2 of 5)

- **Per-TestContext slot for `ObjectTracker` counters.** Requires a robust "shared vs per-test" signal at Track/Untrack time. The existing `_initializedObjects` set in `EventReceiverOrchestrator` only flags seen-before *after* the second test, so the bulk of per-test objects would still hit the shared dict on first track. Needs a dedicated sharing marker on the data-source side.
- **`RestoreExecutionContext` version-counter skip.** Implementation attempted a thread-static "last restored context + version" guard, but regressed `TUnit.TestProject/Bugs/1914/AsyncHookTests` (AsyncLocal-flow chain from `BeforeTestDiscovery` through `BeforeTestSession` / `BeforeAssembly` / `BeforeClass` / `BeforeTest`). The captured `ExecutionContext` on a context can carry AsyncLocal values set by hooks on the active async flow between Restore sites; the skip misses those flows when a continuation lands on a thread whose last-Restored context points elsewhere. Held back for a safer design.

## Test plan

- [x] `TUnit.Core.SourceGenerator.Tests` — 116 passed, 1 skipped (known)
- [x] `TUnit.TestProject` `BasicTests` filter — 3/3 passed
- [x] `TUnit.TestProject` `AsyncHookTests` filter (the 1914 regression guard) — 8/8 passed
- [x] `TUnit.TestProject` `DataDrivenTests` filter — 15/15 passed

### DIFF
--- a/TUnit.Core/Discovery/ObjectGraphDiscoverer.cs
+++ b/TUnit.Core/Discovery/ObjectGraphDiscoverer.cs
@@ -442,39 +442,33 @@ internal sealed class ObjectGraphDiscoverer : IObjectGraphTracker
     /// source-gen registration exists in the type's inheritance chain.
     /// </summary>
     private static InitializerPropertyInfo[]? GetFlattenedInitializerProperties(Type type)
-    {
-        if (FlattenedInitializerPropertiesCache.TryGetValue(type, out var cached))
+        => FlattenedInitializerPropertiesCache.GetOrAdd(type, static t =>
         {
-            return cached;
-        }
+            List<InitializerPropertyInfo>? merged = null;
+            HashSet<string>? seen = null;
 
-        List<InitializerPropertyInfo>? merged = null;
-        HashSet<string>? seen = null;
-
-        for (var currentType = type; currentType != null && currentType != typeof(object); currentType = currentType.BaseType)
-        {
-            var registered = InitializerPropertyRegistry.GetProperties(currentType);
-            if (registered == null)
+            for (var currentType = t; currentType != null && currentType != typeof(object); currentType = currentType.BaseType)
             {
-                continue;
-            }
-
-            merged ??= new List<InitializerPropertyInfo>(registered.Length);
-            seen ??= new HashSet<string>(StringComparer.Ordinal);
-
-            foreach (var p in registered)
-            {
-                if (seen.Add(p.PropertyName))
+                var registered = InitializerPropertyRegistry.GetProperties(currentType);
+                if (registered == null)
                 {
-                    merged.Add(p);
+                    continue;
+                }
+
+                merged ??= new List<InitializerPropertyInfo>(registered.Length);
+                seen ??= new HashSet<string>(StringComparer.Ordinal);
+
+                foreach (var p in registered)
+                {
+                    if (seen.Add(p.PropertyName))
+                    {
+                        merged.Add(p);
+                    }
                 }
             }
-        }
 
-        var result = merged?.ToArray();
-        FlattenedInitializerPropertiesCache[type] = result;
-        return result;
-    }
+            return merged?.ToArray();
+        });
 
     /// <summary>
     /// Traverses the pre-flattened source-generated IAsyncInitializer properties.
--- a/TUnit.Core/Discovery/ObjectGraphDiscoverer.cs
+++ b/TUnit.Core/Discovery/ObjectGraphDiscoverer.cs
@@ -62,6 +62,17 @@ internal sealed class ObjectGraphDiscoverer : IObjectGraphTracker
     // Thread-safe collection of discovery errors for diagnostics
     private static readonly ConcurrentBag<DiscoveryError> DiscoveryErrors = [];
 
+    // Memoize ShouldSkipType — Namespace.StartsWith("System") is expensive when repeated
+    // across every per-test initializer traversal.
+    private static readonly ConcurrentDictionary<Type, bool> ShouldSkipTypeCache = new();
+
+    // Cached flattened InitializerPropertyInfo[] for each type, including base-class
+    // properties with derived-first precedence. Traversal would otherwise walk the
+    // inheritance chain and allocate a HashSet<string> on every call.
+    // - non-null array: type has registered source-gen metadata (at this or an ancestor level).
+    // - null           : no source-gen registration anywhere in hierarchy — use reflection fallback.
+    private static readonly ConcurrentDictionary<Type, InitializerPropertyInfo[]?> FlattenedInitializerPropertiesCache = new();
+
     /// <summary>
     /// Gets all discovery errors that occurred during object graph traversal.
     /// Useful for debugging and diagnostics when property access fails.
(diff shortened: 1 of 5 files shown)


thomhurst:
Addressed review feedback in 05f1046c7:

1. Added `Debug.Assert(testContext.EventReceiversBuilt, ...)` to the three hot-path getters (`GetTestStartReceivers`, `GetTestEndReceivers`, `GetTestSkippedReceivers`). Catches invariant violations in Debug/tests at zero Release cost.
2. Unified dictionary access in `ObjectGraphDiscoverer`: `GetFlattenedInitializerProperties` now uses `GetOrAdd` with a static factory, matching the existing `ShouldSkipType` pattern.
3. Added a `<remarks>` block on `ExecuteTestLifecycleAsync` (plus a pointer comment by the skip branch) documenting the `RegisterReceivers -> EnsureEventReceiversCached` ordering invariant that keeps the early skip path safe.

Verified locally: `dotnet build TUnit.slnx -c Release` clean, source-gen snapshot tests pass (116/116), and BasicTests smoke run passes (3/3).

thomhurst:
Over-engineered Item 1 simplified — kept `EventReceiversBuilt` for TFM unification, dropped the hoist + `Debug.Assert`s, added `[MethodImpl(AggressiveInlining)]` to the `EnsureEventReceiversCached` fast path. Same perf, none of the invariant-maintenance cost. Items #2 and #3 unchanged.

thomhurst:
Addressed the size-budget concern on `EnsureEventReceiversCached`.

Refactored into the canonical fast/slow split:

- `EnsureEventReceiversCached` is now a ~10-byte IL wrapper: load `ClassInstance`, check `EventReceiversBuilt && ReferenceEquals(CachedClassInstance, currentClassInstance)`, return on hit, otherwise tail-call the builder. Well under RyuJIT's inliner size budget, so `[MethodImpl(AggressiveInlining)]` now reliably fires and every call site collapses to a field-load + branch + return in the 99%+ hot case.
- `BuildEventReceiverCaches` holds the ~100 bytes of build logic (invalidation, eligible-objects walk, 5 list builds, 5 `SortAndFilter` calls, `CachedClassInstance` + `EventReceiversBuilt` commit). Marked `[MethodImpl(NoInlining)]` so the JIT keeps a single outlined copy instead of duplicating it into every caller. `currentClassInstance` is threaded through from the gate to avoid a redundant field re-read.

No behavioral change — all invalidation semantics and cache write order are preserved.

Also piggybacked the `#if NET8_0_OR_GREATER` -> `#if NET` unification across the two files already in this PR (`TestScheduler.cs`, `TestContextExtensions.cs`). `#if NET` is the majority convention across `TUnit.Engine` (87 occurrences vs 42), and both forms are equivalent for the `netstandard2.0;net8.0;net9.0;net10.0` target set.

Verification: `dotnet build TUnit.slnx -c Release` (0 errors), source-generator snapshot tests pass (116/116), `BasicTests` smoke pass (3/3).

PR #5729, opened 2026-04-24
perf(engine): remove duplicate object-graph traversal in ObjectLifecycleService (#5718)

thomhurst:
Closes #5718

## Summary
- Every test walks its property / `IAsyncInitializer` object graph **twice**: once during registration (`TrackableObjectGraphProvider.GetTrackableObjects` → `ObjectGraphDiscoverer.DiscoverAndTrackObjects`) to populate `TestContext.TrackedObjects`, and again during execution inside `InitializeObjectWithSpanAsync`, which re-walks every tracked root's nested graph via `InitializeNestedObjectsForExecutionAsync`.
- `InitializeTrackedObjectsAsync` iterates `TrackedObjects` in descending-depth order and every reachable nested object is already flattened into that sorted list at registration time, so deeper dependencies are already initialized by the time a shallower object is processed. The per-object nested walk is redundant.
- Removed the `InitializeNestedObjectsForExecutionAsync(obj)` call inside `InitializeObjectWithSpanAsync`. Each `IAsyncInitializer` is still called exactly once — `ObjectInitializer.InitializeAsync` is deduplicated via `Lazy<Task>`, and initialization order (deepest-first) is preserved by the existing depth-descending iteration.
- `InitializeObjectForExecutionAsync` (the class-data init called before the test class constructor) keeps its nested walk, since `TrackedObjects` has not yet been iterated at that point.

## Impact
Estimated ~0.3-0.5% CPU across the suite; halves the inclusive time of `InitializeObjectWithSpanAsync` and `TraverseInitializerProperties`. No public API change.

## Correctness analysis
`DiscoverAndTrackObjects` (tracking) and `DiscoverNestedObjectGraph` (execution) share the same traversal functions (`TraverseInjectableProperties` + `TraverseInitializerProperties`). They differ only in one early-out (`useSourceRegistrarCheck`) that is a pure micro-optimization when there are no injectable properties — both code paths still descend into `TraverseInitializerProperties`. Tracking is therefore a (super)set of what execution-phase discovery would find, so every object reachable at execution time is already present in `TrackedObjects`.

## Test plan
- [x] `dotnet build TUnit.slnx -c Release` — succeeds, 0 errors
- [x] Snapshot tests: `TUnit.Core.SourceGenerator.Tests` — 116 passed / 1 skipped (pre-existing)
- [x] `TUnit.UnitTests` — 180/180 pass
- [x] `TUnit.TestProject` property-injection / nested-graph smoke tests on net10.0:
[...]

PR #5731, opened 2026-04-24
perf(engine): replace global lock in EventReceiverRegistry with lock-free CAS

### LINKED ISSUE #5715: perf: replace global lock in EventReceiverRegistry with lock-free structure
## Summary
`EventReceiverRegistry.RegisterReceivers` takes a global `Lock` once per test. With `Environment.ProcessorCount * 4` concurrent tests this is a top source of contention in the trace: **`Monitor.Enter_Slowpath` at 2.16% self**. Estimated **~1.5% CPU + removes hot lock**.

## Evidence
Trace top exclusive:
```
7.  Monitor.Enter_Slowpath(...)                   2.16%  (self)
```

Code: `TUnit.Engine/Events/EventReceiverRegistry.cs:29-96, 145-168`. `RegisterReceivers` is called from `TUnit.Engine/Services/TestExecution/TestCoordinator.cs:67` on every test start.

The registry is conceptually a copy-on-write `Dictionary<Type, object[]>` — currently guarded by a single shared lock.

## Proposed fix
- Switch to `ConcurrentDictionary<Type, ImmutableArray<object>>`.
- Update via `AddOrUpdate` with `ImmutableArray.Add` or a snapshot-and-swap via `Interlocked.CompareExchange`.
- Short-circuit `RegisterReceivers` entirely when every object in the incoming span is already in `_initializedObjects` (common case — most eligible objects are shared across tests).

## Expected impact
- ~1.5% CPU reduction on parallel paths.
- Eliminates lock contention that scales with parallelism (gets worse on bigger machines / bigger suites).

thomhurst:
## Summary

- Replaces the global `Lock` in `EventReceiverRegistry` with lock-free structures (`ConcurrentDictionary<Type, object[]>` + CAS).
- Uses `AddOrUpdate` copy-on-write array append for writers; `Interlocked.CompareExchange` loop for the `_registeredEvents` bitflag OR.
- Removes lock-protected `GetReceiversOfType<T>` cache-miss path; uses snapshot + source-identity validation after `GetOrAdd` to detect writer races.

## Why

`EventReceiverRegistry.RegisterReceivers` was called per test from `TestCoordinator` with `ProcessorCount * 4` parallel writers. Trace showed `Monitor.Enter_Slowpath` at **2.16% self** — a top contention source. Estimated **~1.5% CPU** reduction with the hot lock removed.

## Semantics preserved

Receivers remain keyed by interface type and are discoverable through the existing `GetReceiversOfType<T>()` API. Cache invalidation on writer (`TryRemove` per-type) already existed; the new retry in `GetReceiversOfType` closes the race window previously masked by the shared lock.

Closes #5715

## Test plan

- [x] Build: `dotnet build TUnit.Engine/TUnit.Engine.csproj -c Release` (clean; multi-TFM: netstandard2.0, net8, net9, net10)
- [x] TUnit.TestProject smoke: `--treenode-filter "/*/*/BasicTests/*"` passes (3/3)
- [x] TUnit.TestProject smoke: `--treenode-filter "/*/*/ScopedEventReceiverTests/*"` passes (2/2)
- [x] TUnit.TestProject smoke: `--treenode-filter "/*/*/CallEventReceiverTests/*"` passes (2/2)
- [x] TUnit.Core.SourceGenerator.Tests: 116/116 passing (1 pre-existing skip)
- [x] TUnit.Engine.Tests: 146 passing, 99 skipped, 2 pre-existing FSharp/VB-test failures unrelated to this change
- [x] Pre-existing `EventReceiverTests`/`EventReceiverStageTests`/`LastTestEventReceiverTests` failures verified to fail identically on `main` (shared static tracker + parallel tests); not a regression from this change.

### DIFF
--- a/TUnit.Engine/Events/EventReceiverRegistry.cs
+++ b/TUnit.Engine/Events/EventReceiverRegistry.cs
@@ -150,6 +150,10 @@ internal sealed class EventReceiverRegistry
     {
         var typeKey = typeof(T);
 
+        // Termination contract: this retry loop relies on writers (RegisterReceivers) completing
+        // before steady-state reads. All registrations happen during test setup; during execution
+        // the receiver set is effectively frozen. A caller that registers receivers during test
+        // execution could cause this loop to spin under continuous writer pressure.
         while (true)
         {
             // Lock-free fast path: cache hit (common case after warmup).
--- a/TUnit.Engine/Events/EventReceiverRegistry.cs
+++ b/TUnit.Engine/Events/EventReceiverRegistry.cs
@@ -7,6 +7,13 @@ namespace TUnit.Engine.Events;
 /// Fast registry for event receiver presence checks using bit flags.
 /// Thread-safe without locks: uses ConcurrentDictionary + copy-on-write arrays
 /// updated via AddOrUpdate. Reads are lock-free.
+/// <remarks>
+/// Batch <see cref="RegisterReceivers(ReadOnlySpan{object})"/> is NOT atomic: a concurrent
+/// reader may observe a subset of a batch because each receiver is registered independently.
+/// Callers must not depend on all-or-nothing visibility. Current callers register receivers
+/// before any test executes (see <c>EventReceiverOrchestrator</c>, which also dedups), so
+/// this relaxation is safe. Future maintainers must preserve that invariant.
+/// </remarks>
 internal sealed class EventReceiverRegistry
 {
     // Bit flags for fast checking
@@ -27,7 +34,8 @@ internal sealed class EventReceiverRegistry
         All = ~0
     }
 
-    // _registeredEvents is volatile for publication semantics; updated via Interlocked OR.
+    // Accessed via Volatile.Read + Interlocked.CompareExchange to provide acquire/release
+    // semantics without the cost of a volatile field on every write path.
     private int _registeredEvents;
 
     // Copy-on-write: each value is an immutable snapshot array. Writers replace the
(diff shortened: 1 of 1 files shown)


thomhurst:
Addressed the three review comments in c599e084f (doc/comment only, no functional changes):

1. Rewrote the misleading `volatile` comment on `_registeredEvents` to accurately describe the `Volatile.Read` + `Interlocked.CompareExchange` pattern.
2. Added a WHY comment at the `ICollection<KVP>.Remove` cast explaining it's a netstandard2.0-compatible equivalent of `ConcurrentDictionary.TryRemove(KeyValuePair<K,V>)` that only evicts when the value still matches.
3. Added a class-level `<remarks>` noting that batch `RegisterReceivers` is not atomic and that current callers register before any test executes, so partial-batch visibility is safe.

Verified: `dotnet build TUnit.slnx -c Release` clean, snapshot tests 116/116 pass, `BasicTests/*` smoke filter 3/3 pass.

thomhurst:
Applied the non-blocking suggestion in 48cffd24d: added a concise WHY comment above the retry loop in `GetReceiversOfType<T>` documenting the termination contract (writers complete before steady-state reads; adversarial registration during test execution could cause spin under continuous writer pressure). Build green; snapshot tests pass (116/116, 1 skipped as expected).

thomhurst:
Diagnosed the Windows CI failure on run 24889951335.

**Failing test:** `AutoWires_TagsAspNetCoreSpans_WithTestId` in `RunAspNetCoreTestsModule` — `System.NullReferenceException: Object reference not set to an instance of an object.`

**Root cause:** Pre-existing flake — the same NRE on `WebApplicationFactory.DisposeAsync` previously flagged during PR #5729's investigation. Unrelated to this PR.

**Why unrelated to this PR:** The diff touches only `TUnit.Engine/Events/EventReceiverRegistry.cs` — no ASP.NET Core / `WebApplicationFactory` code paths.

Re-running failed jobs.

PR #5739, opened 2026-04-25
fix(migration): preserve comments/XML docs when removing sole attributes (#5698)

### LINKED ISSUE #5698: [Bug]: NUnit migration analyzer removes class comments along with the attribute
### Description

Looks like class comments or docstrings are removed together with `[TestFixture]` attribute.

In many cases where a test class was annotated with `/// <summary>` docstring or just a comment followed by the `[TestFixture]` NUnit attribute, code fixer removed those comments / docstrings along with the attribute while migrating.

### Expected Behavior

Docstrings and comments are kept.

### Actual Behavior

Everything attached to the attribute above it disappears during migration.

### Steps to Reproduce

```csharp
using NUnit.Framework;

// TODO Fix this
/// <summary>
/// This class is testing X
/// </summary>
[TestFixture]
public class MyTest
{
   [Test] 
   public void MyTestMethod()
   {
   }
}
```

Right-click _TestFixture_ and select _Convert NUnit code to TUnit_

### TUnit Version

1.37.36

### .NET Version

.NET 10

### Operating System

Windows

### IDE / Test Runner

```shell
public class MyTest
{
   [Test] 
   public void MyTestMethod()
   {
   }
}
```

### Additional Context

_No response_

### IDE-Specific Issue?

thomhurst:
## Summary

- Fixes #5698 — NUnit/XUnit/MSTest migration code fix dropped class & method comments and XML docstrings whenever the attribute list above them held a single attribute (e.g. `[TestFixture]`).
- `MigrationTransformer.RemoveAttributes` removed the entire `AttributeListSyntax` with `SyntaxRemoveOptions.KeepNoTrivia`, throwing away the leading trivia.
- Now extracts meaningful leading trivia (comments, doc comments, preprocessor directives) and moves it onto the next token before removing the attribute list. Surrounding whitespace is still discarded so no extra indentation is left behind, preserving the original intent of `KeepNoTrivia`.

Repro from the issue now produces the expected output:
```csharp
// TODO Fix this
/// <summary>
/// This class is testing X
/// </summary>
public class MyTest { ... }
```

## Test plan

- [x] New TUnit test reproducing the exact case in the issue (`NUnit_TestFixture_Attribute_Removed_Preserves_Leading_Comments_And_Docstring`)
- [x] Additional coverage for a single-line comment variant
- [x] All NUnit migration tests pass (167)
- [x] All XUnit migration tests pass (82, 1 skipped pre-existing)
- [x] All MSTest migration tests pass (52)

### DIFF
--- a/TUnit.Analyzers.CodeFixers/Base/TwoPhase/MigrationTransformer.cs
+++ b/TUnit.Analyzers.CodeFixers/Base/TwoPhase/MigrationTransformer.cs
@@ -764,9 +764,7 @@ public class MigrationTransformer
 
                 if (attributeList.Attributes.Count == 1)
                 {
-                    // Remove the entire attribute list without keeping its trivia
-                    // This prevents extra indentation from being left behind
-                    currentRoot = currentRoot.RemoveNode(attributeList, SyntaxRemoveOptions.KeepNoTrivia)!;
+                    currentRoot = RemoveAttributeListPreservingComments(currentRoot, attributeList);
                 }
                 else
                 {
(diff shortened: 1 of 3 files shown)


thomhurst:
Thanks for the review. Addressed in 5343b0c6a:

- **Directive list duplication / missing kinds** — `IsMeaningfulTrivia` now uses Roslyn's `SyntaxTrivia.IsDirective`, which covers `#line`, `#line span`, `#warning`, `#error`, etc. automatically. Did not touch `BaseMigrationCodeFixProvider.IsPreprocessorDirective` — it predates this PR and changing it expands scope beyond the bug fix; happy to do it as a follow-up.
- **Implicit-slice intent** — added a brief comment above the `Skip(first).Take(...)` line.
- **MSTest coverage** — added `MSTest_TestClass_Attribute_Removed_Preserves_Leading_Comments_And_Docstring` exercising the shared `RemoveAttributeListPreservingComments` path.
- **`trackedList` rename** — left as-is. The variable name is accurate (it is the tracked node) and you noted the logic is correct, so a rename is pure churn.

All migration tests still pass (NUnit 548, MSTest 156, XUnit 268).

thomhurst:
Thanks for the second pass. Addressed in 31c1ce111:

- **Redundant `nextToken` from two trees** — dropped. The `== default` guard now runs against the tracked-tree token, so there is no cross-tree variable.
- **Multi-attribute scenario** — added `NUnit_TestFixture_Attribute_Removed_Followed_By_Another_Attribute_List_Preserves_Comments` covering `// comment` + `[TestFixture]` + `[Category(...)]`. Confirms preserved trivia lands on the next attribute list's open bracket.
- **EOL-extension edge case** — added a clarifying comment instead of a code change. I tried a synthetic-EOL fallback first and it broke the docstring tests: `SingleLineDocumentationCommentTrivia` embeds its own trailing EOL, so the next trivia is whitespace (not EOL) and a fallback inserts a phantom blank line. The branch only matters for `//` / `/* */` / directives, which the parser always follows with explicit `EndOfLineTrivia` — so the existing condition is correct.
- **xUnit equivalent test** — pushing back: `XUnitTwoPhaseAnalyzer.XUnitRemovableAttributeNames` is empty (xUnit migration converts attributes, never removes a sole one), so an xUnit test would exercise no code path. The shared helper is now covered by NUnit (3 tests) and MSTest (1 test).
- **Consolidating `BaseMigrationCodeFixProvider.IsPreprocessorDirective`** — agreed, separate PR.

Tests: NUnit 672, MSTest 212, XUnit 328 passing.

PR #5741, opened 2026-04-25
perf(build): trim test TFMs and skip viewer dump by default

### LINKED ISSUE #5646: Issues trying to contribute
### Problem Statement

Hi Tom,

I'm trying to contribute as I believe I see a way to get realtime logging done in Resharper... but:

1. `dotnet build TUnit.slnx` is taking like 20+ minutes? I actually haven't been able to get through one dotnet build yet.. and I can't seem to run other instances of dotnet build as it like hangs the others? I feel like I canceled it but now dotnet build or restore for other solutions unrelated to TUnit are affected. What is actually occurring during this, and why is it taking so long? Is it necessary?
2. I can't open TUnit.slnx in Visual Studio without Playground not loading correctly. It looks like maybe there's a buildtarget that is needed for Playground but I can't seem to get dotnet build to finish so...

Thanks

### Proposed Solution

*

### Alternatives Considered

_No response_

### Feature Category

Test Discovery / Attributes

None

### Additional Context

_No response_

### Contribution

thomhurst:
Closes #5646.

## Summary

- **Test project default TFM trimmed to `net10.0`.** `TestProject.props` previously forced every test project to multi-target `net472;net8.0;net9.0;net10.0` — 4× compile work for tests that didn't actually exercise legacy frameworks. Three projects that snapshot output across consumer TFMs (`TUnit.Core.SourceGenerator.Tests`, `TUnit.Assertions.SourceGenerator.Tests`, `TUnit.PublicAPI`) keep the legacy set explicitly.
- **`EmitCompilerGeneratedFiles` is now opt-in** via `-p:EmitCompilerGeneratedFiles=true`. Default-off saves significant disk I/O on every test project build. The `<Compile Remove="SourceGeneratedViewer\**" />` exclusion stays unconditional so stale folders from prior opt-in builds don't poison the implicit C# glob (this was a real footgun — see `TestProject.props` comment).
- **`GenerateDocumentationFile` gated on packing/CI.** Local incremental dev builds skip XML doc emission; CI and `dotnet pack` keep it on. `SignAssembly` left untouched to avoid `InternalsVisibleTo` / `Verify.TUnit` strong-name breakage.
- **`Playground/` deleted.** Scratch project, not referenced by any build target. Was the cause of the VS load failure mentioned in #5646 (no `TargetFramework` / `OutputType`).
- **`CONTRIBUTING.md`** gains a "Building TUnit Locally" section pointing new contributors at `TUnit.Dev.slnx` (44 projects vs 96 in `TUnit.slnx`), `MSBUILDUSESERVER=1`, and the new `EmitCompilerGeneratedFiles` opt-in.
- **CI**: AOT publish step gets `--no-restore` (restore already ran earlier in the job).

## Test plan

- [x] `dotnet build TUnit.Dev.slnx -c Release -graphBuild:True` → 0 errors, ~60s on local Windows
- [x] `dotnet test TUnit.Core.SourceGenerator.Tests` → 464 pass / 4 skip / 0 fail across `net472;net8.0;net9.0;net10.0`
- [x] `dotnet test TUnit.Assertions.SourceGenerator.Tests` → 223 pass / 0 fail across all 4 TFMs
- [x] `dotnet test TUnit.PublicAPI` → 15 pass / 0 fail across all 4 TFMs
- [x] No `*.received.txt` files generated
- [ ] CI green across ubuntu/windows/macos
- [ ] AOT publish step still produces a working binary


thomhurst:
Thanks for the review.

**1. `TUnit.Dev.slnx` / `TUnit.CI.slnx` referencing the deleted Playground:** verified — neither file ever listed `Playground/Playground.csproj`, only `TUnit.slnx` did, and that reference is removed in this PR. `grep -r Playground *.slnx` returns no matches on this branch.

**2. Stale `SourceGeneratedViewer\` on opt-out:** good catch. Fixed in 9be994c85 by dropping the `EmitCompilerGeneratedFiles == 'true'` guard from the `CleanSourceGeneratedViewer` target. Now:
- opt-in (`-p:EmitCompilerGeneratedFiles=true`): fresh dump per build (unchanged)
- opt-out (default): one-time cleanup if a stale folder exists from a prior opt-in run, then no-ops on subsequent builds
- CI: still skipped via `ContinuousIntegrationBuild != 'true'` — ephemeral runners have nothing to clean

Verified `dotnet build TUnit.Dev.slnx -c Release -graphBuild:True` is green.

**3. `--no-restore` implicit coupling on AOT publish:** noted — it's a sequential dependency on the prior `dotnet restore TUnit.CI.slnx` step. Leaving as-is for now (standard CI flow), but a comment in the workflow flagging the dependency would be cheap to add if you'd like — happy to fold that in.

PR #5744, opened 2026-04-25
perf(engine): cache TestNode for non-final states (#5717)

### LINKED ISSUE #5717: perf: cache non-final TestNode PropertyBag instead of rebuilding per state transition

thomhurst:
Closes #5717

## Summary
- Cache Discovered + InProgress TestNode on TestContext, reused across MTP messages
- Terminal-state TestNode still constructed fresh (result data varies)
- ~2 PropertyBag allocs per test removed, less message-bus pressure

## Test plan
- [x] Engine builds clean on net10.0
- [x] Test slice executes and reports correct discovery/in-progress/terminal nodes

### DIFF
--- a/TUnit.Core/TestContext.cs
+++ b/TUnit.Core/TestContext.cs
@@ -433,15 +433,6 @@ public partial class TestContext : Context,
 
     internal ConcurrentDictionary<string, object?> ObjectBag => _testBuilderContext.StateBag;
 
-    // Cached MTP TestNode instances for non-final lifecycle states (Discovered, InProgress).
-    // These nodes are immutable for a given TestId, so we build them once and reuse across the
-    // three message-bus publishes per test. Stored as object? to avoid pulling the
-    // Microsoft.Testing.Platform dependency into TUnit.Core; the engine casts to TestNode on read.
-    // Final-state nodes (Passed/Failed/Skipped/Cancelled/Timeout) are still built fresh because
-    // their PropertyBag carries result-specific data (timing, output, exception).
-    internal object? CachedDiscoveredTestNode { get; set; }
-    internal object? CachedInProgressTestNode { get; set; }
-
     internal AbstractExecutableTest InternalExecutableTest { get; set; } = null!;
 
     internal SortedList<int, HashSet<object>> TrackedObjects { get; } = new();
--- a/TUnit.Core/TestContext.cs
+++ b/TUnit.Core/TestContext.cs
@@ -433,6 +433,15 @@ public partial class TestContext : Context,
 
     internal ConcurrentDictionary<string, object?> ObjectBag => _testBuilderContext.StateBag;
 
+    // Cached MTP TestNode instances for non-final lifecycle states (Discovered, InProgress).
+    // These nodes are immutable for a given TestId, so we build them once and reuse across the
+    // three message-bus publishes per test. Stored as object? to avoid pulling the
+    // Microsoft.Testing.Platform dependency into TUnit.Core; the engine casts to TestNode on read.
+    // Final-state nodes (Passed/Failed/Skipped/Cancelled/Timeout) are still built fresh because
+    // their PropertyBag carries result-specific data (timing, output, exception).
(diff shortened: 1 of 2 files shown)


thomhurst:
Thanks for the careful review — verified and you're correct on every point. Pushed `9f6b589` reverting the cache.

**Verification of the dead-code claim:**
- `TestRequestHandler.HandleDiscoveryRequestAsync` (line 48) is the only `Discovered()` call site; the comment at line 68 in `HandleRunRequestAsync` confirms execution mode skips it.
- `TestCoordinator.ExecuteTestAsync` (line 61) is the only `InProgress()` call site, and it sits **outside** the `RetryHelper.ExecuteWithRetry` loop — retries do not republish.
- `IdeStreamingSink.CreateHeartbeatNode` / `CreateOutputUpdateNode` build `InProgress` `TestNode`s directly with `new PropertyBag(...)` and publish via `PublishOutputUpdate(testNode)`, bypassing `ToTestNode()` entirely.
- Repeats (`RepeatCount`) build distinct `TestContext` instances per iteration in `TestBuilderPipeline`, so each gets its own (cold) cache entry.

So the `Discovered`/`InProgress` cache slots are written-once, never read. Confirmed.

**On Option C (pre-size `PropertyBag` via `EstimateCount`):** also doesn't help here — `PropertyBag` (MTP 2.2.1) stores its state-property and one general property in dedicated fields (`_testNodeStateProperty`, `_property`) and uses a linked-list node (`_property.Next`) for additional entries. There's no internal `List<>` to pre-size, so a count-hint constructor wouldn't change the allocation profile. (`EstimateCount` itself is pre-existing dead code from before this PR — left it alone since it's out of scope.)

Reverted back to the prior shape. PR is now a no-op vs main and ready to close.

thomhurst:
Closing — investigation confirmed the reviewer's analysis: the cache is unreachable in any production path.

## Verified

- `TestRequestHandler.cs:48` — only `Discovered()` site (discovery-mode only)
- `TestRequestHandler.cs:68-69` — explicit comment that execution mode skips Discovered
- `TestCoordinator.cs:61` — only `InProgress()` site, outside the retry loop
- `RetryHelper` — retries never republish InProgress
- `IdeStreamingSink.cs:217-249` — IDE heartbeat/output path uses `PublishOutputUpdate(TestNode)` directly, bypassing `ToTestNode()`
- `TestBuilderPipeline` repeats produce distinct `TestContext` instances → fresh cold cache each time

So both cache slots are written-once and never read.

## Option C also vacuous

The reviewer's "pre-size PropertyBag with EstimateCount" suggestion is not actionable: PropertyBag in MTP 2.2.1 has no internal List/array — it's a linked-list of properties with no `Capacity`. Confirmed via reflection on the shipped DLL.

The branch has been reverted to an empty diff vs main. Nothing to merge here. Issue #5717 closed as no-op.

PR #5758, opened 2026-04-26
fix(fscheck): register default CancellationToken arbitrary that surfaces TestContext token

### LINKED ISSUE #5757: [Bug]: FsCheckProperty CancellationToken parameter is pre-cancelled on roughly half of invocations
### Description

A property method declared with [Test, FsCheckProperty] that accepts a CancellationToken parameter receives a token that is pre-cancelled on roughly half of invocations, silently defeating [Timeout] cooperation.

FsCheck's built-in reflection-based arbitrary for CancellationToken invokes the public CancellationToken(bool canceled) constructor with a randomly generated bool, producing a pre-cancelled token whenever that bool is true.

### Expected Behavior

The CancellationToken parameter should be the timeout-backed token exposed by TestContext.Current.Execution.CancellationToken, so that property bodies can honour [Timeout] cooperatively the same way other TUnit test methods do.

### Actual Behavior

The token's IsCancellationRequested is true on roughly 50% of property invocations. Awaits using the token (e.g. Task.Delay(N, ct)) throw OperationCanceledException immediately, regardless of [Timeout] settings.

### Steps to Reproduce

1. Add a TUnit.FsCheck test class with the property below.
2. Run via `dotnet run -- --treenode-filter "/*/*/Repro/*"`.
3. Observe that the test fails on roughly half of MaxTest invocations.

```csharp
using FsCheck;
using TUnit.Core;
using TUnit.FsCheck;

public class Repro
{
    [Test, FsCheckProperty(MaxTest = 100)]
    public bool CtShouldNotBePreCancelled(CancellationToken ct)
    {
        return !ct.IsCancellationRequested;
    }
}
```

### TUnit Version

1.39.0

### .NET Version

.NET 8.0 / 9.0 / 10.0

### Operating System

Windows

### IDE / Test Runner

```shell

```

### Additional Context

FsCheck version: 3.3.2 (current TUnit pin in Directory.Packages.props).
Reproduces in both source-generator and reflection modes.
Not AOT-specific. The bug is in FsCheck's default CancellationToken arbitrary registration, before any TUnit timeout setup.

### IDE-Specific Issue?

JohnVerheij:
## Description

`FsCheckPropertyTestExecutor.CreateConfig` now registers a default `Arbitrary<CancellationToken>` that returns the timeout-backed token from `TestContext.Current`. Without this, FsCheck's built-in reflection-based arbitrary invokes the `CancellationToken(bool canceled)` constructor with a randomly generated bool, producing a pre-cancelled token on roughly half of invocations and silently defeating `[Timeout]` cooperation.

This applies the data-source-generator-attribute approach @thomhurst sketched in #2125 to address the lifecycle constraint stated in #3803 (*"test data is constructed before tests so doesn't have a test context available yet"*). The fix registers an attribute-shaped arbitrary that resolves the token at property-invocation time rather than at discovery.

Three artifacts:

- New `internal static CancellationTokenArbitrary` with a static factory. Annotated via `[DynamicDependency]` on `CreateConfig` for AOT/trimming preservation.
- `CreateConfig` registers user-supplied arbitraries first, then appends the default. FsCheck's `WithArbitrary` resolves earlier entries with higher priority, so user overrides win on conflict.
- Seven regression tests in `TUnit.Example.FsCheck.TestProject/CancellationTokenBehaviourTests.cs`.

**Alternatives considered:**

| Approach | Tradeoff |
|---|---|
| TestContext CT (this PR) | Full `[Timeout]` cooperation |
| `CancellationToken.None` always | Stops the pre-cancellation surprise but no `[Timeout]` |
| Analyzer exemption for FsCheck methods | Hides the symptom, doesn't fix the root cause |

Happy to reshape if a different approach is preferred.

**Out of scope:** No analyzer changes (TUnit0015's advice "add a `CancellationToken`" becomes correct for FsCheck properties after this fix). No public API changes (`CancellationTokenArbitrary` is `internal static`).

**Breaking changes:** None. Users who supply their own `Arbitrary<CancellationToken>` via `FsCheckPropertyAttribute.Arbitrary` keep their override because user registrations appear first in the type list. Verified by the regression test `UserSuppliedArbitraryOverridesDefault`.

## Related Issue

Fixes #5757

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

### Required

- [ ] **Snapshot Tests**: If I changed source generator output or public APIs:
- [x] **AOT Compatibility**: If this change uses reflection:
  - [x] I added appropriate `[DynamicallyAccessedMembers]` annotations

### Testing

## Additional Notes

The change lives entirely in `FsCheckPropertyTestExecutor.CreateConfig`, which is shared by both source-generator and reflection engine modes; Dual-Mode and Snapshot-Tests checkboxes are not applicable. AOT preservation uses `[DynamicDependency(DynamicallyAccessedMemberTypes.PublicMethods, typeof(CancellationTokenArbitrary))]` on `CreateConfig`; this is the appropriate annotation for the `typeof(...)` flow into FsCheck's `WithArbitrary(IEnumerable<Type>)`. The repo's `IsAotCompatible=true` (in `Library.props` for net8+) runs the AOT analyzers during normal build; my build produced no new IL warnings on net8.0/net9.0/net10.0. Tested locally on net8.0/net9.0/net10.0 in source-generator mode (49/49 passes across 7 stability iterations). Local `dotnet publish -p:PublishAot=true` hits NETSDK1207 because TUnit's library set multi-targets to netstandard2.0, which doesn't support AOT; CI's AOT lane will exercise that path.


JohnVerheij:
Thanks for the review. Addressed each point in 4ad9e9320:

1. **Token capture timing comment**: added inside `CancellationToken()`.

2. **Excessive XML docs**: stripped the `<remarks>` block on `CancellationTokenArbitrary`; kept only the brief type-level `<summary>`.

3. **Empty array initializer**: switched `new Type[0]` to `[]` (collection expression), and applied the same modernisation to the non-empty `Arbitrary = [typeof(...)]` cases for consistency within the test file.

4. **Timing window**: tightened from `<= 2_000` to `<= 1_500` (3x the configured 500ms timeout). Still CI-safe locally (49/49 passes across 7 stability iterations); sharper signal if cancellation propagation regresses.

5. **Per-test XML docs**: dropped the redundant `<summary>` blocks (method names self-document the assertion). Kept the class-level summary, the `AlwaysNoneTokenArbitrary` doc (cross-references the consumer test), and the `CooperativeCancellationViaTimeout` doc (timing mechanism is non-obvious).

Regarding the method-naming nit: yes, intentional. It matches the FsCheck convention demonstrated by `PositiveIntArbitrary.PositiveInt()` in the same project.

JohnVerheij:
CI is hitting a main-branch regression (`CollectionAssertionTests.cs:270`, also failing #5728). Holding for the upstream fix.

thomhurst:
Yep my bad. On it 😄

thomhurst:
Thanks!

PR #5764, opened 2026-04-26
Fix item-at Satisfies source typing

### LINKED ISSUE #5706: .HasItemAt(i).Satisfies() loses specialised assertion source for item
Related: #5703, #5704, #5705 (same pattern).

## Problem

`.HasItemAt(index).Satisfies(item => ...)` passes a generic `IAssertionSource<TItem>` into the lambda. When `TItem` is a `string`, collection, dictionary, or set, specialised assertion methods are unreachable.

## Repro

```csharp
// IList<string>
await Assert.That(list).HasItemAt(0).Satisfies(s => s.Contains(\"foo\")); // fails: string-specific Contains not on IAssertionSource<string>

// IList<List<int>>
await Assert.That(listOfLists).HasItemAt(0).Satisfies(l => l.HasCount(3)); // fails
```

## Affected signatures

- `TUnit.Assertions/Conditions/ListAssertions.cs:126` `Satisfies(Func<IAssertionSource<TItem>, Assertion<TItem>?> assertion, ...)`
- `TUnit.Assertions/Conditions/ReadOnlyListAssertions.cs:129` — mirror signature

## Proposed fix

Add specialised `Satisfies` overloads on the item-at helper type, keyed on `TItem` shape (string / IEnumerable / IDictionary / ISet / ...), passing the specialised source into the lambda. Use `OverloadResolutionPriority` for compiler steering.

## Acceptance

- Specialised assertions reachable inside the `.HasItemAt(i).Satisfies(...)` lambda for string/collection/dict/set items.
- Same on `ReadOnlyListAssertions` path.
- Tests cover each specialised element shape, on both list and read-only-list sources.

thomhurst:
## Summary
- add specialised ItemAt(...).Satisfies(...) overloads for collection-like item values
- keep generic Satisfies behavior while allowing type-erased specialised assertions internally
- add issue #5706 regression coverage and public API snapshots

## Verification
- dotnet test TUnit.Assertions.Tests\TUnit.Assertions.Tests.csproj --framework net10.0 --no-restore -- --treenode-filter "/*/*/Issue5706Tests/*"
- dotnet test TUnit.PublicAPI\TUnit.PublicAPI.csproj --no-restore

Closes #5706


thomhurst:
Addressed the review feedback in ce722f924:

- moved specialised Satisfies construction behind source-owned CreateSatisfiesAssertion helpers
- removed the exposed InternalListContext/InternalIndex accessors
- added smoke coverage for array, concrete Dictionary, HashSet, read-only dictionary, and read-only set item overloads

Verification rerun:
- dotnet test TUnit.Assertions.Tests\TUnit.Assertions.Tests.csproj --framework net10.0 --no-restore -- --treenode-filter "/*/*/Issue5706Tests/*" (14/14 passed)
- dotnet test TUnit.PublicAPI\TUnit.PublicAPI.csproj --no-restore (15/15 passed after cleaning stale PublicAPI outputs)

PR #5785, opened 2026-04-28
feat: add TUnit.Assertions.Should package

thomhurst:
## Summary

Adds `TUnit.Assertions.Should`, a source-generated FluentAssertions-style `Should()` API layered on top of `TUnit.Assertions` with zero duplication of assertion logic. Every `[AssertionExtension]` class in the referenced assembly gets a Should-flavored counterpart (`IsEqualTo`→`BeEqualTo`, `Contains`→`Contain`, `IsNotNull`→`NotBeNull`, `Throws`→`Throw`, etc.) emitted into `TUnit.Assertions.Should.dll` at build time.

## What's in the box

- **Runtime** (`TUnit.Assertions.Should`):
  - `IShouldSource<T>` marker, `ShouldSource<T>` (struct entry for values), `ShouldCollectionSource<T>` (collection entry exposing element-typed instance methods `BeInOrder`/`BeInDescendingOrder`/`All`/`Any`/`HaveSingleItem`/`HaveDistinctItems`), `ShouldDelegateSource<T>` (delegate entry with `Throw<E>`/`ThrowExactly<E>`).
  - `ShouldAssertion<T>` per-call wrapper — awaitable, exposes `.And`/`.Or` returning `ShouldContinuation<T>` so chains stay Should-flavored throughout.
  - `Should()` extension overloads on `T`, `string`, `IEnumerable<T>`, `Action`, `Func<T?>`, `Func<Task>`, `Func<Task<T?>>` — entry-overload narrowing mirrors `Assert.That`'s pattern (priority-resolved, `string` doesn't bind to `IEnumerable<char>`, etc.).
  - `[ShouldName("BeFoo")]` — per-class override hatch for irregular names. The Should generator's name-conjugation rule turns `Is*`→`Be*`, `IsNot*`→`NotBe*`, `Has*`→`Have*`, `DoesNot*`→`Not*`, `Does*`→[strip prefix], and drops the 3rd-person `-s`/`-es` ending; classes whose names land awkwardly under those rules opt out per-class with `[ShouldName]`. Negated forms (`[AssertionExtension(NegatedMethodName = "...")]`) get their own conjugation pass and can be overridden by placing a separate `[ShouldName]` on the negated assertion class.
- **Source generator** (`TUnit.Assertions.Should.SourceGenerator`):
  - Cross-assembly metadata scan for `IAssertionSource<T>` extension methods. Per-reference walks cached by `MetadataReference` instance via `ConditionalWeakTable` so Roslyn drops stale entries automatically when references change.
  - Conjugation table: `Is`→`Be`, `IsNot`→`NotBe`, `Has`→`Have`, `DoesNot`→`Not`, `Does`→[strip], 3rd-person `-s`/`-es` drop. CamelCase boundary-aware so `Issue` doesn't become `Besue`.
  - `[ShouldGeneratePartial(typeof(...))]` marks Should-flavored wrapper classes (e.g. `ShouldCollectionSource<TItem>`) for which the generator emits instance-method partials covering each public instance method on the wrapped TUnit assertion type.
  - Forwards `[Obsolete]` and `[EditorBrowsable]` from `[AssertionExtension]` and `[GenerateAssertion]` source methods through to the Should-flavored API surface so deprecation/visibility decisions stay consistent across both entry styles.
  - Skips referenced assertions whose `Should{Name}Extensions` is already baked in another reference (prevents CS0121 duplicates) — user-authored assertions still emit fresh.
  - Honors `[RequiresUnreferencedCode]` propagation from class & ctor levels.
  - Incremental-cache friendly via `EquatableArray<T>` wrappers; per-compilation lookups hoisted to one-time hash sets.
  - Shares `CovarianceHelper` and `AttributeForwardingFormatters` with `TUnit.Assertions.SourceGenerator` via linked compile.

## Test plan

- [x] 103 Should consumer tests pass on net10.0
- [x] 34 Should source generator tests (5 scenarios × 3 TFMs `.verified.txt` snapshots) pass on net8/9/10
- [x] Value, string, collection (List/Array/HashSet/Stack/Queue), DateTime/DateTimeOffset/DateOnly/TimeOnly/TimeSpan
- [x] Numeric: BeBetween, BeCloseTo, BeWithinPercentOf for int/long/decimal/float/double
- [x] Chaining: `.And` 2-deep / 3-deep, `.Or` short-circuit, mixed `And`/`Or` throws `MixedAndOrAssertionsException`, `Because` propagates to failure message
- [x] `Assert.Multiple` aggregation works
- [x] Delegate: `Throw<E>`, `ThrowExactly<E>` strictness, `Func<Task>` async, single-evaluation, no-exception-throws, generic exception type names render without backtick mangling
- [x] Element-type inference works without explicit generics for collection shapes (List, IReadOnlyCollection, HashSet, Stack, Queue, custom IEnumerable)
- [x] User-defined `[AssertionExtension]` class in test assembly gets generated Should counterpart
- [x] Failure messages include the Should-flavored method name in the expression chain
- [x] Range obsolete shim: `IsAll`→`BeAll` flagged with `[Obsolete]`, replaced by `IsFullRange`/`BeFullRange`
- [x] All 810 analyzer tests pass (TUnit.Analyzers.Tests, TUnit.AspNetCore.Analyzers.Tests, TUnit.Mocks.Analyzers.Tests, TUnit.Assertions.Analyzers.Tests, TUnit.Assertions.Analyzers.CodeFixers.Tests)
- [x] Full `TUnit.slnx` builds clean, no new warnings


thomhurst:
Addressed in 5b89e90:

- **HIGH 1 — `ShouldEntryPointAttribute` unread**: removed the attribute entirely. After re-checking, the generated extensions live on `IShouldSource<T>` (not `T`) and never could have collided with FluentAssertions' `Should()` — only the hand-written `Should()` overloads in `ShouldExtensions` collide, and an attribute can't suppress a `using` directive's bindings. The attribute's stated purpose wasn't achievable, so dropped rather than half-wired.
- **HIGH 2 — `EquatableArray` 4th copy**: deleted local copy, linked from `TUnit.Core.SourceGenerator/Models/EquatableArray.cs` via `<Compile Include="..." Link="Shared\..." />` (same pattern already in use for `CovarianceHelper`).
- **MED 3 — dead covariance code**: stripped the unreachable `if (data.NeedsMethodLevelCovariance)` branch from `EmitMethod`, removed `NeedsMethodLevelCovariance` / `TypeParamIsTypeParameter` / `TypeParamConstraintName` fields from `AssertionData`. Emission is now unconditionally `IShouldSource<typeParam>`.
- **MED 4 — collection mixing**: added `InstanceMethodAssertions` skip-set covering the seven assertion classes whose Should counterpart is a hand-crafted instance method on `ShouldCollectionSource<T>` (`CollectionIsInOrderAssertion`, `CollectionAllAssertion`, `HasSingleItemAssertion`, etc.). Generator skips them, so drift / shadowing isn't possible.
- **MED 5 — null propagation**: added `NullSourceTests` exercising `null` string, `null` list `.Contain`, `null` list `.BeInOrder`. All produce `AssertionException` (not NRE) — `CollectionAssertion` handles null internally as expected.
- **MIN 6 — `typeof(T).Name`**: kept as-is; matches the existing `FuncAssertion.Throws` / `AsyncDelegateAssertion.Throws` pattern in TUnit.Assertions. Worth a separate pass across all of TUnit, not Should-specific.

Test count: 90/TFM × 3 TFMs = 270 (was 261).

thomhurst:
Addressed in 2830fed:

**1. `TUnit.CI.slnx`** — added all four new projects (Should runtime + tests, SourceGenerator + tests) under matching folders to the dev solution. CI now builds, tests, packs, and publishes them.

**2. SourceGenerator tests** — added `ShouldExtensionGeneratorTests.cs` with five in-memory compilation cases:
- Simple `[AssertionExtension]`-style class → conjugated extension method
- Generic `<TValue>` assertion → method-level generic propagation
- `[ShouldName(...)]` override wins over auto-conjugation
- `IsNot*` prefix → `NotBe*`
- `[CallerArgumentExpression]` forwarded to the generated method

The harness compiles inline source snippets, runs the generator, and string-matches key tokens — same style as the existing `TUnit.Assertions.SourceGenerator.Tests`. Five tests, all pass on net8/9/10.

**3. `ShouldNameAttribute.Negated`** — removed. The property was parsed but always discarded; emitting a negated form requires correlating two extension methods to one assertion class, which the method-scan path doesn't do. The natural double-emission from `[AssertionExtension(NegatedMethodName = "...")]` already handles negated forms via independent auto-conjugation (`Contains` → `Contain`; `DoesNotContain` → `NotContain`). Attribute XML doc now points users to placing a separate `[ShouldName]` on the negated assertion class for the rare case where conjugation isn't enough.

thomhurst:
Addressed in 8f88cc4:

**Docs example** — `should-syntax.md` "Custom names" section rewritten. Removed the broken `Negated = "..."` example. Now shows the single-class override pattern (`BeAnOddNumber`) and explicitly documents the separate-class pattern for assertion families like `EqualsAssertion`/`NotEqualsAssertion` where positive and negated forms live on distinct classes (each with its own `[ShouldName]`).

**NameConjugator** — picked option 2: `Conjugate` now returns `string` directly. The `(Name, Matched)` tuple was advertising a `TUSHOULD001` diagnostic that was never implemented and never going to be — unrecognised names should pass through unchanged so users with novel naming conventions stay forward-compatible. Both call sites updated. Doc clarifies the new contract: layer `[ShouldName]` for one-off renames.

5 generator tests + 102 runtime tests still pass.

thomhurst:
Cleanup applied in dd7931a:

- Removed `Verify`, `Verify.TUnit`, `xunit`, `xunit.runner.visualstudio` from `TUnit.Assertions.Should.SourceGenerator.Tests.csproj` — template carry-over; the tests run via `Microsoft.Testing.Platform` only.
- Replaced the comment-only `if (!isCurrentAssembly)` block in `CollectWrapper` with a comment co-located on the unconditional `builder.Add(...)` site that actually does the work.

5 generator tests still pass.

thomhurst:
## Round 7 review — addressed in 2273360

### Fixed
- **`TryDropTrailingS` -es bug**: `MatchesRegex` was conjugating to `MatcheRegex`. `TryDropTrailingS` now recognises English -es third-person endings (sibilants ch/sh/ss/x/z plus -o) and drops both letters. Generic -es (e.g. `Writes`) still drops just the trailing s; plain -ss (e.g. `Pass`) still leaves the name unchanged.
- **`NameConjugator` unit tests**: 29 cases locking in the full conjugation table (Is/IsNot/Has/DoesNot/Does prefixes, the various -s/-es endings, plus boundary cases like `Issue`/`Pass`).
- **`CollectionContext` record → sealed class**: it's a mutable bag threaded through the namespace walk, not an incremental-pipeline cache key, so `record` was misleading.

### Pushed back on
- **`typeof(TException).Name`** in `ShouldDelegateSource`: this matches the existing pattern in `FuncAssertion.Throws<T>` in `TUnit.Assertions`. Keeping it consistent with the canonical surface; if it's wrong, it's wrong in both places and should be a separate change.
- **Parallel `ShouldSource` / `IShouldSource` / `IAssertionSource` hierarchies**: the API isolation is intentional. `IShouldSource<T>` only exposes Should-flavored extensions, so chains can't accidentally drop back into `Is*`/`Has*` mid-chain.
- **`IsShouldAssertionType` namespace check** in analyzers: defence-in-depth so we don't fire on unrelated `.Should()` extensions in other libraries (FluentAssertions, custom user code). Already stated in the docs as the contract.

### Deferred
- `CompilationProvider` → `SyntaxProvider` perf refactor, `ShouldCollectionSource.Clear()`, per-method `[ShouldName]`, dedup naming convention, analyzer test DLL workaround — none of these are blocking for v1; can land in follow-ups once the surface is stable.

All 29 NameConjugator + 5 generator tests + 102 consumer tests still pass.

thomhurst:
## Snapshot tests added — commit \`4c54f08\`

Round 7 reviewer flagged that the generator tests used token-level \`Contains\` assertions instead of Verify-based snapshots, against CLAUDE.md project policy.

### What changed
- Added \`Verify\` + \`Verify.TUnit\` to the test project, plus the standard \`GlobalSetup\` (\`DiffRunner.Disabled = true\`) and \`GlobalUsings\` matching the existing \`TUnit.Assertions.SourceGenerator.Tests\` infrastructure.
- \`RunGenerator\` now wraps the generated source through \`Verify(...).UniqueForTargetFrameworkAndVersion()\` after the existing \`Contains\` checks. The token assertions stay as explicit guard-rails on the user-visible API surface; the snapshots additionally lock whitespace, ordering, and \`global::\` prefixes.
- 15 \`.verified.txt\` files committed (5 tests × 3 TFMs).
- Test project switched from \`net472;net8.0;net9.0;net10.0\` to \`net8.0;net9.0;net10.0\`. Reasoning is documented in the csproj: the generator's output is determined by the input compilation rather than the consumer runtime, so cross-TFM snapshots are identical text — and the in-memory compilation harness can't resolve user-declared generic assertions reliably against the .NET Framework BCL, producing spurious mismatches with no signal.
- \`GetReferences()\` switched from the .NET-Core-only \`TRUSTED_PLATFORM_ASSEMBLIES\` lookup to \`AppDomain.CurrentDomain.GetAssemblies()\` (matching \`ReferencesHelper\` in the existing test project) so the harness works on every supported TFM.

All 34 tests pass on net8.0, net9.0, and net10.0.

thomhurst:
## Round 9 — addressed in \`33a903e\`

### Fixed
**\`typeof(TException).Name\` mangling for generic exception types.** Reviewer was right — \`typeof(MyException<int>).Name\` returns \`"MyException\`1"\`, which leaked into Should failure messages. Added a small \`FormatTypeName\` helper that strips the backtick-arity and recurses into generic arguments, so call-site expressions now read \`Throw<MyException<Int32>>()\` instead of \`Throw<MyException\`1>()\`. Added a regression test to \`DelegateTests\`. The inner assertion's \`Expected to throw…\` prefix still comes from TUnit's own \`ThrowsAssertion\` and uses \`Type.Name\` — that's a separate concern in the core library and out of scope for this PR.

### Pushed back (carry-over)
- **Parallel \`IShouldSource\`/\`IAssertionSource\` hierarchies.** This isn't maintenance debt — it's the load-bearing piece of the chain-isolation guarantee. \`IShouldSource<T>\` exists *specifically so* \`Should()\` chains can't accidentally drop back into \`Is*\`/\`Has*\` mid-chain. Making it inherit from \`IAssertionSource<T>\` (or sharing a wide common interface) defeats the entire reason for the second hierarchy. The "duplication" in analyzers is three small if-branches each — well within the cost of the architectural property they enable.
- **\`CompilationProvider\` perf.** Reviewer themselves marks it medium and notes the output is correct. The single-trigger pipeline is intentional for v1 — the input set is stable across source edits, so the Roslyn incremental cache absorbs most fires. A \`SyntaxProvider.ForAttributeWithMetadataName\` decomposition is a pre-GA optimisation, not a beta blocker. Tracking for follow-up.

All 103 Should consumer tests + 34 generator tests still pass.

thomhurst:
## Round 10 review — triaged in \`ce5dc38\`

### Verified false alarms — no change needed

**High: AOT incompatibility in \`FormatTypeName\`** — verified by building \`TUnit.Assertions.Should\` (\`<IsAotCompatible>true</IsAotCompatible>\`) across all TFMs: **0 trim/AOT warnings**. The reviewer cited \`Type.GetGenericArguments()\` but the implementation uses \`Type.GenericTypeArguments\`, a property accessor that returns the captured array of type arguments without performing reflection on the runtime types — neither annotated with \`[RequiresUnreferencedCode]\` nor flagged by \`IL2026\`/\`IL3050\`/etc. analyzers. No annotation needed.

**High: \`[OverloadResolutionPriority]\` on \`netstandard2.0\`** — verified by explicit \`-f netstandard2.0\` build of \`TUnit.Assertions.Should\`: **0 warnings, 0 errors**. The repo's \`Directory.Build.props\` includes the \`Polyfill\` package as a \`PrivateAssets="all"\` reference for every internal project, which provides \`OverloadResolutionPriorityAttribute\` for older TFMs. The string overload's priority is correctly enforced on \`netstandard2.0\`.

**Low: Missing test for \`Func<T>\` sync source** — already exists at \`TUnit.Assertions.Should.Tests/DelegateTests.cs:54\` (\`FuncT_evaluates_only_once\` uses \`Func<int>\` and verifies single-evaluation under chained assertions).

### Addressed (clarifications, no behavior change)

- **Medium: \`FindInvocationInChain\` parentName: null docs.** Added a call-site comment on \`FindShouldInChain\` explaining that constraining \`parentName\` would break the suppressor for user-defined entry points. The helper itself was already documented; the call site now matches.
- **Medium: \`ExpressionBuilder.Clear()\` fragility.** The behaviour is already covered by \`CollectionTests.Failure_message_uses_Should_flavored_expression\`, which fails immediately if \`CollectionAssertion\`'s base ctor ever appends additional content beyond the seed. Cross-referenced the test from the comment so the safety net is discoverable from the call site.
- **Low: \`DoesRun\` → \`Run\` conjugation gap.** Added a paragraph in \`should-syntax.md\` documenting that the \`Does*\` strip rule can produce awkward names for non-verb auxiliaries, and that \`[ShouldName]\` is the supported escape hatch. (Adding a word-list-driven check in the conjugator would expand its English-grammar surface significantly for marginal benefit; the override is the right level.)

### Pushed back (carry-over)

- **Critical: \`CompilationProvider\` perf.** Reviewer is escalating this each round. \`Collect\` returns \`EquatableArray<MethodData>\` + \`EquatableArray<WrapperData>\` — both fully equatable — so when no \`[AssertionExtension]\` types change in either current or referenced compilations, \`RegisterSourceOutput\` does not re-fire and no source is re-emitted. The walk itself runs per-keystroke but visits cheap symbol checks (\`SymbolEqualityComparer.Default.Equals\` against pre-resolved sentinel symbols, no reflection, no allocation in the hot fields). A \`SyntaxProvider.ForAttributeWithMetadataName\` decomposition is the right pre-GA optimisation but doesn't affect correctness; tracking for follow-up.

103 consumer + 34 generator tests still pass.

thomhurst:
## Round 11 — addressed in \`a3cba8e\`

### Critical: \`CompilationProvider\` perf — fixed

The generator's pipeline still triggers from \`CompilationProvider\` (necessary because per-reference symbol resolution requires a \`Compilation\` context — \`MetadataReferencesProvider.Combine(CompilationProvider)\` re-fires the transform per keystroke too because the combined input changes whenever the compilation does). Instead, the actual cost — the cross-assembly walk — is now gated by a static \`ConcurrentDictionary<MetadataReference, ReferenceData>\` keyed on the \`MetadataReference\` instance. Roslyn typically reuses the same \`MetadataReference\` across compilations until the underlying assembly is rebuilt, so once each reference has been scanned its data is reused for every subsequent keystroke. Cache entries store raw walk results; the union dedup is applied at merge time so cache validity is not sensitive to changes in other references or in the current compilation.

Per-keystroke cost reduces to: a O(1) \`MetadataReference\` lookup per relevant reference, plus the unchanged walk of the current compilation (which has to run because the user is editing it).

### Medium: \`FormatDefaultValue\` — fixed

Numeric defaults now emit the C# type suffix: \`1.5F\` (float), \`1.5D\` (double), \`1.5M\` (decimal), \`1L\` (long), \`1UL\` (ulong), \`1U\` (uint). The catch-all path also goes through invariant culture so comma-decimal locales can't produce malformed literals like \`1,5F\`. No assertion in TUnit currently has a numeric-typed default, but adding one would have silently produced uncompilable generated code — exactly the latent-bug class the reviewer flagged.

### Minor: \`IsShouldAssertionType\` redundancy — fixed

\`ShouldAssertion<T>\` implements \`IShouldSource<T>\`, so it's already matched by the \`isAssertionSource\` branch in \`MixAndOrOperatorsAnalyzer\`. Removed the redundant check and the helper.

### Pushed back

**Medium: \`ExpressionBuilder.Clear()\` mutation in \`ShouldCollectionSource\`.** The reviewer's proposed fix — accept a pre-built \`StringBuilder\` externally — requires a new public constructor on \`CollectionAssertion\` in \`TUnit.Assertions\`, which is a core-library API change out of scope for this PR. The hypothetical risk ("if \`CollectionAssertion\` ever wraps the builder…") is guarded by \`Failure_message_uses_Should_flavored_expression\`: any drift in \`CollectionAssertion\`'s base ctor surfaces in that test. If/when \`CollectionAssertion\`'s API surface is reviewed for v2, the constructor overload is the right place to address this.

All 34 generator + 103 consumer tests pass. The pre-existing \`MixAndOrOperatorsAnalyzerTests.Flag_When_Mixing_And_With_Or\` failure on main is unrelated to this PR (verified by stashing the round-11 changes and re-running — same failure on the unmodified branch base).

thomhurst:
## Reviewer follow-up — addressed in \`6bd1516\`

Reviewer flagged the static \`ConcurrentDictionary<MetadataReference, ReferenceData>\` cache as an unbounded leak: when Roslyn replaces a \`MetadataReference\` after a dependency rebuild, the dictionary pins the old key→value pair indefinitely. CI is unaffected (process exits) but long IDE sessions with frequent rebuilds would accumulate stale entries.

Swapped to \`ConditionalWeakTable<MetadataReference, ReferenceData>\`. Weak keys mean entries become GC-eligible the moment Roslyn drops the reference. Same O(1) lookup semantics; the \`Add\` (no \`TryAdd\` overload in netstandard2.0) is wrapped in \`try/catch ArgumentException\` to absorb concurrent races between two compilations seeing the same uncached reference (both compute identical \`ReferenceData\`; first writer wins).

Also pulled the per-reference scan into a separate \`ScanReference\` helper so \`GetOrComputeReferenceData\` stays focused on cache lookup + insertion.

All 34 generator + 103 consumer tests still pass.

### Other items in the review (already-✅)

The reviewer also flagged a "minor caveat" on the analyzer-test fix: if the netstandard2.0 build hasn't been produced yet (e.g. partial clean build), the helper falls back to \`typeof(...).Assembly.Location\` which silently reintroduces CS1705. That's an acknowledged tradeoff — fail-fast would hurt CI ergonomics on first builds. The csproj copies the netstandard2.0 build via \`PreserveNewest\` from the project's bin, which is reliably present after the standard build chain. Left as-is.

thomhurst:
## Round 13 — addressed in \`8a726e9\`

### Fixed: Should test modules wired into the release gate

\`RunEngineTestsModule\` (the gate before \`UploadToNuGetModule\`) now \`[DependsOn]\` both:
- \`RunAssertionsShouldTestsModule\` (the consumer tests, already existed but wasn't gated)
- \`RunAssertionsShouldSourceGeneratorTestsModule\` (newly added — covers \`TUnit.Assertions.Should.SourceGenerator.Tests\` across net8/9/10)

A Should regression in any of the three layers (consumer behaviour, generator output, or analyzer) now blocks a release.

### Fixed: doc variable named \`async\`

Renamed \`async\` to \`asyncFunc\` in the delegate-entry example in \`docs/should-syntax.md\`. Avoids the contextual-keyword shadow that some editors flag with CS1948.

### Fixed: redundant \`value!\` suppressor

\`ShouldExtensions.Should(this string?)\` was passing \`value!\` into \`AssertionContext<string>\`, which accepts \`string?\` directly — the \`!\` suppressed a warning that doesn't exist and contradicted the parameter type. Zero behaviour change.

103 consumer + 34 generator tests still pass.

thomhurst:
## Round 14 — addressed in \`4769882\`

### Fixed: \`GetCompatibleDllPath\` centralized

Reviewer flagged the duplicated netstandard2.0 fallback helper across four analyzer test projects. Extracted to \`SharedTestHelpers/AnalyzerTestCompatibility.cs\` and linked into each test csproj via \`<Compile Include="..\SharedTestHelpers\..." Link="Shared\..." />\`. Matches the existing \`CovarianceHelper\`/\`EquatableArray\` linked-source pattern. The per-project wrappers (\`GetCompatibleCoreDllPath\` / \`GetCompatibleAssertionsDllPath\` / \`GetCompatibleShouldDllPath\`) now delegate; the resolution logic and its "why" comment live in one place.

### Documented (#3, #4): analyzer flow-insensitivity

Both \`AwaitAssertionAnalyzer.IsAwaited\` and \`IsNotNullAssertionSuppressor.WasAssertedNotNull\` are intentionally syntactic / statement-order based rather than control-flow aware. Documented the limitations:
- \`IsAwaited\`: false positive on split-variable patterns (\`var src = value.Should(); await src.X();\`) — the parent walk hits the declaration's block before any await.
- \`WasAssertedNotNull\`: false suppression when the assertion sits in one branch of an \`if\`/\`try\` and the usage runs unconditionally.

Both fixes would require dataflow analysis (\`Roslyn.IFlowAnalysis\`) — significant complexity for niche imprecision; the current behaviour is documented as a known limitation rather than a surprise.

### Pushed back (carry-over)

**\`ShouldCollectionSource\` ExpressionBuilder.Clear():** Reviewer's proposal — having \`AssertionContext\` accept an \`expressionSeed\` ctor parameter — is a public API change to TUnit.Assertions core, out of scope for this PR. The existing safety net (the \`Failure_message_uses_Should_flavored_expression\` test) catches any drift in \`CollectionAssertion\`'s base ctor immediately. If/when AssertionContext's API surface is re-examined for v2, the seed-via-constructor approach is the right place to land it.

**ShouldNameAttribute.Negated PR description drift:** Documentation nit; \`Negated\` was dropped during review (round 6) when the per-class pattern proved sufficient. PR description is now stale relative to the implementation. Will tidy when squashing.

103 consumer + 34 generator + 810 analyzer tests pass.

thomhurst:
## Round 15 — addressed in \`4f9f779\`

### Fixed (#1): PR description accuracy

Removed the stale \`Negated\` property and \`ShouldEntryPoint\` attribute references from the description. Added an accurate summary of the current shape (\`[ShouldName]\` per-class override hatch with the negated-class workaround, \`ConditionalWeakTable\` reference cache, generic-attribute → typeof workaround for CS8968, etc.).

### Fixed (#2): \`ShouldCollectionSource\` no longer mutates a borrowed \`AssertionContext\`

The \`_inner\` field held dead state and the seed-overwrite was a fragile coupling — both noted across rounds 11/12/14/15. The constructor now creates the \`AssertionContext\` directly:

\`\`\`csharp
var sb = new StringBuilder(...);
sb.Append(expression ?? "?").Append(".Should()");
Context = new AssertionContext<IEnumerable<TItem>>(value, sb);
\`\`\`

\`[ShouldGeneratePartial]\` now takes the wrapped type explicitly via \`typeof()\` so the source generator no longer scans private fields. C# 11 generic attributes were the natural fit but can't take an enclosing type parameter as a type argument (CS8968), so the open-generic form is the workaround — the generator substitutes the wrapper class's type parameters in <code>CloseWrappedType</code>. \`_inner\` deleted; the \`value!\` null-bang deleted with it.

The wrapper-vs-extension dedup that was filtering extensions by inner-assertion return type turned out to be over-aggressive: the simple-factory wrapper template only emits methods whose source overload exactly matches a public ctor, so overloads with optional/default parameters land only on the extension surface — and the filter was removing those even though instance methods take precedence at call sites. Removed; both surfaces co-exist by design.

### Fixed (#4): \`FormatObsolete\`/\`FormatEditorBrowsable\` centralized

Extracted into \`TUnit.Assertions.SourceGenerator/Shared/AttributeForwardingFormatters.cs\` and linked into both generator projects via \`<Compile Include>\`. The formatters can't drift now.

### Pushed back

- **#3 \`ShouldAssertion<T>\` as struct** — non-trivial perf optimization with care needed for the \`_inner\` reference field. Tracking as a follow-up; not blocking.
- **#5 \`SyntaxProvider\` for current compilation** — already pushed back across rounds 9–11. The \`ConditionalWeakTable\` cache absorbs the cross-assembly cost; current-compilation walk runs per keystroke but is bounded by user-defined assertion-extension count.
- **#6 \`FormatTypeName\` CLR vs display names** — runtime code, no Roslyn at runtime. \`Int32\` vs \`int\` in failure-message exception types is consistent with existing \`Type.Name\` usage in TUnit core (\`FuncAssertion.Throws<T>\`); changing it requires either a runtime keyword-mapping table or a separate fix landing in core first.
- **#7 \`FindShouldInChain\` asymmetry** — already documented in round 14: \`null\` parentName is intentional because \`Should()\` is an extension method with arbitrary receiver. The asymmetry vs \`FindAssertThatInChain\` reflects the structural difference (\`Assert\` is a static class).
- **#8 \`AnalyzerTestCompatibility\` MSBuild .targets** — round 14 already centralized the helper to a single shared linked file. Reviewer's further suggestion of a shared \`.targets\` file is stylistic — \`<None Include>\` + \`<Compile Include>\` works at the same scope and matches the existing \`CovarianceHelper\`/\`EquatableArray\` pattern.

103 consumer + 34 generator + 810 analyzer + 2104 TUnit.Assertions tests still pass.

thomhurst:
## Round 16 — addressed in \`46ba85a\`

### Fixed (#6): redundant \`GetCompatibleDllPath\` wrappers removed

The per-project wrappers (one in each of the 4 analyzer test projects) added nothing over the shared \`SharedTestHelpers/AnalyzerTestCompatibility.GetCompatibleDllPath\` helper they delegated to. Callers invoke the shared helper directly now; the wrappers and their stale doc comments are gone.

### Documented (#2, #3, #4, #5)

- **\`ShouldAssertion<T>\` as \`sealed class\`**: comment explains \`GetAwaiter\` forwards to \`Assertion<T>\`'s state-machine awaiter which mutates instance state during evaluation. A struct would either copy the mutation away from the receiver or have to box through \`IShouldSource<T>\`. Single per-call allocation is the conscious cost; entry types stay structs because they're pure routers.
- **\`AwaitAssertionAnalyzer.IsAwaited\` split-variable false positive**: comment spells out the shared limitation between \`Assert.That\` and \`Should()\` paths so users know it's a known imprecision.
- **\`ReferencesAssertionsAssembly\` one-level depth**: documented as safe because \`IAssertionSource<T>\` lives in TUnit.Assertions, so any assembly declaring extensions on it must directly reference TUnit.Assertions.
- **\`FormatTypeName\` CLR vs C# aliases**: documented the asymmetry vs the source-generator's display-format output — acceptable for failure messages on rare generic exception types.

### Pushed back / deferred

- **#1 SyntaxProvider for current compilation**: documented inline as a follow-up at the generator's \`Initialize\` entry. Cross-assembly cost is already absorbed by the \`ConditionalWeakTable\` cache; current-compilation walk re-runs per keystroke but is bounded by static-class count. Refactor is structurally non-trivial and not blocking for v1.
- **#7 Generator NuGet packaging**: out of scope for this PR — applies equally to the existing \`TUnit.Assertions\` analyzer/generator packaging pattern this PR follows.

103 consumer + 34 generator + 810 analyzer + 2104 TUnit.Assertions tests still pass.

PR #5806, opened 2026-05-01
fix(should): add specialized assertion sources

thomhurst:
## Summary
- add specialized Should() routing and wrappers for collections, dictionaries, sets, and collection-returning delegates so missing APIs like ContainKeyWithValue, BeSupersetOf, and HaveAtLeast are available
- fix ShouldExtensionGenerator deduplication and specialized receiver detection so baked entrypoints are not re-emitted in consumer projects
- update generator and public API snapshots and verify the focused Should/runtime/public API coverage

## Verification
- dotnet build "TUnit.Assertions.Should/TUnit.Assertions.Should.csproj" -f net10.0
- dotnet build "TUnit.Assertions.Should.SourceGenerator.Tests/TUnit.Assertions.Should.SourceGenerator.Tests.csproj" -f net10.0
- dotnet build "TUnit.Assertions.Should.Tests/TUnit.Assertions.Should.Tests.csproj" -f net10.0
- dotnet run --framework net10.0 --project "TUnit.Assertions.Should.SourceGenerator.Tests" -- --treenode-filter "/*/*/ShouldExtensionGeneratorTests/Assert_That*"
- dotnet run --framework net10.0 --project "TUnit.Assertions.Should.Tests" -- --treenode-filter "/*/*/CollectionTests/*Dictionary_ContainKeyWithValue|/*/*/CollectionTests/*HashSet_BeSupersetOf|/*/*/CollectionTests/*Func_collection_HaveAtLeast"
- dotnet run --framework net8.0 --project "TUnit.PublicAPI" -- --treenode-filter "/*/*/Tests/Should_Library_Has_No_API_Changes"
- dotnet run --framework net9.0 --project "TUnit.PublicAPI" -- --treenode-filter "/*/*/Tests/Should_Library_Has_No_API_Changes"
- dotnet run --framework net10.0 --project "TUnit.PublicAPI" -- --treenode-filter "/*/*/Tests/Should_Library_Has_No_API_Changes"


thomhurst:
Pushed 246251b1c addressing the review:

- **`BuildExpression` dedup**: extracted to a single `ShouldExpressionBuilder.Build` helper in `ShouldSourceBase.cs`; removed the six duplicates (collection/dictionary x2/set x3 sources) plus the local copy in `ShouldExtensions.cs`. The struct (`ShouldDelegateCollectionSource`) now also uses the shared helper.
- **`FormatTypeName` dedup**: extracted to `DelegateExceptionTypeFormatter.FormatTypeName` (kept the existing doc comment about the runtime-vs-Roslyn primitive-name asymmetry); both `ShouldDelegateSource<T>` and `ShouldDelegateCollectionSource<TItem>` now call into it.
- **`Because` struct-copy footgun**: added explicit XML-doc warnings on `ShouldDelegateSource<T>.Because` and `ShouldDelegateCollectionSource<TItem>.Because` calling out that the result MUST be consumed inline because the type is a `readonly struct`.
- **`GetWrapperMethodKey` hardcoded `'0'`**: added a comment tying it to the `TryDescribeWrapperMethod` filter that rejects methods with method-level type parameters — `WrapperMethodData` literally cannot carry any, so the literal is correct by construction. Switching to a derived count would have required adding a `MethodGenericParams` field that's always empty; the comment makes the invariant explicit instead.

**On point 2 (specialized return type), pushing back**: I think there's a misread of the code. `Assert.That(IReadOnlyDictionary<TKey,TValue>, ...)` returns `DictionaryAssertion<TKey,TValue>`, **not** `ShouldDictionarySource<TKey,TValue>` — `ShouldDictionarySource` only exists as a hand-coded `Should()` extension in `ShouldExtensions.cs`. The generator's `EmitShouldEntry` is the fallback for `Assert.That` overloads with no hand-coded `Should()` counterpart; for those, `DictionaryAssertion<…>` (an `IAssertionSource<T>`) would be the wrong return type for a Should-flavored entry point — wrapping in `ShouldSource<T>` is the intended convention. For dictionary/set/collection types where a richer Should source exists, the hand-coded overloads in `ShouldExtensions.cs` take priority; deduplication via `CollectExistingShouldEntryKeys` keeps the generator from emitting a competing wrapper. So the cast-and-wrap isn't a downgrade — it's the correct shape for the generated fallback.

Snapshot tests (46) and Should runtime tests (130) all green locally.

PR #5840, opened 2026-05-06
feat(playwright): add composition-based fixtures

thomhurst:
## Summary

Adds an additive composition-based fixture API to `TUnit.Playwright` as an alternative to the existing `PlaywrightTest` / `BrowserTest` / `ContextTest` / `PageTest` inheritance hierarchy. Users who already have their own base classes — or who want fine-grained control over which Playwright objects are shared at which scope — can now compose via `[ClassDataSource<T>]` instead of inheriting.

```csharp
public class MyTests
{
    [ClassDataSource<PageFixture>]
    public required PageFixture Page { get; init; }

    [Test]
    public async Task Loads() => await Page.Page.GotoAsync("https://example.com");
}
```

Each fixture implements `IAsyncInitializer`/`IAsyncDisposable` and references the next layer via a property attributed with `[ClassDataSource<T>(Shared = …)]`:

- `PlaywrightFixture` — owns `IPlaywright`
- `BrowserFixture` — `[ClassDataSource<PlaywrightFixture>(PerTestSession)]`, owns `IBrowser`, supports the Microsoft Playwright Service env-var contract
- `ContextFixture` — `[ClassDataSource<BrowserFixture>(PerTestSession)]`, owns `IBrowserContext`, performs W3C trace-header propagation
- `PageFixture` — `[ClassDataSource<ContextFixture>]`, owns `IPage`

### Multi-context use case

Two `[ClassDataSource<PageFixture>]` properties yield two distinct `IBrowserContext` instances (cookie / storage isolation) while sharing one `IBrowser` — useful for testing two accounts interacting:

```csharp
[ClassDataSource<PageFixture>] public required PageFixture Alice { get; init; }
[ClassDataSource<PageFixture>] public required PageFixture Bob { get; init; }
```

### Refactor

To avoid drift, two pieces of integration logic that previously lived only on the inheritance side were lifted to internal helpers and shared with both surfaces:

- `PlaywrightServiceConnector.LaunchAsync` — used by both `BrowserService.CreateBrowser` and `BrowserFixture.InitializeAsync`
- `PlaywrightTelemetryHeaders.Merge` — used by both `BrowserTest.NewContext` and `ContextFixture.InitializeAsync`

The existing `PlaywrightTest` / `BrowserTest` / `ContextTest` / `PageTest` classes are untouched in their public surface.

## Test plan

- [x] `dotnet build TUnit.Playwright/TUnit.Playwright.csproj` clean across `netstandard2.0;net8.0;net9.0;net10.0`
- [x] Smoke test added to the Playwright template project (`TwoContextFixtureTests.cs`): two `PageFixture` injections; sets `localStorage` in Alice, asserts Bob's `localStorage` is empty; asserts `Alice.Page != Bob.Page`, `Alice.ContextFixture.Context != Bob.ContextFixture.Context`, `Alice.…Browser == Bob.…Browser`
- [x] Mirrored snapshot under `TUnit.Templates.Tests/Snapshots/.../TUnit.Playwright/` keeps `PlaywriteTemplateTests.InstantiationTest` green
- [ ] CI pipeline runs the template project via `RunPlaywrightTestsModule` (after `InstallPlaywrightModule`) — exercises the smoke test end-to-end against real Chromium


thomhurst:
Thanks for the review.

**Issue 1 (`null!` field pattern) — pushing back.** Switching `Browser`/`Context`/`Page` to nullable would force every consumer to deal with `T?` at the API surface (`fixture.Browser!.NewContextAsync(...)` everywhere, or null-check noise in test code). The current pattern matches every fixture-style class already in this repo — `PlaywrightTest.Playwright`, `BrowserTest.Browser`, `ContextTest.Context`, `PageTest.Page`, `WorkerAwareTest._currentWorker`, and the same applies in `TUnit.Aspire.AspireFixture`. Diverging here would create two conventions side-by-side. The dispose null-check is genuinely defensive (covers init-throws-before-assignment), and `null!` accurately describes the steady-state contract from the consumer's perspective. Happy to revisit if you'd like to align all fixture-style types together in a separate PR.

**Issue 2 (hardcoded scope on inherited init properties)** — fair call, added doc comments to `BrowserFixture` and `ContextFixture` calling out that the scope is not overridable via subclassing and the workaround is to author a new fixture class. Commit `0a52084`.

**Minor (PageFixture default scope)** — added the doc note explaining `SharedType.None` is intentional for two-context-per-test scenarios. Same commit.

Also fixed the unrelated CI failure in the same commit: the template smoke test references new types added in this PR, but CI's restore pulls `TUnit.Playwright 1.*` from nuget.org which doesn't have them yet. Added a `Directory.Build.props` swap mirroring the existing TUnit.Aspire pattern, so solution builds resolve to the local project while end-user `dotnet new` instantiations keep the floating PackageReference.

thomhurst:
Round 2 follow-ups:

**New issue (`ContextFixture.GetContextOptions` defaults) — pushing back.** The defaults `Locale = "en-US"` and `ColorScheme = ColorScheme.Light` are not new — they're copied verbatim from `ContextTest.ContextOptions` (`TUnit.Playwright/ContextTest.cs:19-21`):

```csharp
public virtual BrowserNewContextOptions ContextOptions(TestContext testContext)
{
    return new() { Locale = "en-US", ColorScheme = ColorScheme.Light, };
}
```

`ContextTest` is the inheritance counterpart of `ContextFixture` (the layer that actually owns the `IBrowserContext`). The review compared against `BrowserTest` instead, which is the layer below — `BrowserTest` doesn't own contexts, so its `NewContext(options)` accepts whatever the caller passes. A user migrating from `ContextTest` → `[ClassDataSource<ContextFixture>]` keeps the exact same defaults. Skipping this one.

**Minor (launch-options forwarding comment)** — added one-liner in `PlaywrightServiceConnector` explaining `BrowserTypeLaunchOptions` are local-process only and don't carry to `ConnectAsync`. Commit `4b20368`.

**Minor (trailing newline on `Directory.Build.props`)** — fixed. Same commit.

Also pushed: `TUnit.PublicAPI` `.verified.txt` snapshots for the four new public types across all TFMs (this was the actual cause of the previous CI failures — turned out the build was fine; the snapshot test legitimately tripped on the API surface delta).

thomhurst:
Round 3 follow-ups in commit \`d134b06\`:

**`ContextFixture.GetContextOptions` doc** — implementing per the suggestion. Doc comment now states the defaults match `ContextTest.ContextOptions` for migration parity (same `Locale = "en-US"` / `ColorScheme = Light`) and how to restore browser defaults via `new BrowserNewContextOptions()`. My earlier reading still holds — those defaults exist verbatim in `ContextTest.cs:19-21`, so a migration from `ContextTest` → `ContextFixture` is a no-op behaviourally — but the doc makes that explicit for readers, which is the actually useful change.

**Separately — fixed a real CI regression caused by my own swap.** The earlier `Directory.Build.props` ProjectReference swap broke `dotnet run` on the template TestProject with *"test framework adapter has not been registered"*. Root cause: when `TUnit.Playwright` is consumed as a NuGet package, `build/TUnit.props` is auto-imported and sets `IsTestingPlatformApplication=true` + registers `TestingPlatformBuilderHook`. ProjectReference doesn't propagate build assets, so the project compiled but had no test-framework entry point at runtime. Fix: explicit `<Import Project="...\TUnit\TUnit.props" />` alongside the swap. Verified locally — test is now discovered and executes (fails at Playwright runtime because I don't have browsers installed locally; CI has them via `InstallPlaywrightModule`).

PR #5870, opened 2026-05-10
Generated mocks live in the same namespace as the mocked type

thomhurst:
## Summary
- Generated mock types (`{Type}Mock`, `{Type}Mockable`, `{Type}MockFactory`, `{Type}_MockStaticExtension`, `{Type}_MockMemberExtensions`, plus events extensions) now emit into the **same namespace as the mocked type** instead of `TUnit.Mocks.Generated.{Namespace}`.
- When a user-declared type in the target namespace would collide with one of the generator's emitted names, generation falls back automatically to `TUnit.Mocks.Generated[.{ns}]` — the existing scheme — so mocks never overwrite user code.
- Drops a long-standing inconsistency where `MockEventsBuilder` and `MockMembersBuilder` always emitted into `TUnit.Mocks.Generated`, regardless of the mocked type's namespace.
- Adds `MockNamespaceConflictDetector` which walks the consumer's compilation symbol table during the semantic transform phase. Cost is negligible — runs once per `MockTypeModel`, not per syntax node.
- Cross-assembly contract: types from referenced assemblies are checked against the consumer's namespace contents, not the foreign assembly's, so collisions in the user's project are correctly detected.

## Breaking change

Code that imports a generated sub-namespace such as `using TUnit.Mocks.Generated.MyApp.Services;` will no longer compile, because mocks now live in `MyApp.Services` directly. The fix is to drop the qualified `using` (the file's own namespace usually already imports it). The post-init `namespace TUnit.Mocks.Generated;` safety stub is preserved, so plain `using TUnit.Mocks.Generated;` continues to compile.

If a future framework version introduces a type whose name collides with one of the generator's suffixes, the source generator automatically detects the new collision on the next build and flips the affected mocks back to the `TUnit.Mocks.Generated.*` namespace — no manual migration required.

## Test plan
- [x] `TUnit.Mocks.SourceGenerator.Tests` — 174/174 pass across net8.0, net9.0, net10.0 (snapshots refreshed for the new namespace placement).
- [x] `TUnit.Mocks.Tests` — 968/968 pass on net10.0 (one stale `using TUnit.Mocks.Generated.TUnit.Mocks.Tests;` removed).
- [x] `TUnit.Mocks.Analyzers.Tests` — 30/30 pass on net10.0.
- [x] New `Global_Namespace_Mock_Does_Not_Emit_Empty_Namespace_Prefix` regression test asserts the generator does not emit invalid `global::.TypeName` syntax for globally-namespaced mocks.
- [x] New `Conflict_With_Existing_Type_Falls_Back_To_Generated_Namespace` snapshot exercises the fallback path.

## Notes
- No engine/runtime changes; source-gen-only behaviour. AOT and dual-mode rules satisfied.
- Mocks for external-assembly types (e.g. `System.IO.Stream`) emit into the original namespace (`StreamMock` lives in `System.IO`). Legal C# and uniform with user-authored types; conflict detection on the next build automatically reroutes if a collision later appears.


thomhurst:
Thanks for the review @claude. Addressed point-by-point:

**1. Divergent conflict detection** — fixed in `f76553729`. The two call sites passed `hasEvents` differently (`MockTypeDiscovery` used filtered events from `MemberDiscovery`, `GetGeneratedMockNamespace` used raw symbol members), so the edge case you flagged would genuinely produce mismatched FQNs. I dropped the `hasEvents` parameter from `HasConflict` entirely — the detector now always checks all seven emitted suffixes (`Mock`, `Mockable`, `MockFactory`, `_MockStaticExtension`, `_MockMemberExtensions`, `_MockEvents`, `_MockEventsExtensions`). Slightly more conservative (a colliding `T_MockEvents` flips fallback even when the target has no events), but eliminates the divergence and falling back is always safe. Updated `EventsExtensionsCollision_AlwaysFlagged` test.

**2. Framework namespace pollution** — keeping current behaviour. The risk is real but the source generator re-evaluates conflict detection on every build, so if .NET (or any reference) later adds a colliding type the next compile automatically flips the affected mocks to the fallback namespace. The "no obvious cause" failure mode the reviewer worries about would manifest as a normal compile error pointing at the colliding type — same UX as any other namespace conflict. The user explicitly chose this trade-off when designing the change; the auto-recovery property makes the choice safe rather than fragile.

**3. Breaking-change scope** — accurate but acceptable for the same reason as (2): existing `using TUnit.Mocks.Generated.MyApp.Services;` directives will fail to compile, the user removes them, and the next compile resolves cleanly via the new namespace placement (or implicit usings if the file already shares the namespace). I'll add a note to the PR description.

**Minor — namespace resolution scan**: agreed, not a hot path. Leaving as-is.

PR #5880, opened 2026-05-10
fix(core): fill optional params when invoking MethodDataSource via reflection

### LINKED ISSUE #5879: [Bug]: Optional CancellationToken parameter not respected as parameter while using MethodDataSource with abstract class
### Description

```
[InheritsTests]
public class TUnitMinimalReproduction : TUnitMinimalReproAbstract<string, string>
{
    protected override (string, string) Value => ("test", "test");
}

public abstract class TUnitMinimalReproAbstract<TInput, TExpected>
{
    [Test]
    [MethodDataSource(nameof(DataSource))]
    public async ValueTask Reproduce((TInput, TExpected) value)
    {
        var (input, expected) = value;
        await Assert.That(input).IsEquivalentTo(expected);
    }

    public async IAsyncEnumerable<(TInput, TExpected)> DataSource([EnumeratorCancellation] CancellationToken cancellationToken = default)
    {
        yield return Value;
    }
    
    protected abstract (TInput, TExpected) Value { get;  }
}
```

### Expected Behavior

The test should behave as if the class was fully implemented using no base class

### Actual Behavior

Fails with `TargetParameterCountException`

### Steps to Reproduce

1. Create an abstract class with a MethodDataSource and method that accepts a CancellationToken as an optional parameter (example in description)
2. Inherit from that class and add `[InheritsTests]`
3. Run the tests

### TUnit Version

1.43.11

### .NET Version

.NET 10

### Operating System

Linux

### IDE / Test Runner

```shell
System.Reflection.TargetParameterCountException: Parameter count mismatch.
   at System.Reflection.MethodBaseInvoker.ThrowTargetParameterCountException()
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at TUnit.Core.MethodDataSourceAttribute.GetDataRowsAsync(DataGeneratorMetadata dataGeneratorMetadata)+MoveNext()
   at TUnit.Core.MethodDataSourceAttribute.GetDataRowsAsync(DataGeneratorMetadata dataGeneratorMetadata)+System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult()
   at TUnit.Engine.Building.TestBuilder.GetInitializedDataRowsAsync(IDataSourceAttribute dataSource, DataGeneratorMetadata dataGeneratorMetadata, CancellationToken cancellationToken)+MoveNext()
   at TUnit.Engine.Building.TestBuilder.GetInitializedDataRowsAsync(IDataSourceAttribute dataSource, DataGeneratorMetadata dataGeneratorMetadata, CancellationToken cancellationToken)+MoveNext()
   at TUnit.Engine.Building.TestBuilder.GetInitializedDataRowsAsync(IDataSourceAttribute dataSource, DataGeneratorMetadata dataGeneratorMetadata, CancellationToken cancellationToken)+System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult()
   at TUnit.Engine.Building.TestBuilder.BuildTestsFromMetadataAsync(TestMetadata metadata, TestBuildingContext buildingContext, CancellationToken cancellationToken)
   at TUnit.Engine.Building.TestBuilder.BuildTestsFromMetadataAsync(TestMetadata metadata, TestBuildingContext buildingContext, CancellationToken cancellationToken)
   at TUnit.Engine.Building.TestBuilder.BuildTestsFromMetadataAsync(TestMetadata metadata, TestBuildingContext buildingContext, CancellationToken cancellationToken)
```

### Additional Context

Removing the `cancellationToken` parameter causes the tests to run correctly. Implementing the same behavior without the abstract class causes the tests to run correctly.

Examples of valid workarounds:

```
[InheritsTests]
public class TUnitMinimalReproduction : TUnitMinimalReproAbstract<string, string>
{
    protected override (string, string) Value => ("test", "test");
}

public abstract class TUnitMinimalReproAbstract<TInput, TExpected>
{
    [Test]
    [MethodDataSource(nameof(DataSource))]
    public async ValueTask Reproduce((TInput, TExpected) value)
    {
        var (input, expected) = value;
        await Assert.That(input).IsEquivalentTo(expected);
    }

    public async IAsyncEnumerable<(TInput, TExpected)> DataSource()
    {
        yield return Value;
    }
    
    protected abstract (TInput, TExpected) Value { get;  }
}

public class TUnitMinimalRepro
{
    [Test]
    [MethodDataSource(nameof(DataSource))]
    public async ValueTask Reproduce((string, string) value)
    {
        var (input, expected) = value;
        await Assert.That(input).IsEquivalentTo(expected);
    }

    public async IAsyncEnumerable<(string, string)> DataSource([EnumeratorCancellation] CancellationToken cancellationToken = default)
    {
        yield return Value;
    }
    
    protected (string, string) Value => ("test", "test");
}
```

### IDE-Specific Issue?

thomhurst:
## Summary

- Closes #5879. `[InheritsTests]` derived class on a generic abstract base + a `[MethodDataSource]` method with an optional parameter (e.g. `[EnumeratorCancellation] CancellationToken ct = default`) threw `TargetParameterCountException` at discovery.
- Root cause: `MethodDataSourceAttribute.GetDataRowsAsync` called `methodInfo.Invoke(instance, Arguments)` with `Arguments == []`; reflection's `Invoke` does not auto-fill defaults like a C# call site. The source-gen `Factory` path emits a direct C# call (compiler fills defaults), but the reflection fallback used for the generic abstract base case did not — so the optional CT parameter triggered the count mismatch.
- Fix: added `BuildInvokeArgs(MethodInfo, object?[])` that pads the supplied args to the method's parameter count, filling missing slots with `ParameterInfo.DefaultValue`, then `CancellationToken.None` for `CancellationToken`-typed params without an explicit default, then falling back to the original array so genuine required-param mismatches still surface the original `TargetParameterCountException`.

## Behaviour parity with source-gen

The source generator emits `DataSource()` (`TestMetadataGenerator.cs:1628`) — the C# compiler then fills `default(CancellationToken)`. The reflection path now matches that exactly (`CancellationToken.None`). Plumbing a build-time CT through `IDataSourceAttribute.GetDataRowsAsync` would be a public-API change and is out of scope for this bug fix.

## Test plan

- [x] Regression test: `TUnit.TestProject/Bugs/5879/Repro.cs` — exact repro from the issue, marked `[EngineTest(ExpectedResult.Pass)]`. Throws `TargetParameterCountException` before the fix; passes after.
- [x] `dotnet test TUnit.TestProject --treenode-filter "/*/*/Issue5879Tests/*"` → 1/1 pass.
- [x] `dotnet test TUnit.UnitTests` → 191/191 pass (no regressions in adjacent paths).

### DIFF
--- a/TUnit.Core/Attributes/TestData/MethodDataSourceAttribute.cs
+++ b/TUnit.Core/Attributes/TestData/MethodDataSourceAttribute.cs
@@ -324,9 +324,7 @@ public class MethodDataSourceAttribute : Attribute, IDataSourceAttribute
         }
     }
 
-    // Reflection's MethodBase.Invoke does not auto-fill optional parameters the way a C# call site does.
-    // The source generator emits a direct C# call (compiler fills defaults) but the reflection fallback
-    // — used for [InheritsTests] generic abstract bases (issue #5879) — must build the args array itself.
+    // MethodInfo.Invoke does not auto-fill optional parameters the way a C# call site does.
     private static object?[] BuildInvokeArgs(MethodInfo methodInfo, object?[] suppliedArguments)
     {
         var parameters = methodInfo.GetParameters();
--- a/TUnit.Core/Attributes/TestData/MethodDataSourceAttribute.cs
+++ b/TUnit.Core/Attributes/TestData/MethodDataSourceAttribute.cs
@@ -328,8 +328,9 @@ public class MethodDataSourceAttribute : Attribute, IDataSourceAttribute
     private static object?[] BuildInvokeArgs(MethodInfo methodInfo, object?[] suppliedArguments)
     {
         var parameters = methodInfo.GetParameters();
-        if (parameters.Length == suppliedArguments.Length)
+        if (parameters.Length <= suppliedArguments.Length)
         {
+            // Exact match passes through; surplus supplied args are left to Invoke to surface as a mismatch.
             return suppliedArguments;
         }
 
--- a/TUnit.Core/Attributes/TestData/MethodDataSourceAttribute.cs
+++ b/TUnit.Core/Attributes/TestData/MethodDataSourceAttribute.cs
@@ -196,7 +196,7 @@ public class MethodDataSourceAttribute : Attribute, IDataSourceAttribute
                 instance = await GetOrCreateInstanceAsync(dataGeneratorMetadata, targetType);
             }
 
-            methodResult = methodInfo.Invoke(instance, Arguments);
+            methodResult = methodInfo.Invoke(instance, BuildInvokeArgs(methodInfo, Arguments));
         }
         else
         {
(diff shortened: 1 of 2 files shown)


thomhurst:
Thanks for the review.

**Truncation regression** — applied in af2d29d. Changed the early-return guard from `==` to `<=`, so any surplus supplied args now fall through to `Invoke` and surface the original `TargetParameterCountException` as before.

**`Type.Missing`** — pushing back on this one. The runtime layer (`MethodBase.CheckArguments` / `MethodInvoker`) treats `Type.Missing` as "use the parameter's default" only when `ParameterInfo.HasDefaultValue` is `true` — it doesn't gain anything beyond what the `HasDefaultValue` branch already does. The explicit `CancellationToken` branch is specifically the safety net for `[Optional]`-without-`[DefaultParameterValue]`, where `HasDefaultValue` is `false`; `Type.Missing` would throw there too. So the manual approach is actually slightly more general than the suggested rewrite, not less. I'd rather keep the targeted CT fallback (which matches what the source-generator emits — `default(CancellationToken)`) and revisit if other "no-default-attribute" optional types come up.

thomhurst:
Applied the `Type.Missing` + `Array.Copy` simplification in 5833192. Regression test still green. Thanks!

PR #5881, opened 2026-05-10
Mocks: structural fix for Mock<T> / mocked-member name collisions

thomhurst:
## Summary
- Move framework operations (`Reset`, `VerifyAll`, `Invocations`, …) off `Mock<T>` as instance members onto `IMockControl<T>` explicit interface implementations plus `Mock.X(mock)` static helpers. `Mock<T>` instance surface shrinks to `Object` + constructor + implicit conversion to `T`.
- Collision set is now structurally closed: `{ Object, Equals, GetHashCode, ToString }`. Generator's existing `Object_` / `*Of` renames keep working, with new iterative-suffix escalation when the user interface declares the renamed form (`Object` + `Object_` → generated extension lands on `Object__`).
- On net9.0+ the generator emits low-priority extension polyfills inside the per-type `*_MockMemberExtensions` class so `mock.Reset()`, `mock.VerifyAll()`, `mock.Invocations`, `mock.SetState("x")`, `mock.DefaultValueProvider = …` remain ergonomic. `[OverloadResolutionPriority(-1)]` only ranks within one containing type, so the polyfills must live alongside the generated extensions; they skip emission when the mocked interface declares a member of the same name (no CS0111).

Background: GitHub Discussion [#4981](https://github.com/thomhurst/TUnit/discussions/4981) raised the collision concern. Prior handling covered ~4 names by ad-hoc rename table and would have silently shadowed generator output for ~10 other public `Mock<T>` members. With this change, future framework additions go on `IMockControl<T>` / static `Mock.*` and cannot regress collision safety.

## Breaking changes (TUnit.Mocks is in beta)

- `Mock<T>` no longer exposes `Reset()`, `VerifyAll()`, `VerifyNoOtherCalls()`, `Invocations`, `Behavior`, `DefaultValueProvider`, `SetupAllProperties()`, `GetDiagnostics()`, `SetState()`, `InState()` as instance members.
- **net9.0+ consumers:** code keeps compiling — generator emits ergonomic polyfills as low-priority extensions, so `mock.Reset()` etc. continue to work.
- **net8.0 / netstandard2.0 consumers:** call sites must move to the static helpers (`Mock.Reset(mock)`, `Mock.VerifyAll(mock)`, etc.). `[OverloadResolutionPriority]` isn't available on those targets so the polyfill path can't apply.
- `Mock.DefaultValueProvider` getter/setter overload split into `Mock.GetDefaultValueProvider(mock)` and `Mock.SetDefaultValueProvider(mock, provider)` per .NET naming conventions.

## Test plan
- [x] `NameCollisionTests` (new) exercises every former framework name plus `Object_` / `EqualsOf` overflow paths.
- [x] On net9.0+, polyfill tests verify `mock.Reset()` / `mock.VerifyAll()` / `mock.Invocations` / property setter `mock.DefaultValueProvider = …` route through and that polyfills are skipped when colliding.
- [x] `TUnit.Mocks.Tests` full suite: 977 pass (net10.0), 967 pass (net8.0).
- [x] Full solution `dotnet build TUnit.slnx` succeeds.
- [ ] CI: source-generator snapshot tests (`TUnit.Mocks.SourceGenerator.Tests`) — local runner blocked by an unrelated `Verify.TUnit` / `TUnit.Core` version mismatch; CI will validate.

### DIFF
--- a/TUnit.Mocks.SourceGenerator.Tests/Snapshots/Class_Implementing_Static_Abstract_Interface.verified.txt
+++ b/TUnit.Mocks.SourceGenerator.Tests/Snapshots/Class_Implementing_Static_Abstract_Interface.verified.txt
@@ -66,6 +66,52 @@ public static class StaticAbstractImpl_MockMemberExtensions
         public global::TUnit.Mocks.PropertyMockCall<int> InstanceValue
             => new(global::TUnit.Mocks.MockRegistry.GetEngine(mock), 0, 0, "InstanceValue", true, false);
     }
+
+    #if NET9_0_OR_GREATER
+    [global::System.Runtime.CompilerServices.OverloadResolutionPriority(-1)]
+    public static void Reset(this global::TUnit.Mocks.Mock<global::StaticAbstractImpl> mock)
+        => global::TUnit.Mocks.Mock.Reset(mock);
+
+    [global::System.Runtime.CompilerServices.OverloadResolutionPriority(-1)]
+    public static void VerifyAll(this global::TUnit.Mocks.Mock<global::StaticAbstractImpl> mock)
+        => global::TUnit.Mocks.Mock.VerifyAll(mock);
+
+    [global::System.Runtime.CompilerServices.OverloadResolutionPriority(-1)]
+    public static void VerifyNoOtherCalls(this global::TUnit.Mocks.Mock<global::StaticAbstractImpl> mock)
+        => global::TUnit.Mocks.Mock.VerifyNoOtherCalls(mock);
+
+    [global::System.Runtime.CompilerServices.OverloadResolutionPriority(-1)]
+    public static void SetupAllProperties(this global::TUnit.Mocks.Mock<global::StaticAbstractImpl> mock)
+        => global::TUnit.Mocks.Mock.SetupAllProperties(mock);
+
+    [global::System.Runtime.CompilerServices.OverloadResolutionPriority(-1)]
+    public static global::TUnit.Mocks.Diagnostics.MockDiagnostics GetDiagnostics(this global::TUnit.Mocks.Mock<global::StaticAbstractImpl> mock)
+        => global::TUnit.Mocks.Mock.GetDiagnostics(mock);
+
+    [global::System.Runtime.CompilerServices.OverloadResolutionPriority(-1)]
+    public static void SetState(this global::TUnit.Mocks.Mock<global::StaticAbstractImpl> mock, string? stateName)
+        => global::TUnit.Mocks.Mock.SetState(mock, stateName);
+
+    [global::System.Runtime.CompilerServices.OverloadResolutionPriority(-1)]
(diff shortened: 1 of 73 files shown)


thomhurst:
Thanks for the review. Status per item:

**#2 (O(N²) HashSet allocation)** — already addressed in commit `7e47b9ff3` (pushed before this review landed). `GetUserMemberNames` now memoizes per `MockTypeModel` via `ConditionalWeakTable`, so a single HashSet is built per type per generator pass. Re-confirmed.

**#3 (DefaultValueProvider get/set overloads)** — fixed in `0f5b969dc`. Renamed the static helpers to `Mock.GetDefaultValueProvider(mock)` / `Mock.SetDefaultValueProvider(mock, provider)`. The net9.0+ polyfill extension property keeps surfacing as `mock.DefaultValueProvider` (get/set) for the natural API at use sites.

**#4 (IMock vs IMockControl<T> split)** — intentional, but now documented on `IMockControl<T>`. The split is driven by `MockRepository`, which iterates mocks as a heterogeneous `IMock[]` for batch operations and so needs `Reset`/`VerifyAll`/`VerifyNoOtherCalls` in non-generic form. Members that depend on the type parameter (`Invocations`, `GetDiagnostics`, typed `Object`, state-machine helpers) only make sense generic and live on `IMockControl<T>`. Added a `<remarks>` block explaining this.

**#5 (blank line before empty `#if/#endif`)** — fixed in `0f5b969dc`. `GenerateMockControlPolyfills` now short-circuits when every polyfill name collides with a user member, so the `#if NET9_0_OR_GREATER` block isn't emitted at all in that (rare) case.

**#1 (net8.0 ergonomic regression)** — declined. The instance-style API on net8.0 was the source of the silent shadowing the PR exists to fix; restoring it on that TFM requires either polyfilling `OverloadResolutionPriorityAttribute` ourselves (consumer LangVersion ≥ 13 not guaranteed) or accepting collision risk to come back. TUnit.Mocks is beta and a single clean break across all TFMs is preferred over a partial fix. net8.0 users have `Mock.X(mock)` available; net9.0+ users get the polyfill ergonomics back.

Verified 977/977 pass on net10.0 and 967/967 on net8.0.

thomhurst:
Fixed in 02a0b3b4e — moved the separator blank-line emission inside `GenerateMockControlPolyfills` after the early-return guard, so nothing is written when every polyfill name collides with a user member. `firstMember` is now passed in as `isFirstMember` rather than acted on at the call site.

Re item 1: added to PR description as a release-notes follow-up.

thomhurst:
Declining the C# 14 / net10 gate split. TUnit.Mocks already requires C# 14 across the library — the existing generator emits extension blocks unconditionally for mocked-interface properties, and consumers must already opt into a later LangVersion. Splitting the polyfill gate would add real complexity for an effectively non-existent compatibility envelope. Keeping the single `#if NET9_0_OR_GREATER` block.

PR #5894, opened 2026-05-11
feat(mocks): add Mock.HttpClientFactory() helper

thomhurst:
## Summary

- Adds `MockHttpClientFactory : IHttpClientFactory` backed by `MockHttpHandler`(s); each `CreateClient` returns a fresh non-disposing `HttpClient` so captured requests survive across `using` blocks.
- Adds `Mock.HttpClientFactory()` / `Mock.HttpClientFactory(handler)` extensions.
- Supports named clients via `WithHandler(name, handler)` / `HandlerFor(name)` for `services.AddHttpClient("name")` DI scenarios; unregistered names fall back to the default handler.

Resolves the UX pain from discussion #5885: today users have to write `Returns(() => new HttpClient(_handler, disposeHandler: false))` and tolerate `ObjectDisposedException` when a single mock client is reused across `using` blocks.

### Before

```csharp
var handler = new MockHttpHandler();
var factory = IHttpClientFactory.Mock();
factory.CreateClient(Any()).Returns(() => new HttpClient(handler, disposeHandler: false));
```

### After

```csharp
var factory = Mock.HttpClientFactory();
factory.Handler.OnGet("/api/users").RespondWithJson("[]");
// inject factory; SUT can `using var c = factory.CreateClient()` freely
factory.Handler.Verify(r => r.Path("/api/users"), Times.Once);
```

## Test plan

- [x] `MockHttpClientFactoryTests` (9 tests, all pass on net10.0)
  - default handler responses
  - empty-name/default-client fallback
  - survives multiple `using` block disposals
  - fresh `HttpClient` instance per call
  - named-client routing
  - fallback to default when name unregistered
  - case-insensitive name matching
  - factory disposal prevents further use
  - `Verify` works across multiple client lifetimes
- [x] Full `TUnit.Mocks.Http.Tests` suite passes on net10.0 in source-generated and reflection modes
- [x] Docs section added to `docs/docs/writing-tests/mocking/http.md`


thomhurst:
Addressed in `6c0d80a`:

- resolved the `Directory.Packages.props` merge conflict against current `main`, keeping the `10.0.8` Microsoft package versions and retaining `Microsoft.Extensions.Http`
- made `MockHttpClientFactory` implement `IDisposable`
- switched named handler/base-address storage to `ConcurrentDictionary`
- added coverage for `CreateClient(string.Empty)` and post-dispose use

Validation:

- `TUnit.Mocks.Http.Tests.exe --no-progress --minimum-expected-tests 1` passed 54/54 on `net10.0`
- `TUnit.Mocks.Http.Tests.exe --no-progress --reflection --minimum-expected-tests 1` passed 54/54 on `net10.0`

PR #5919, opened 2026-05-13
fix(sourcegen): emit `default` literal for value-type assertion parameters

### LINKED ISSUE #5917: [Bug]: assertion generators emit null as default for non-nullable value-type parameters (CS1750)
### Description

Both source generators in `TUnit.Assertions.SourceGenerator` (the `[GenerateAssertion]` path and the `[AssertionExtension]` path) emit `= null` for non-nullable value-type parameters that were declared with the `default` literal. The literal `null` cannot convert to a value type, so the generated extension does not compile.

### Expected Behavior

A `[GenerateAssertion]` method or `[AssertionExtension]` class with a non-nullable value-type parameter declared `= default` (for example `CancellationToken ct = default`) should produce a generated extension that compiles cleanly.

### Actual Behavior

The generated extension's parameter is emitted as `Type name = null`, producing `CS1750: A value of type 'null' cannot be used as a default parameter because there are no standard conversions to type 'Type'`.

### Steps to Reproduce

**Variant A. `[GenerateAssertion]` path:**

```csharp
{
    [GenerateAssertion]
    public static bool RespectsToken(this int value, CancellationToken token = default)
        => !token.IsCancellationRequested && value > 0;
}
```

**Variant B. `[AssertionExtension]` path:**

```csharp
[AssertionExtension("RespectsToken")]
public class StringRespectsTokenAssertion : Assertion<string>
{
    public StringRespectsTokenAssertion(AssertionContext<string> context, CancellationToken token = default)
        : base(context) { }

    protected override Task<AssertionResult> CheckAsync(EvaluationMetadata<string> metadata)
        => Task.FromResult(AssertionResult.Passed);

    protected override string GetExpectation() => "to respect the supplied token";
}
```

Both variants fail at build time with CS1750.

### TUnit Version

1.44.0

### .NET Version

.NET 10.0

### Operating System

Windows

### IDE / Test Runner

```shell
error CS1750: A value of type 'null' cannot be used as a default parameter because there are no standard conversions to type 'System.Threading.CancellationToken'
```

### Additional Context

The bug exists in both source generators in `TUnit.Assertions.SourceGenerator`. Both code paths return the same string for a `null` Roslyn-reported default, so any non-nullable value-type parameter declared with the bare `default` literal triggers it. `Nullable<T>` parameters are unaffected (the literal `null` is a valid default for those).

### IDE-Specific Issue?

JohnVerheij:
## Description

Both source generators in `TUnit.Assertions.SourceGenerator` carried a near-duplicate private `FormatDefaultValue` helper, and both returned `"null"` when the parameter's Roslyn-reported default-value constant was `null`. For a non-nullable value-type parameter declared with the `default` literal (e.g. `CancellationToken ct = default`), Roslyn reports the default as `null`, so the generator emitted `= null`. That literal is invalid as the default for a value type and produces CS1750.

Fix: when the constant is `null`, branch on the parameter type. Non-nullable value types emit the bare `default` literal (the compiler infers the target type from the parameter, matching the user's original declaration). Reference types and `Nullable<T>` continue to emit `null`.

Unified the two private helpers into a single shared `DefaultValueFormatter.FormatDefaultValue` (new file `Generators/DefaultValueFormatter.cs`, sibling to the existing `CovarianceHelper`). The helper takes a `bool useFullyQualifiedEnumName` parameter so each generator preserves its intentional enum-emit style: `MethodAssertionGenerator` passes `true` (emits `Namespace.Enum.Member` to match its fully-qualified-everywhere style), and `AssertionExtensionGenerator` passes `false` (emits `Enum.Member` because it relies on `using` directives at the top of its generated files). Both generators now share a single source of truth for default-value emit while keeping the call-site-specific behaviour explicit.

Test coverage:
- New `ValueTypeDefaultParameter` test on `MethodAssertionGeneratorTests` (`[GenerateAssertion]` path).
- New `ValueTypeDefaultParameter` test on `AssertionExtensionGeneratorTests` (`[AssertionExtension]` path).
- Both tests pin the surface (`token = default,`) and run a structural compile-clean gate.
- Extracted a shared `CompileChecker.AssertNoErrors(generatedFiles)` test helper (sibling to `ReferencesHelper`) so the compile-clean assertion isn't duplicated across the two new tests.

## Related Issue

Fixes #5917

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

### Required

### Testing

## Additional Notes

- **Dual-Mode rule is not applicable.** Per `.claude/docs/mandatory-rules.md` Rule 1, dual-mode does not apply to assertion library changes. This PR fixes an emit defect in `TUnit.Assertions.SourceGenerator` only.
- **Performance / AOT rules are not applicable.** Trivial branch in the formatter; no allocation, no reflection.
- **Helper unification preserves existing behaviour.** All existing snapshots compare identically before and after the unification. The `useFullyQualifiedEnumName` parameter encodes the previously-duplicated semantic difference between the two generators, so neither generator's output for any existing case changes.
- **net472 CS0122 conditional in the new `CompileChecker` test helper:** on the net472 leg of CI, the test project's `Polyfill` assembly marks `CallerArgumentExpressionAttribute` as `internal`, producing a CS0122 ('inaccessible due to its protection level') false positive when Roslyn compiles the generator's output through the test reference set. Conditional `#if NETFRAMEWORK` filter pre-empts this in the helper. On modern TFMs the BCL attribute is public, so CS0122 (if it ever appears) remains a real failure signal. All four TFMs (net472 + net8.0 + net9.0 + net10.0) verified green locally before push.
- Multi-target snapshots updated/added for net8.0, net9.0, net10.0, net472.


thomhurst:
Thanks!

PR #5920, opened 2026-05-13
fix(sourcegen): drop spurious ')' in MethodAssertionGenerator Task<bool> emit

### LINKED ISSUE #5916: [Bug]: [GenerateAssertion] on Task<bool> method emits extra ')' producing invalid C#
### Description

`MethodAssertionGenerator` emits an extra `)` in the failure-result arm when the source method returns `Task<bool>`. The generated extension method does not compile.

### Expected Behavior

A `[GenerateAssertion]`-decorated method returning `Task<bool>` should produce a generated extension that compiles cleanly.

### Actual Behavior

The generated source has a stray `)` before the terminating `;` in the failure-return arm of the `Task<bool>` branch. Building fails with CS1002 / CS1513.

### Steps to Reproduce

1. Create a partial static class with a `[GenerateAssertion]` method that returns `Task<bool>`:

```csharp
{
    [GenerateAssertion]
    public static async Task<bool> IsPositiveAsync(this int value)
        => await Task.FromResult(value > 0);
}
```

2. Build the project.
3. The build fails on the source-generator output.

### TUnit Version

1.44.0

### .NET Version

.NET 10.0

### Operating System

Windows

### IDE / Test Runner

```shell
error CS1002: ; expected
error CS1513: } expected
```

### Additional Context

Reproduces on a fresh project referencing TUnit + TUnit.Assertions. The defect is in the source generator's emit for the `Task<bool>` return-kind branch, so it manifests on every TFM and every IDE. The defect is not IDE-specific.

### IDE-Specific Issue?

JohnVerheij:
## Description

The `ReturnTypeKind.TaskBool` branch in `MethodAssertionGenerator` emitted an extra `)` before the terminating `;` in the failure-result arm, producing invalid C# for any `[GenerateAssertion]` on a `Task<bool>`-returning method.

Single character removed in the emit. Added an `AsyncBoolMethod` content-anchor that pins the absence of `Failed($"found {value}"));` (the buggy two-paren tail) and the presence of `Failed($"found {value}");` (the correct one-paren tail). Added a companion `AsyncBoolAssertion_GeneratesValidCSharp` test in `XmlDocsRegressionTests` that compiles the generator output through Roslyn and asserts zero error-severity diagnostics, so the whole class of emit-syntax bugs in async-return arms is pinned regardless of which specific diagnostic id surfaces.

## Related Issue

Fixes #5916

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

### Required

### Testing

## Additional Notes

- **Dual-Mode rule is not applicable.** Per `.claude/docs/mandatory-rules.md` Rule 1, the dual-mode requirement does not apply to assertion library changes. This PR fixes an emit defect in `TUnit.Assertions.SourceGenerator` only.
- **Performance / AOT rules are not applicable.** Single-character change in a generator string emit; no allocation or reflection changes.
- **net472 CS0122 conditional in the new compile-clean test:** on the net472 leg of CI, the test project's `Polyfill` assembly marks `CallerArgumentExpressionAttribute` as `internal`, producing a CS0122 ('inaccessible due to its protection level') false positive when Roslyn compiles the generator's output through the test reference set. Conditional `#if NETFRAMEWORK` filter pre-empts this. On modern TFMs the BCL attribute is public, so CS0122 (if it ever appears) remains a real failure signal. All four TFMs (net472 + net8.0 + net9.0 + net10.0) verified green locally before push.
- **Inline compile-clean check is intentional given parallel filing.** A sibling PR (Bug 1, titled "fix(sourcegen): emit `default` literal for value-type assertion parameters") introduces a shared `CompileChecker.AssertNoErrors` test helper. Both PRs were authored in parallel, so the helper isn't on `main` yet, and the inline form here matches the helper's body. Can rebase this branch and replace the inline copy with the shared helper once Bug 1 merges, whichever order you prefer to land them in.
- Multi-target snapshots updated for net8.0, net9.0, net10.0, net472.


thomhurst:
Thanks!

PR #5921, opened 2026-05-13
fix(sourcegen): merge generic parameter lists in [AssertionExtension] emit

### LINKED ISSUE #5918: [Bug]: [AssertionExtension] on generic class with concrete receiver emits two angle-bracket lists
### Description

When an `[AssertionExtension]`-decorated class introduces its own generic type parameter **and** derives from `Assertion<TConcrete>` for a non-sealed concrete receiver, `AssertionExtensionGenerator` emits the receiver's covariant type parameter and the class's own generic parameter as two adjacent angle-bracket lists in the generated extension method signature.

### Expected Behavior

The generator should emit a single merged generic-parameter list (for example `<TActual, T>`) so the generated extension compiles.

### Actual Behavior

The generator emits something shaped like `<TActual><T>`, with two separate generic-parameter blocks back-to-back. This is invalid C# syntax; the file fails to parse.

### Steps to Reproduce

1. Declare an `[AssertionExtension]` class that introduces a generic parameter and derives from `Assertion<NonSealedConcreteType>`:

```csharp
[AssertionExtension("Matches")]
public class MatchesAssertion<T> : Assertion<System.Exception>
{
    private readonly System.Func<T, bool> _predicate;

    public MatchesAssertion(AssertionContext<System.Exception> context, System.Func<T, bool> predicate)
        : base(context) { _predicate = predicate; }

    protected override System.Threading.Tasks.Task<AssertionResult> CheckAsync(
        EvaluationMetadata<System.Exception> metadata)
        => System.Threading.Tasks.Task.FromResult(AssertionResult.Passed);

    protected override string GetExpectation() => "to match";
}
```

2. Build the project. The generated extension method has two adjacent `<...>` blocks in its signature and the build fails.

### TUnit Version

1.44.0

### .NET Version

.NET 10.0

### Operating System

Windows

### IDE / Test Runner

```shell
error CS1003: Syntax error, ',' expected
```

### Additional Context

The bug surfaces when both conditions hold:
1. The assertion class declares its own generic type parameter (`class MyAssertion<T> : ...`).
2. The base assertion type is `Assertion<TConcrete>` where `TConcrete` is a non-sealed class (i.e. a covariance candidate per the generator's existing covariance logic).

A class that derives from `Assertion<TConcrete>` for a sealed type, or a class without its own generic parameter, is not affected.

### IDE-Specific Issue?

JohnVerheij:
## Description

`AssertionExtensionGenerator` emitted the covariant receiver-type parameter (when applied) and the assertion class's own generic parameter as two adjacent angle-bracket blocks in the generated extension method signature, for example `Matches<TActual><T>(this IAssertionSource<TActual> source, ...)`. That's not valid C#; the generated file fails to parse for any `[AssertionExtension]`-decorated class that derives from `Assertion<ConcreteNonSealedType>` while also declaring its own generic type parameter.

Fix: collect both the covariant receiver parameter and the class's own generic parameters into a single list and emit one `<...>` block before the method signature opens.

Test coverage:
- New `ConcreteReceiverWithExtraGenericAssertion` test fixture in `TestData/`. Uses `System.Exception` as the receiver. Exception is a BCL non-sealed class that the generator's existing covariance-candidate check recognises, so the fixture trips the affected branch without adding a stand-in type to the test surface.
- New `ConcreteReceiverWithExtraGeneric` test on `AssertionExtensionGeneratorTests`. Pins the absence of adjacent generic-parameter blocks (`DoesNotContain "><"`) and runs an inline Roslyn compile-clean gate so the whole class of emit-syntax bugs is pinned regardless of which specific diagnostic id the parser produces.

## Related Issue

Fixes #5918

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

### Required

### Testing

## Additional Notes

- **Dual-Mode rule is not applicable.** Per `.claude/docs/mandatory-rules.md` Rule 1, dual-mode does not apply to assertion library changes. This PR fixes an emit defect in `TUnit.Assertions.SourceGenerator` only.
- **Performance / AOT rules are not applicable.** Small string-builder change in the emit; no allocation or reflection concern.
- **Call-site ergonomics for this shape** (callers having to repeat `TActual` at every site) is a separate design call. It's tracked as a follow-up issue rather than expanded here.
- **net472 CS0122 conditional in the inline compile-clean check:** on the net472 leg of CI, the test project's `Polyfill` assembly marks `CallerArgumentExpressionAttribute` as `internal`, producing a CS0122 ('inaccessible due to its protection level') false positive when Roslyn compiles the generator's output through the test reference set. Conditional `#if NETFRAMEWORK` filter pre-empts this. On modern TFMs the BCL attribute is public, so CS0122 (if it ever appears) remains a real failure signal. All four TFMs (net472 + net8.0 + net9.0 + net10.0) verified green locally before push.
- **Inline compile-clean check is intentional given parallel filing.** A sibling PR (Bug 1, titled "fix(sourcegen): emit `default` literal for value-type assertion parameters") introduces a shared `CompileChecker.AssertNoErrors` test helper. Both PRs were authored in parallel, so the helper isn't on `main` yet, and the inline form here matches the helper's body. Can rebase this branch and replace the inline copy with the shared helper once Bug 1 merges, whichever order you prefer to land them in.
- Multi-target snapshots added for net8.0, net9.0, net10.0, net472.


thomhurst:
Thanks!

Issue #5922, opened 2026-05-13
[Feature]: ergonomic call-site for [AssertionExtension] classes that introduce their own generic parameter


JohnVerheij:
### Problem Statement

Once the dual-generic emit bug (#5918) is fixed, an `[AssertionExtension]` class with its own generic parameter and a concrete (covariance-candidate) receiver compiles cleanly. The remaining friction is the call-site: the generated extension's signature is `Matches<TActual, T>(this IAssertionSource<TActual> source, Func<T, bool> predicate, ...) where TActual : Exception`, which means callers have to specify `TActual` explicitly when they only care about `T`:

    Assert.That(ex).Matches<Exception, MyPayload>(p => p.IsValid);

Type inference can pick up `T` from the predicate argument, but it can't pick up `TActual` from the receiver expression alone, so the redundant `Exception` has to be written out at every call site.

### Proposed Solution

This is a design call rather than a prescription, but the desired ergonomic end-state is: when the receiver type is concrete (the covariance-candidate branch the generator already detects), the call-site only needs to specify the class's own generic parameters. The receiver type isn't repeated.

There are a few plausible shapes for this (omit the covariant `TActual` and pin the receiver to `IAssertionSource<TConcrete>` directly; emit an additional inference-friendly overload alongside the covariant one; encode it through a constraint shape the inferencer can resolve from the receiver). Would prefer to wait on your steer about which shape fits the rest of the generator's design before opening a PR for this.

### Alternatives Considered

- **Live with the explicit type argument.** Works, but every call site repeats the receiver type. Verbose for the most common case.
- **Drop covariance support for this shape.** Restores inference but loses the ability to assert against a subclass of the declared receiver.
- **Workaround at the call site:** wrap the assertion in a helper that fixes `TActual`. Adds boilerplate that the generator could in principle absorb.

### Feature Category

Assertions

### How important is this feature to you?

Nice to have - would improve my experience

### Additional Context

This is downstream of the dual-generic-emit fix (#5918). The emit fix is structural. Without it the generated source doesn't compile at all. This issue is purely about call-site ergonomics once the compile-time bug is gone.

### Contribution

- [x] I'm willing to submit a pull request for this feature

JohnVerheij:
[...]

PR #5935, opened 2026-05-15
fix(sourcegen): drop covariant TActual when [GenerateAssertion] method has its own type parameters

### LINKED ISSUE #5934: [Bug]: [GenerateAssertion] generic method on a non-sealed receiver fails CS1929 at call sites with explicit type arguments
### Description

When `[GenerateAssertion]` is applied to a generic source method whose receiver is a concrete non-sealed reference type, the generator prepends a covariant receiver-type parameter (`TActual`) to the produced extension. The resulting two-type-parameter signature cannot bind at a call site that supplies explicit type arguments for the method's own type parameters, because C# does not permit partial type-argument specification. The call fails with CS1929.

Four cases isolate the trigger:

1. generic method on a non-sealed receiver, call with `.Method<T>(...)`: CS1929
2. generic method on a sealed receiver, same call: compiles (covariance disabled)
3. non-generic method on a non-sealed receiver: compiles (no type arg to specify)
4. generic method on a non-sealed receiver, full inference: compiles (no explicit arg)

### Expected Behavior

`Assert.That(container).HasItem<int>(42)` compiles and runs the generated assertion.

### Actual Behavior

CS1929 at the call site. The generator emits:

```csharp
public static ... HasItem<TActual, T>(this IAssertionSource<TActual> source, T item, ...)
    where TActual : NonSealedContainer
```

The call site `.HasItem<int>(42)` supplies one explicit type argument, but the method has two type parameters, and C# does not permit specifying only some of them. Overload resolution falls through to the source method (`Extensions.HasItem<int>(NonSealedContainer, int)`), which requires a raw `NonSealedContainer` receiver rather than `IAssertionSource<NonSealedContainer>`. That produces the CS1929 message above.

### Steps to Reproduce

Minimal repro against TUnit 1.44.39 / .NET 10:

```csharp
using TUnit.Assertions.Attributes;

public class NonSealedContainer { }

{
    [GenerateAssertion]
    public static bool HasItem<T>(this NonSealedContainer container, T item) => true;
}

public class Tests
{
    [Test]
    public async Task Repro()
    {
        var container = new NonSealedContainer();
        await Assert.That(container).HasItem<int>(42);
        // error CS1929: 'ValueAssertion<NonSealedContainer>' does not contain a
        // definition for 'HasItem' and the best extension method overload
        // 'Extensions.HasItem<int>(NonSealedContainer, int)' requires a receiver
        // of type 'NonSealedContainer'
    }
}
```

The same call with full inference (`.HasItem(42)`) compiles. The same call with explicit `<int>` against a `sealed` receiver compiles. The same non-generic call against a non-sealed receiver compiles.

### TUnit Version

1.44.39

### .NET Version

.NET 10.0 (also reproduces on .NET 8.0, .NET 9.0, and net472 via Polyfill).

### Operating System

Windows

### IDE / Test Runner

```shell
error CS1929: 'ValueAssertion<NonSealedContainer>' does not contain a definition for 'HasItem' and the best extension method overload 'Extensions.HasItem<int>(NonSealedContainer, int)' requires a receiver of type 'NonSealedContainer'
```

### Additional Context

The covariant receiver-type parameter helps when the call site uses full inference, but the source method's own explicit type arguments are the dominant call shape for generic `[GenerateAssertion]` methods on non-`string` receivers, and the two patterns can't both be satisfied by a single emitted signature.

Related: #5922 reports the same family of friction in the `[AssertionExtension]` generator path (class-attribute) rather than the `[GenerateAssertion]` (method-attribute) path covered here.

Have a repro test, fix, and multi-TFM snapshots prepared locally; will follow up with a PR.

### IDE-Specific Issue?

JohnVerheij:
## Description

Suppresses receiver-type covariance in `MethodAssertionGenerator` when the source method declares its own type parameters. The covariant `TActual` parameter previously prepended to the generated extension caused CS1929 at call sites supplying explicit type arguments, because C# does not permit partial type-argument specification. The fix is a single boolean conjunct (`genericParams.Count == 0`) plus an explanatory comment. Receivers of a more-derived static type can still reach the assertion via upcast.

## Related Issue

Fixes #5934

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

### Required

### Testing

### Decision matrix covered by the regression tests

`TUnit.Assertions.Tests/GenerateAssertionGenericMethodOnNonSealedReceiverTests.cs` adds nine cases:

1. generic method on non-sealed receiver with explicit type arg: was CS1929, now compiles
2. generic method on sealed receiver: control, still works (covariance was already disabled for sealed)
3. non-generic method on non-sealed receiver: control, still works
4. generic method on non-sealed receiver with full inference: control, still works
5. generic method with multiple type parameters (`<TFirst, TSecond>`): uniform behaviour, asserts the `genericParams.Count > 0` guard is not 1-specific
6. generic method on an interface receiver: `IsCovariantCandidate` accepts both class and interface; the fix applies to both
7. generic method with `where T : IParsable<T>` constraint: the constraint itself is incidental to the trigger, but represents the dominant real-world consumer shape
8. async result (`Task<AssertionResult>`) return type: orthogonal to the extension signature
9. derived static receiver via upcast: verifies the documented tradeoff

### Generator-level coverage

- `MethodAssertionGeneratorTests.MethodOnConcreteNonSealedReceiver` asserts the new emit shape: single `<T>` type parameter, exact-receiver `IAssertionSource<NonSealedReceiverType>`, no `TActual`, no `where TActual : ...`.
- Four new `*.verified.txt` snapshots (one per TFM: net472, net8.0, net9.0, net10.0). All four are byte-identical.

### Impact on existing snapshots

None. The TUnit codebase has no `[GenerateAssertion]` method that matches the affected shape (generic source method on a concrete non-sealed reference-type receiver). All existing snapshots regenerated identically. Full `TUnit.Assertions.Tests`, `TUnit.Assertions.SourceGenerator.Tests`, `TUnit.Assertions.Should.Tests`, and `TUnit.Assertions.Should.SourceGenerator.Tests` pass green locally.

### Related

`TUnit.Assertions.Should.SourceGenerator` was inspected; it does not use the covariance helper, so the bug is not present there.

The analogous case for the other generator path is covered by `AssertionExtensionGeneratorTests.ConcreteReceiverWithExtraGeneric`, which asserts the single merged generic-parameter list (no adjacent `<X><Y>` blocks) for `[AssertionExtension]` classes. The bug in `MethodAssertionGenerator` is the same family (two type parameters on the call signature versus one explicit argument at the call site), and the fix follows the same instinct.

Issue #5922 reports the user-visible friction in the `[AssertionExtension]` path; that is out of scope for this PR but tracked there.


thomhurst:
Thanks!

PR #5940, opened 2026-05-16
fix(generator): place CallerArgumentExpression before params in [GenerateAssertion] emit

### LINKED ISSUE #5939: [Bug]: [GenerateAssertion] emits CallerArgumentExpression after params, producing CS0231
### Description

`[GenerateAssertion]` emits invalid C# when the source method has a `params` array preceded by any other parameter. The generated extension method appends `[CallerArgumentExpression(...)]` diagnostic parameters AFTER the `params` array, which violates C#'s "params must be the last parameter" rule and produces compile error CS0231.

### Expected Behavior

The generator produces an extension method that compiles cleanly, with each `[CallerArgumentExpression]` parameter placed before the `params` array.

### Actual Behavior

The generated extension method puts `[CallerArgumentExpression]` parameters after the `params` array. The C# compiler rejects the file with `error CS0231: A params parameter must be the last parameter in a parameter list`.

### Steps to Reproduce

1. Create a static partial class with a `[GenerateAssertion]` source method that has at least one non-params parameter followed by `params`:

```csharp
using TUnit.Assertions.Attributes;

public static partial class Demo
{
    [GenerateAssertion]
    public static bool ContainsAny(this string value, string label, params string[] candidates)
        => candidates.Length > 0;
}
```

2. Build the containing project.

3. Observe the build fails with CS0231 in the generated file `Demo.GeneratedAssertions.g.cs`.

The generator emits:

```csharp
public static String_ContainsAny_String_StringArray_Assertion ContainsAny(
    this IAssertionSource<string> source,
    string label,
    params string[] candidates,
    [CallerArgumentExpression(nameof(label))] string? labelExpression = null)
```

The `params string[] candidates` parameter is followed by another parameter, which is illegal.

### TUnit Version

1.44.39 (also confirmed against current main at 420248e02)

### .NET Version

Reproduces on all targets supported by TUnit.Assertions (net472 / net8.0 / net9.0 / net10.0)

### Operating System

Windows

### IDE / Test Runner

```shell
error CS0231: A params parameter must be the last parameter in a parameter list
```

### Additional Context

Why production `[GenerateAssertion]` methods do not currently trip this: `GenericAssertions.IsIn` / `IsNotIn` are the only `[GenerateAssertion]` methods in `TUnit.Assertions` that use `params`, and both declare ONLY the `params` parameter (no preceding non-params source parameter). The generator's existing `if (!param.IsParams)` skip prevents emitting a `CallerArgumentExpression` for the `params` parameter itself, so the single-parameter shape stays legal C#. CS0231 fires only when at least one non-params source parameter sits before a `params` source parameter.

Workaround: drop the `params` modifier and accept a regular array; consumers then need C# 12 collection-expression syntax (`["a", "b"]`) at the call site instead of positional `"a", "b"`.

Fix direction: reorder the generator emit so the `[CallerArgumentExpression]` parameters for non-params source parameters precede the `params` parameter:

```csharp
public static String_ContainsAny_String_StringArray_Assertion ContainsAny(
    this IAssertionSource<string> source,
    string label,
    [CallerArgumentExpression(nameof(label))] string? labelExpression = null,
    params string[] candidates)
```

C# allows optional-with-default parameters to precede `params`, so the reorder is semantically transparent at every call site.

A PR with this fix and a regression-test fixture is ready to go alongside this issue.

### IDE-Specific Issue?

JohnVerheij:
## Description

Reorders the `[GenerateAssertion]` source generator emit so the auto-supplied `[CallerArgumentExpression]` parameters are placed BEFORE any `params` parameter in the generated extension method signature. The previous emit appended them AFTER the `params` array, which violates C#'s "params must be the last parameter" rule and produced `error CS0231` for any source method shaped like `(T first, params T[] rest)`.

### Implementation

`MethodAssertionGenerator.GenerateExtensionMethod` previously walked `data.AdditionalParameters` twice: once to emit the source parameters (including the `params` modifier in place), then once to emit `[CallerArgumentExpression]` diagnostic parameters at the end (skipping the `params` parameter itself). The skip stopped a CallerArgumentExpression from landing ON the `params` parameter, but did not stop diagnostic parameters for OTHER source parameters from landing AFTER it, which is the CS0231 trigger.

The fix splits the walk into three passes:

1. Emit each non-params source parameter, preserving default values.
2. Emit `[CallerArgumentExpression(nameof(...))]` for each non-params source parameter.
3. Emit the `params` source parameter last.

The constructor invocation and the expression-builder string already use the source order independently of the signature emission, so no other emit sites change.

### Behaviour for existing assertions

`IsIn` / `IsNotIn` in `GenericAssertions.cs` are the only production `[GenerateAssertion]` methods that use `params`. Both have ONLY the `params` parameter (no preceding non-params source parameter), so the previous emit was already legal C# for them and the new emit produces a byte-identical signature. The PublicAPI snapshots therefore do not change.

The CS0231 trigger only fires when a source method has at least one non-params source parameter sitting before a `params` source parameter, which production assertion sources do not currently use. The fix is verified against six edge cases in `TestData/ParamsParameterAssertion.cs`.

### Out of scope

A separate but related shape exists in `AssertionExtensionGenerator.cs:407-424`: it silently drops the `params` modifier, producing an ergonomic regression rather than CS0231 for `[AssertionExtension]` classes whose constructor takes `params`. Not in scope here; will file a separate issue if the direction in this PR is welcome.

## Related Issue

Fixes #5939

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

### Required

### Testing

### Regression tests added

A single test `MethodAssertionGeneratorTests.ParamsParameter` exercises the new `TestData/ParamsParameterAssertion.cs` fixture, which contains six `[GenerateAssertion]` shapes:

1. One required parameter followed by `params` (the canonical CS0231 trigger).
2. Two required parameters followed by `params`.
3. An optional defaulted parameter followed by `params`.
4. A generic-typed required parameter alongside a generic-typed `params` array.
5. The same canonical shape with `[GenerateAssertion(InlineMethodBody = true)]` to cover the inline emit branch.
6. A `params`-only source method (no preceding non-params parameter) to pin the byte-identical regression baseline used by `IsIn` / `IsNotIn`.

The test asserts the exact emitted signature for each shape with inline `.Contains(...)` substrings, then calls `CompileChecker.AssertNoErrors` to parse the generator output as C# and fail on any error diagnostic. That structural gate is the regression sentinel for CS0231 and any other emit defect that a content-only check cannot see.

Snapshots are committed for net472 / net8.0 / net9.0 / net10.0; the generator emit is TFM-agnostic and the four files are byte-identical.

### Local verification

- `TUnit.Assertions.SourceGenerator.Tests`: 66 / 66 pass on net10.0, 65 / 65 on net472 (`RefStructParameter` is gated `#if NET8_0_OR_GREATER`).
- `TUnit.Assertions.Tests`: 2106 / 2106 pass on net10.0.
- Build clean on net8.0 / net9.0 / net10.0 / netstandard2.0.


JohnVerheij:
Addressed in ef76e335f.

- MEDIUM (line 34, `fallback != null` on unconstrained `T`): replaced with
  `EqualityComparer<T>.Default.Equals(fallback, value)`. The body of
  `IsOneOfWithDefault` is test-fixture filler and the snapshot is byte-identical
  (not marked `InlineMethodBody`, so the body isn't reproduced in the generated
  extension method). CompileChecker and the six edge-case assertions stay green.

- MINOR (line 12, "partial is gratuitous"): left as-is. The source generator
  emits the second partial half of this class; without `partial` on the test-data
  side we get CS0260. Same pattern in every other TestData file in the directory.

PR #6008, opened 2026-05-24
Preserve specialized chaining after null assertions

thomhurst:
## Summary
- Add specialized `IsNull()` / `IsNotNull()` methods for collection, list, read-only list, dictionary, mutable dictionary, set, and async enumerable assertion bases.
- Preserve the domain-specific fluent surface after null checks, e.g. `.IsNotNull().And.ContainsKey(...)` and `.IsNull().Or.ItemAt(...)`.
- Add focused coverage for collection and async enumerable chaining after null assertions.

## Verification
- `dotnet build TUnit.Assertions\TUnit.Assertions.csproj --no-restore -p:NoWarn=CS1591 -clp:ErrorsOnly`
- `dotnet test TUnit.Assertions.Tests\TUnit.Assertions.Tests.csproj --no-restore --treenode-filter "/*/*/AsyncEnumerableAssertionTests/*" --no-progress --output Normal`
- `dotnet test TUnit.Assertions.Tests\TUnit.Assertions.Tests.csproj --no-restore --treenode-filter "/*/*/CollectionAssertionTests/*" --no-progress --output Normal`

## Notes
- I attempted the public API snapshot test, but the local `TUnit.PublicAPI` build is currently blocked by unrelated assembly version conflicts between `99.99.99.0` and `1.45.46.0` project outputs, so no `.received.txt` files were generated.

### DIFF
--- a/TUnit.Assertions.Tests/AsyncEnumerableAssertionTests.cs
+++ b/TUnit.Assertions.Tests/AsyncEnumerableAssertionTests.cs
@@ -149,6 +149,21 @@ public class AsyncEnumerableAssertionTests
             .Or.Contains(99); // First passes, so overall passes
     }
 
+    [Test]
+    public async Task Test_AsyncEnumerable_NullAssertions_Preserve_Chaining()
+    {
+        var items = AsyncRange(1, 5);
+
+        await Assert.That(items)
+            .IsNotNull()
+            .And.Contains(3);
+
+        IAsyncEnumerable<int>? nullItems = null;
+        await Assert.That(nullItems!)
+            .IsNull()
+            .Or.Contains(3);
+    }
+
     // Null handling
     [Test]
     public async Task Test_AsyncEnumerable_Null_Fails()
--- a/TUnit.Assertions.Tests/CollectionAssertionTests.cs
+++ b/TUnit.Assertions.Tests/CollectionAssertionTests.cs
@@ -130,6 +130,28 @@ public class CollectionAssertionTests
     }
 
     [Test]
+    public async Task NullAssertions_IsNotNull_Fails_When_Null_Through_Chain()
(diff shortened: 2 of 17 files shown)


thomhurst:
@
Addressed the latest review:

- **🔴 Dead extension method / return regression** — removed the shadowed `IsNotNull` extension in `AssertionExtensions.cs`. The base `IsNull()`/`IsNotNull()` instance methods now return the concrete `CollectionNullAssertion`/`CollectionNotNullAssertion` types, restoring the specific return type the extension gave.
- **🟡 Visibility asymmetry** — `CollectionNullAssertion` is now `public`, matching `CollectionNotNullAssertion`.
- **Failure-path tests** — already present (`NullAssertions_IsNotNull_Fails_When_Null_Through_Chain`, `NullAssertions_IsNull_Fails_When_NotNull_Through_Chain`).
- **Class proliferation** — already collapsed to a `bool expectNull` ctor for List/ReadOnlyList/Dictionary/MutableDictionary/Set in the current diff.

Pushing back on two:

- **🟡 `new` → virtual/override** — keeping `new` shadowing. It is the same idiom already used for `And`/`Or` on `CollectionAssertionBase`, and the only way to make `virtual`/`override` preserve each subtype's specific return type is covariant returns, which are **not honored at runtime on the `netstandard2.0` target** consumed by .NET Framework. Added a comment documenting this.
- **🟡 Exception-branch message** — keeping `"threw {ExceptionType}"`. This matches the convention of the sibling null-family assertions (`IsDefault`/`IsNotDefault*` all emit `"threw {name}"`); only the scalar `NotNullAssertion` swallows it into `"received null"`, which is the less informative path. These collection null classes are new, so no existing test depends on the old message.
@

PR #6025, opened 2026-05-28
fix(engine): isolate per-session state under MTP server-mode concurrency (#6001)

### LINKED ISSUE #6001: MTP server mode: multi-session findings from a mutation-testing prototype
Built a mutation-testing prototype that drives TUnit over MTP's JSON-RPC server mode (many sequential `testing/runTests` RPCs to one long-lived test process) and wanted to share what came up. Full writeup w/ architecture + benchmarks: [FINDINGS.md](https://github.com/slang25/TUnit/blob/slang25/mutation-testing-prototype/TUnit.Mutation.Prototype/FINDINGS.md).

## Already fixed in main ✅

`CreateTestSessionAsync` and `TestsCollector` drained `Sources.*` queues on first session — second `runTests` RPC saw empty queues and ran 0 tests. PR #2600 ("A bit of a rewrite") removed `TestsCollector.cs` and moved discovery to `Sources.TestEntries` (`ConcurrentDictionary` iterated with `foreach`). Verified main (`7df126c73`) handles 3 sequential `testing/runTests` RPCs correctly — each delivered all 114 expected test results.

## Still observed on main

**1. Stochastic test results under warm-pool concurrency.** Dispatching N concurrent `testing/runTests` RPCs to one test process sometimes produces 0 test results from a session that should produce ~100+. With N=1 (sequential) it's deterministic and correct; concurrency introduces variance. We work around it client-side by treating \`results.Count == 0\` as a worker-broken signal and respawning. Probably timing-dependent static state. The prototype + a 114-test MathLib workload reproduces it under \`--warm-pool 4\` / \`--warm-pool 6\` (stable under \`--warm-pool 1\` and \`--warm-pool 8\`).

**2. Stryker.NET 4.14.2 \`--test-runner mtp\` fails for 271 of 335 mutants against the same benchmark** — `Pending` status with "Mutation X was not fully tested" warnings. Same project with `vstest` + xUnit completes cleanly. Likely downstream of (1) or something else in the MTP integration.

**3. \`runId\` silently coerced to \`Guid.Empty\` on malformed Guid input.** This may be MTP rather than TUnit but surfaces via TUnit: a JSON-RPC client sending a non-Guid \`runId\` string to \`testing/runTests\` gets all \`testNodeUpdate\` notifications attributed to \`Guid.Empty\`, no error or warning. Cost me ~20 minutes to find. Worth a separate issue against \`Microsoft.Testing.Platform\` if not TUnit's domain.

## Benchmark context

On a 280-line library with 114 tests (apples-to-oranges noted: different operator catalogs, different test frameworks):

| Tool | Wall-clock | Mutants | Killed | Score | ms / mutant |
|---|---|---|---|---|---|
| Stryker.NET 4.14.2 (vstest + xUnit) | 24.5s | 277 | 252 | 90.18% | 88 ms |
| Prototype (TUnit + MTP, --max-parallel 6) | **16.8s** | 253 | 219 | 87.3% | **66 ms** |

MTP server-mode + warm-pool is a real performance win for tooling like this when the server-mode reliability story is solid. The prototype's coverage-directed batching + single-compile architecture wins 31% wall-clock and 25% per-mutant despite a smaller operator catalog (9 families vs Stryker's ~25). The bottleneck is currently the stochastic concurrency issue above — fix that and IDEs / mutation testers / coverage trackers all win.

Happy to provide more concrete repros, or PR specific fixes if useful. Otherwise just sharing the findings. Thanks for the great framework!

thomhurst:
Closes #6001 (finding 1; finding 2 is downstream; finding 3 is upstream MTP).

## Summary

Concurrent `testing/runTests` RPCs against one long-lived MTP test process were producing zero results from sessions that should produce ~100. Root cause was shared mutable state across sessions/calls — not a single race. This PR isolates each leak at its source rather than serialising RPCs per session.

## Findings & fixes

| # | File | Was | Now |
|---|------|-----|-----|
| 1 | `TUnit.Core/EngineCancellationToken.cs` | `Initialise` reassigned `CancellationTokenSource`+`Token` every call; loser inherited winner's cancellation | Session CTS established at construction; `Initialise` is idempotent and only hooks Ctrl+C / ProcessExit. Per-RPC cancellation flows through `context.CancellationToken` explicitly. |
| 2 | `TUnit.Engine/TUnitInitializer.cs` | Added new Trace listener + AppDomain/TaskScheduler closures per RPC, capturing stale `ExecuteRequestContext` | Split into per-process (Interlocked-guarded) handlers and per-session work; AppDomain handler no longer captures context |
| 3 | `TUnit.Engine/Discovery/ReflectionTestDataCollector.cs` | `_scannedAssemblies` + `_discoveredTests` were `static`; session N+1 saw assemblies as already-scanned and returned `[]` | Both are now instance fields |
| 4 | `TUnit.Engine/Discovery/ReflectionInstanceFactory.cs` | Process-wide instance cache leaked across sessions | Dropped the cache (reflection generic-type discovery is a cold path; isolation > dedup) |
| 5 | `TUnit.Engine/Discovery/ReflectionHookDiscoveryService.cs` | `Interlocked.Increment` guard short-circuited concurrent first callers past empty `Sources.*` bags | `ManualResetEventSlim` gate: concurrent callers block until the populator finishes |
| 6 | `TUnit.Engine/Logging/Standard{Out,Error}ConsoleInterceptor.cs` | Each session called `Console.SetOut(this)` and reset it on dispose, tearing down sibling sessions' interception | Install exactly once per process. Interceptor dispatches via `Context.Current` (AsyncLocal), so one instance serves every session |
| 7 | `TUnit.Core/Sources.cs` | No documented invariant; reflection mode wiped these | Doc invariant: hook bags are append-only after init |
| 8 | `TUnit.Engine/Reporters/Html/ActivityCollector.cs` | `_current` static single-slot, first-wins | Limitation documented; full fix deferred (orthogonal to the discovery/execution race and not exercised by the reporter's workload) |

## New tests

- `TUnit.RpcTests/Tests.cs::RunTests_ConcurrentRpcsOnSameSession_AllReceiveFullResults` — dispatches 4 concurrent `testing/runTests` RPCs against one `TestHostSession`, across 3 iterations, asserts every RPC receives the full result set. Pins finding 1.
- `TUnit.RpcTests/Tests.cs::RunTests_WithMalformedRunId_ServerAttributesUpdatesToGuidEmpty` — `[Skip]`ped repro for finding 3 (MTP coerces non-Guid `runId` to `Guid.Empty`). Will be linked from an upstream `microsoft/testfx` issue.

## Why not serialise per session

It would slow legitimate concurrent runs and would hide the underlying leaks rather than fix them — the static caches and per-call mutations would still misbehave for the reflection-mode discovery path, sequential N=many runs, and any future caller. Each leak is fixed at its source instead.

## Verification

- New concurrent regression test passes locally for `net8.0` and `net10.0` (10 RPCs across 3 iterations, no zero-result sessions).
- All existing `TUnit.RpcTests` pass (`Discovery_*`, `RunTests_With*`).
- Public API surface for `EngineCancellationToken` unchanged at the snapshot level (`Token { get; }`, `Dispose()`).

## Test plan

- [ ] CI green on all matrix legs
- [ ] Reporter (@slang25) re-runs the mutation-testing prototype with `--warm-pool 4/6` against this branch and confirms no zero-result sessions
- [ ] Stryker.NET 4.14.2 `--test-runner mtp` against the same workload — expect `Pending` count to drop from 271/335 to ~0 (finding 2)
- [ ] Upstream issue filed against `microsoft/testfx` for finding 3 with the skipped test as the repro

### DIFF
--- a/TUnit.Core/EngineCancellationToken.cs
+++ b/TUnit.Core/EngineCancellationToken.cs
@@ -1,4 +1,4 @@
-﻿using TUnit.Core.Settings;
+using TUnit.Core.Settings;
 
 namespace TUnit.Core;
 
@@ -10,25 +10,36 @@ public class EngineCancellationToken : IDisposable
     /// <summary>
     /// Gets the internal cancellation token source.
     /// </summary>
-    internal CancellationTokenSource CancellationTokenSource { get; private set; } = new();
+    internal CancellationTokenSource CancellationTokenSource { get; } = new();
 
     /// <summary>
     /// Gets the cancellation token.
     /// </summary>
-    public CancellationToken Token { get; private set; }
+    public CancellationToken Token { get; }
 
+    private int _initialised;
     private volatile bool _forcefulExitStarted;
 
+    public EngineCancellationToken()
+    {
+        Token = CancellationTokenSource.Token;
+    }
+
     /// <summary>
-    /// Initializes the cancellation token with a linked token source.
+    /// Hooks up process-wide cancellation signals (Ctrl+C / ProcessExit) the first time it's
+    /// called for this instance. Idempotent — subsequent calls are no-ops so that concurrent
+    /// MTP server-mode RPCs against one session don't clobber each other's cancellation chain.
+    /// Per-call cancellation flows through the explicit <c>CancellationToken</c> threaded into
+    /// discovery/execution, not through this session-scoped token.
     /// </summary>
-    /// <param name="cancellationToken">The cancellation token to link with.</param>
     internal void Initialise(CancellationToken cancellationToken)
     {
-        CancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
-        Token = CancellationTokenSource.Token;
+        _ = cancellationToken;
 
-        Token.Register(_ => Cancel(), this);
+        if (Interlocked.CompareExchange(ref _initialised, 1, 0) != 0)
+        {
+            return;
+        }
 
         // Console.CancelKeyPress is not supported on browser platforms
 #if NET5_0_OR_GREATER
--- a/TUnit.Core/EngineCancellationToken.cs
+++ b/TUnit.Core/EngineCancellationToken.cs
@@ -32,10 +32,8 @@ public class EngineCancellationToken : IDisposable
     /// Per-call cancellation flows through the explicit <c>CancellationToken</c> threaded into
     /// discovery/execution, not through this session-scoped token.
     /// </summary>
-    internal void Initialise(CancellationToken cancellationToken)
+    internal void Initialise()
     {
-        _ = cancellationToken;
-
         if (Interlocked.CompareExchange(ref _initialised, 1, 0) != 0)
         {
             return;
--- a/TUnit.Core/Sources.cs
+++ b/TUnit.Core/Sources.cs
@@ -5,9 +5,6 @@ using TUnit.Core.Interfaces.SourceGenerator;
 
 namespace TUnit.Core;
 
-#if !DEBUG
-[System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]
-#endif
 /// <remarks>
 /// INVARIANT: these collections are populated once at module initialization (by source-gen
 /// in source-gen mode, by ReflectionHookDiscoveryService once-per-process in reflection mode)
@@ -16,6 +13,9 @@ namespace TUnit.Core;
 /// hooks (#6001). The reflection-mode discovery's one-time <c>ClearSourceGeneratedHooks</c>
 /// runs before any session executes — adding new clear calls during a session is a bug.
 /// </remarks>
+#if !DEBUG
+[System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]
+#endif
 public static class Sources
 {
     public static readonly ConcurrentQueue<Func<Assembly>> AssemblyLoaders = [];
--- a/TUnit.Core/Sources.cs
+++ b/TUnit.Core/Sources.cs
@@ -8,6 +8,14 @@ namespace TUnit.Core;
 #if !DEBUG
 [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]
 #endif
+/// <remarks>
+/// INVARIANT: these collections are populated once at module initialization (by source-gen
+/// in source-gen mode, by ReflectionHookDiscoveryService once-per-process in reflection mode)
+/// and must be treated as append-only for the remainder of the process lifetime. Engine code
+/// that calls <c>.Clear()</c> on any field here will starve concurrent MTP sessions of their
+/// hooks (#6001). The reflection-mode discovery's one-time <c>ClearSourceGeneratedHooks</c>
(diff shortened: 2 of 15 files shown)


thomhurst:
Thanks for the thorough review. Round-up of how each finding has been addressed:

**Issue 1 — discarded `CancellationToken` in `Initialise`**: Addressed in `f66f803e2` (pushed before this review): parameter removed entirely from the signature; `TUnitTestFramework.ExecuteRequestAsync` now calls `Initialise()`. The signature no longer claims to accept a token it ignores. Per-RPC cancellation flows through `context.CancellationToken` which `TestRequestHandler.HandleDiscoveryRequestAsync` and `HandleRunRequestAsync` already thread into `DiscoverTests(... cancellationToken ...)` and `ExecuteTests(... context.CancellationToken)` directly.

**Issue 2 — `_sessionInitialised` per-instance vs per-session confusion**: Addressed in `4dc07592d`: comment now explicitly states the flag is per-instance, protects concurrent ExecuteRequestAsync calls within ONE session, and points to `ReflectionHookDiscoveryService._discoveryStarted` / `_discoveryCompleted` as the cross-session gate.

**Issue 3 — exception during `DiscoverHooks` leaves concurrent waiters in partial state**: Addressed in `4dc07592d`: added `_discoveryException` slot. The first caller's exception is captured before the finally sets `_discoveryCompleted`; concurrent waiters detect the failure and rethrow it wrapped, so no session silently uses partial `Sources.*` bags.

**Issue 4 — static `s_parametersParsed` + instance `commandLineOptions` smell**: Acknowledged but deferred — fixing properly means moving parameter parsing out to an explicit per-process startup callback, which is a bigger surface change than belongs in this PR. In practice MTP only ever instantiates one `ICommandLineOptions` per process, so the race is theoretical. Filed as a follow-up consideration.

**Issue 5 — `Close`/`DisposeAsync` call no-op `ResetDefault`**: Addressed in `4dc07592d`: added comments at both call sites in `OptimizedConsoleInterceptor` explaining the intent so future readers don't reintroduce per-session Console.SetOut/Reset churn.

**Minor — CA1416 suppression placement / justification**: Addressed in `f66f803e2`: justification rewritten to say "Reflection-mode hook discovery is not reachable on browser targets — those scenarios use AOT/source-gen mode which never invokes this method." The whole-method suppression is correct since the only platform-gated call site (`Console.CancelKeyPress` / `AppDomain.ProcessExit`) lives on a downstream code path that the method enters before checking.

thomhurst:
Pushed `415ffa7ec` to address review finding 4 properly — no more deferral.

Split the process-wide setup out of `TUnitInitializer` into a new `TUnitProcessInitializer` static type:

- `TUnitProcessInitializer.EnsureInitialised(ICommandLineOptions)` — global exception handlers, `ThrowListener`, and `--test-parameter` parsing. Single static guard; explicit `ICommandLineOptions` parameter, so the "first instance's options win" smell is gone — the dependency is passed in at the call site rather than smuggled through a captured constructor argument.
- `TUnitInitializer` — now only does per-session work (hook discovery, working directory). Delegates process setup to the new type. Per-instance `_sessionInitialised` flag remains and now exclusively serves its real purpose (gating concurrent ExecuteRequestAsync within one session).

The type name makes the lifetime distinction visible up front; the static-flag + instance-dep mismatch is gone.

PR #6122, opened 2026-05-29
feat(arguments): bind array values to a single array test parameter

### LINKED ISSUE #6120: [Feature]: String arrays in data driven test arguments
### Problem Statement

It'd be useful if array parameters can be passed through the `Arguments` attribute

### Proposed Solution

MSTest supports this through a string[] constructor:
https://github.com/microsoft/testfx/blob/191e07b66b2216dc37b857b1e481350293849109/src/TestFramework/TestFramework/Attributes/DataSource/DataRowAttribute.cs#L33

It'd be great if TUnit could support something like:
```csharp
[Test]
[Arguments(["Chloe"])]
[Arguments(["Skipper", "Lucy"])]
public async Task FavouriteNamesTest(string[] names)
{
...
}
```

Maybe a constructor like this could be added?
```csharp
public Arguments(string[] values)
{
...
}
```

### Alternatives Considered

_No response_

### Feature Category

Test Discovery / Attributes

### Additional Context

_No response_

### Contribution

- [x] I'm willing to submit a pull request for this feature

thomhurst:
Closes #6120

## Problem

`[Arguments(["Chloe"])]` / `[Arguments(["Skipper", "Lucy"])]` could not target a `string[] names` parameter (MSTest-style `string[]` DataRow support).

`ArgumentsAttribute` takes `params object?[]`, so a collection expression `["a","b"]` binds to the params array — **identical** to `[Arguments("a","b")]`. There is no syntactic way for the attribute to know "this is one array arg"; disambiguation must come from the test method signature. Before this change a non-`params` `T[]` parameter either dropped the extra values or threw `Expected exactly 1 argument`.

## Fix

A trailing array test-parameter (`params T[]` **or** a plain `T[]`) now collects the supplied values into the array, mirroring `params` binding:

```csharp
[Test]
[Arguments(["Chloe"])]
[Arguments(["Skipper", "Lucy"])]
public async Task Names(string[] names) { ... }   // works
```

Applied in **both** execution modes:
- **Source-gen** — `TupleArgumentHelper` + `TestMetadataGenerator` switch-case generation, via a new `IParameterSymbol.CollectsTrailingArguments()` extension.
- **Reflection** — `ReflectionTestDataCollector`.

The existing single-boxed-array passthrough (`x is T[] arr ? arr`) and null handling are preserved. The strongly-typed generic form `[Arguments<T[]>([...])]` already worked and is now documented as the unambiguous path in the `ArgumentsAttribute` XML docs.

## Tests / verification

- New dual-mode cases in `ArgsAsArrayTests` (`string[]`, `int[]`, single value, multiple values, generic) — pass in source-gen **and** reflection (`--reflection`) on net10.0.
- Source-generator snapshot tests: 464 pass / 0 fail (`.verified.txt` updated for the intended codegen diffs).
- No regression: `params string[]` and Issue2589 params tests still green.

## Notes

- Out of scope: jagged/multidimensional arrays.
- Pre-existing (not introduced here): the two `params IEnumerable<string>` tests in the same fixture (`ParamsEnumerable`, `Following_Non_Params`) fail on `main` too — a separate C#13 params-collections binding gap, left for a follow-up.


thomhurst:
Re: the per-TFM snapshot finding — the `net8.0`/`net9.0`/`net4.7` variants of the `ConflictingNamespaceTests` fixtures were stale because the earlier accept only ran the `net10.0` test host. The generator output is **not** TFM-forked — all four variants now emit the same LINQ trailing-array fallback. Regenerated and verified green under all four TFMs (net8.0/net9.0/net10.0/net472) and pushed. There's no behavioural divergence between targets.

PR #6123, opened 2026-05-29
feat(analyzers): suppress VSTHRD200 on test and hook methods

### LINKED ISSUE #6121: [Feature]: Suppress VSTHRD200 on methods with BaseTestAttribute/TestAttribute
### Problem Statement

VSTHRD200 has suggestions for adding `Async` at the end of all async methods. This is good for more methods but isn't ideal for tests. It'd be great if it can be suppressed

### Proposed Solution

Add an analyzer to suppress VSTHRD200

### Alternatives Considered

MSTest and XUnit have suppressions for this:
MSTest (MIT): https://github.com/microsoft/testfx/pull/2926
XUnit (apache 2.0 license): https://github.com/xunit/xunit.analyzers/commit/4f2297baf360f712ab6135b9ca996130124fb6cf

I think it would be possible to copy one of their implementations with minimal changes as long as it's properly attributed

### Feature Category

Other

### Additional Context

_No response_

### Contribution

thomhurst:
## What

Adds `Vsthrd200AsyncSuffixSuppressor`, a Roslyn `DiagnosticSuppressor` that silences **VSTHRD200** ("Use `Async` suffix for async methods") on:

- **Test methods** — any method decorated with an attribute deriving from `TUnit.Core.BaseTestAttribute` (covers `[Test]`, `[DynamicTestBuilder]`, and any future subclass).
- **Hook methods** — `[Before]` / `[After]` / `[BeforeEvery]` / `[AfterEvery]`.

VSTHRD200 is useful for production code but noisy on tests: `[Test] async Task MyTest()` shouldn't be forced to `MyTestAsync`. This mirrors the convention MSTest and xUnit already ship.

## How

- New suppressor follows the existing `MarkMethodStaticSuppressor` (CA1822) pattern.
- New `MethodExtensions.HasTestAttribute` walks the attribute inheritance chain via the existing `GetSelfAndBaseTypes()` helper. Left the existing exact-match `IsTestMethod` untouched to avoid changing behavior of the ~14 analyzers that call it.
- Test project references `Microsoft.VisualStudio.Threading.Analyzers` (analyzer dll, `Analyzer="false"`) so the VSTHRD200 analyzer type is loadable in suppressor tests.

No `AnalyzerReleases.*.md` change — suppressors are not tracked there.

## Tests

`Vsthrd200AsyncSuffixSuppressorTests` — VSTHRD200 suppressed on `[Test]` + 4 hook variants; **not** suppressed on a plain async non-test method. All pass.

Closes #6121

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -35,6 +35,7 @@
     <PackageVersion Include="Microsoft.CodeAnalysis.SourceGenerators.Testing" Version="1.1.2" />
     <PackageVersion Include="Microsoft.CodeAnalysis.Workspaces.Common" Version="4.7.0" />
     <PackageVersion Include="Microsoft.CSharp" Version="4.7.0" />
+    <PackageVersion Include="Microsoft.VisualStudio.Threading.Analyzers" Version="17.13.61" />
     <PackageVersion Include="Microsoft.EntityFrameworkCore" Version="10.0.8" />
     <PackageVersion Include="Microsoft.Extensions.DependencyInjection" Version="10.0.8" />
     <PackageVersion Include="Microsoft.Extensions.Http" Version="10.0.8" />
--- a/TUnit.Analyzers.Tests/TUnit.Analyzers.Tests.csproj
+++ b/TUnit.Analyzers.Tests/TUnit.Analyzers.Tests.csproj
@@ -49,6 +49,13 @@
              Link="Shared\AnalyzerTestCompatibility.cs" />
   </ItemGroup>
 
+  <!-- Reference the VSTHRD analyzer assembly (Analyzer="false") so its analyzer types are
+       usable in suppressor tests via WithAnalyzer<>, without running it against this project. -->
+  <ItemGroup>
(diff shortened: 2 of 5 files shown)


thomhurst:
Addressed in the latest commit:
- **Suppress() static**: hoisted the single descriptor into a `static readonly` field so `Suppress` is now `static` and references it directly.
- **[0] vs First()**: settled on direct index access with a documenting comment (this oscillated between review passes — the two are equivalent given one descriptor; `[0]` avoids the per-call scan).

Declined:
- **Custom `BaseTestAttribute` subclass test**: not viable — `BaseTestAttribute`'s constructor is `internal`, so external code cannot derive from it. The inheritance-aware `HasTestAttribute` path is already validated by the `[DynamicTestBuilder]` case, which is a genuine `BaseTestAttribute` subclass distinct from the exact `[Test]`.
- **`IsTestMethod` null-forgiveness** (item 4): pre-existing and out of scope, as noted.

PR #6124, opened 2026-05-29
fix: populate retry/flaky attempt history in HTML report (#6119)

### LINKED ISSUE #6119: [Bug]: HTML Report flaky test detection and summary not showing
### Description

When using TUnit retry functionality (e.g. [assembly: Retry(3)]), retries are executed correctly, but the generated HTML report does not display retry information.

The report renderer contains support for:

Retry badges (↻n)
"Flaky in this run" panel
Attempt timeline (attempts section)

However, the generated report payload only contains the final test result and retry-related span attributes (tunit.test.retry_attempt), but does not contain the attempts collection expected by the UI.

As a result, tests that fail and later pass are shown as ordinary passing tests with no indication that retries occurred.

### Expected Behavior

When a test is retried:

The report should display retry information.
Tests that eventually pass after one or more failures should be marked as flaky.
The "Flaky in this run" panel should contain the test.
Retry badges should be visible in the test list.
Attempt history should be available in the test details view.

### Actual Behavior

Retries are executed correctly, but no retry information is displayed in the report.

The report shows:

Test status = Passed
No retry badge
No flaky test entry
No attempt timeline

### Steps to Reproduce

1. Create a test that fails several times before passing:
``` csharp
private static int _attempt;

[Test]
public void RetryTest()
{
    _attempt++;

    Console.WriteLine($"Attempt {_attempt}");

    if (_attempt < 3)
    {
        Assert.Fail();
    }
}
```
2. Enable retries:
```csharp
[assembly: Retry(3)]
```
3. Execute the tests.
4. Open the generated HTML report.
The test passes.

Output contains:

```
Attempt 1
Attempt 2
Attempt 3
```
Retry spans are present in the report payload.
No retry badge, flaky marker, or attempt history is displayed in the report.

### TUnit Version

1.47.0

### .NET Version

.NET 10.0

### Operating System

Windows

### IDE / Test Runner

```shell

```

### Additional Context

_No response_

### IDE-Specific Issue?

thomhurst:
@-


thomhurst:
Re: the remaining `ITestExecution.RetryAttempts` source-break item — the proposed default interface member can't be used here. `TUnit.Core` targets `netstandard2.0;net8.0;net9.0;net10.0`, and default interface implementations require a runtime that supports them; adding `{ get => []; }` to the interface fails to compile for the `netstandard2.0` target (CS8701).

The alternatives are (a) move `RetryAttempts` off the interface onto the concrete `TestContext` only, or (b) accept the additive interface break. We're going with (b): keeping it on `ITestExecution` (where it's useful to custom reporters/adapters) and accepting that external implementors of the interface must add the member — consistent with how the API otherwise evolves. Leaving as-is intentionally.

PR #6132, opened 2026-05-31
feat(html-report): baked-in C# syntax highlighting on Source tab

slang25:
## What

Adds a small, **zero-dependency C# syntax highlighter** to the HTML test report's **Source** tab. Previously the fetched source was rendered as plain, uncolored text.

## How

- A hand-rolled tokenizer (`highlightCSharpLines`) classifies comments (line/block/XML-doc), strings (regular, verbatim `@"…"`, interpolated `$"…"`, raw `"""…"""` including multi-line), char and numeric literals (hex/binary/float suffixes/digit separators), keywords, method calls (`ident(`) and PascalCase types.
- It tokenizes the whole snippet so multi-line constructs render correctly, then splits into per-line HTML. All output is HTML-escaped via the existing `esc()`.
- Token colors are driven by `--syn-*` CSS custom properties defined **per theme**, so highlighting follows the existing light/dark toggle instantly with no JS re-render.
- The palette mirrors the docs site's Prism themes: **github** for light, **dracula** for dark.
- Non-`.cs` files fall back to plain escaped text.

All changes live in the embedded `TestReport.template.html` resource, outside the `SAMPLE_DATA` markers and `__REPORT_DATA__` placeholder, so report generation is unaffected. No C# changes, no new dependencies.

## Preview

Light (github) / Dark (dracula), switching with the theme toggle:

| Light | Dark |
|-------|------|
| keywords `#00009f`, types `#6f42c1`, strings `#e3116c`, comments `#999988` italic, numbers `#36acaa`, calls `#d73a49` | keywords `#bd93f9`, types `#8be9fd`, strings `#ff79c6`, comments `#6272a4`, numbers `#bd93f9`, calls `#50fa7b` |

## Notes

- The tokenizer deliberately trades edge-case accuracy for a tiny footprint. Contextual keywords that are commonly ordinary identifiers (`value`, LINQ query words, etc.) are intentionally excluded to avoid miscoloring.
- Verified that token line counts always equal source line counts (line numbers stay aligned), and visually checked rendering in both themes.

### DIFF
--- a/TUnit.Engine/Reporters/Html/TestReport.template.html
+++ b/TUnit.Engine/Reporters/Html/TestReport.template.html
@@ -35,6 +35,14 @@
     --shadow-sm: 0 1px 2px rgba(15,15,15,0.04);
     --shadow:    0 1px 2px rgba(15,15,15,0.04), 0 1px 4px rgba(15,15,15,0.04);
     --shadow-pop:0 12px 32px rgba(15,15,15,0.10), 0 0 0 1px rgba(15,15,15,0.05);
+    /* C# syntax highlighting — matches the docs site's Prism "github" light theme. */
+    --syn-plain:   #393a34;
+    --syn-comment: #999988;
+    --syn-string:  #e3116c;
+    --syn-keyword: #00009f;
+    --syn-number:  #36acaa;
+    --syn-func:    #d73a49;
+    --syn-type:    #6f42c1;
   }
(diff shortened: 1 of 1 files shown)


slang25:
The alternative is something like highlight.js and we reference that, this would be more accurate in terms of highlighting

thomhurst:
Nice! 😄

PR #6141, opened 2026-06-01
fix(sourcegen): fully-qualify Linq calls in params array binding (#6140)

### LINKED ISSUE #6140: [Bug]: regression in v1.48.0 (non-compiling generated code)
### Description

After upgrading from v1.47 to v1.48, several compiler errors are reported in TUnit-generated source files. 

Each test method impacted uses an array (either `params` or plain) as its last argument, so this may likely be related to https://github.com/thomhurst/TUnit/pull/6122

A concrete example -- this test method:

```csharp
[Test]
[Arguments(RollingOrder.NewestFirst, 2, 1)]
[Arguments(RollingOrder.NewestLast, 1, 2)]
public void Add_ShouldAddItemToList_WhenListIsNotFull(RollingOrder rollingOrder, params int[] expected)
```

Results in the following code generated by `TUnit.Core.SourceGenerator.Generators.TestMetadataGenerator` (reformated slightly):

```csharp
private static global::System.Threading.Tasks.ValueTask __Invoke(global::Namespace.RollingListTests instance, int methodIndex, object?[] args, global::System.Threading.CancellationToken cancellationToken)
{
    switch (methodIndex) {
    case 0: {
        try {
            switch (args.Length) {
            case 1: /* ... */
            case 2: /* ... */
            case 3: /* ... */
            case 4: /* ... */
            case 5: /* ... */
            case 6: /* ... */
            case 7: /* ... */
            default:
                if (args.Length < 1) {
                    throw new global::System.ArgumentException($"Expected at least 1 argument, but got {args.Length}");
                }
                instance.Add_ShouldAddItemToList_WhenListIsNotFull(global::TUnit.Core.Helpers.CastHelper.Cast<global::Namespace.RollingOrder>(args[0]),
                    (args.Length > 1 ?
                        global::System.Linq.Enumerable.Range(1, args.Length - 1)/*👉*/.Select(i => global::TUnit.Core.Helpers.CastHelper.Cast<int>(args[i])).ToArray()
                        : new int[0]));
                return default(global::System.Threading.Tasks.ValueTask);
            }
        }
        catch (global::System.Exception ex) {
            return new global::System.Threading.Tasks.ValueTask(global::System.Threading.Tasks.Task.FromException(ex));
        }
    }
```

Note the LINQ `Select` extension method call at the location marked with <mark>👉</mark>, which is not full-qualified (like `global::System.Linq.Enumerable.Range` earlier) and whose declaring namespace is not imported at this point.

### Expected Behavior

The generated code compiles.

### Actual Behavior

A compiler error is reported in the generated code at the location marked with <mark>👉</mark>

> Namespace_RollingListTests__TestSource.g.cs(88,261): Error CS1061 : 'IEnumerable<int>' does not contain a definition for 'Select' and no accessible extension method 'Select' accepting a first argument of type 'IEnumerable<int>' could be found (are you missing a using directive or an assembly reference?)

### Steps to Reproduce

I tried to extract a self-contained minimal repro but did not manage to get one in reasonable time, so instead I am reporting the symptoms observed.

### TUnit Version

1.48.0

### .NET Version

.NET 10

### Operating System

macOS

### IDE / Test Runner

### IDE-Specific Issue?

thomhurst:
Fixes #6140

## Problem

After #6122, the source generator's dynamic-count params/array argument binding emits:

```csharp
global::System.Linq.Enumerable.Range(start, count).Select(i => ...).ToArray()
```

`Select` and `ToArray` are used here as **extension methods**, which require `using System.Linq` to be in scope. Generated `.g.cs` files only get `System.Linq` when the consuming project has `ImplicitUsings` enabled (via the auto-generated global usings). Projects with implicit usings disabled fail to compile:

> CS1061: 'IEnumerable<int>' does not contain a definition for 'Select' ...

(This is why the reporter couldn't make a minimal repro — it depends on the project's `ImplicitUsings` setting, not the test itself.)

## Fix

`TupleArgumentHelper.cs` now emits the **fully-qualified static** form:

```csharp
global::System.Linq.Enumerable.ToArray(
    global::System.Linq.Enumerable.Select(
        global::System.Linq.Enumerable.Range(start, count),
        i => ...))
```

No extension-method resolution, so it compiles regardless of the consuming project's usings.

## Tests

Affected snapshots regenerated and committed (ArgsAsArrayTests, DataDrivenTests, MethodDataSourceDrivenTests, Tests2112, ConflictingNamespaceTests across all TFMs).

> Note: the existing snapshot harness masks this class of bug — `TestsBase.cs` injects `global using global::System.Linq;` and only checks generator diagnostics, not the resulting compilation's C# errors. The regenerated snapshots now lock the fully-qualified form, guarding against regression.


thomhurst:
Thanks for the reviews. Responding to each:

**Review #2 — "regression test never exercises the dynamic LINQ path":** This is incorrect. The premise conflates *runtime reachability* with *compilation*. The `default:` arm (`TestMetadataGenerator.cs:2697`) is emitted for **every** params method regardless of arg count, and the C# compiler compiles the `default:` arm whether or not it is reached at runtime. CS1061 is a compile error, so a 3-argument test still triggers it.

I verified empirically: reverting only the `TupleArgumentHelper.cs` fix and re-running the test produces exactly the issue's error, twice (one per method's `default:` arm):

```
MyTests_RollingListTests__TestSource.g.cs(85,201): error CS1061: 'IEnumerable<int>' does not contain a definition for 'Select' ...
MyTests_RollingListTests__TestSource.g.cs(136,124): error CS1061: 'IEnumerable<int>' does not contain a definition for 'Select' ...
```

So the test does guard the fix. No change needed; an 11+ arg case would only add runtime-reachability coverage, which is irrelevant for a compile-only assertion.

**Review #1, point 3 — `TestsBase` warning comment:** Good call, added in 591b07d9e.

**Review #1, point 1 — `Array.ConvertAll` / hand-emitted loop:** Agreed it's a reasonable micro-optimization, but this runs at test-discovery argument binding (not a per-iteration hot path), and the reviewer flagged it as non-blocking. Leaving the fully-qualified LINQ form to keep the change minimal; can revisit as a follow-up.

PR #6153, opened 2026-06-02
feat(mocks): discriminate generic-method mocks by type argument

thomhurst:
## Problem

[Discussion #4981](https://github.com/thomhurst/TUnit/discussions/4981) (latest comment) reported that TUnit.Mocks cannot distinguish generic-method setups by their type argument:

```csharp
public interface IGreeter { string Greet<T>() where T : class; }

var mock = IGreeter.Mock();
mock.Greet<Class1>().Returns("Hello!");
mock.Greet<Class2>().Returns("Goodbye!");

greeter.Greet<Class1>(); // returned "Goodbye!" — both setups collided
```

**Root cause:** the whole match key was `(memberId, argument-matchers)`. A generic method's *type arguments* (`typeof(Class1)` vs `typeof(Class2)`) were dropped at the runtime boundary — `MethodSetup`, `CallRecord`, and `FindMatchingSetup` had no notion of them. The existing `Generic_Method_Different_Type_Arguments` test only passed because it also varied a normal `int` arg, which masked the bug. The reported case has zero parameters, so the two setups were byte-identical.

## Fix

Thread an optional `Type[]? typeArguments` (closed `typeof(T)` captured at the call site) through setup registration, call dispatch, and verification. `null` ⇒ non-generic ⇒ identical behaviour to before.

- **Runtime:** `MethodSetup`/`CallRecord` carry the type args; `MockEngine` gains `object[]`+`Type[]` `HandleCall`/`HandleCallWithReturn` overloads and a `TypeArgumentsMatch` gate in `FindMatchingSetup`; `CallVerificationBuilder` filters recorded calls by type argument.
- **Generator:** generic methods emit `new Type[] { typeof(T), ... }` and route through the fallback dispatch (typed dispatch can't carry type args), preserving any auto-mock factory.
- **Wildcards:** new `AnyType` / `AnyValueType` markers (`TUnit.Mocks.Arguments`) let a setup match any type argument, e.g. `mock.Greet<AnyType>().Returns(...)`.

### Scope / limitations
- Exact closed-type matching plus the `Any` wildcard.
- Type params constrained to a specific base class/interface support exact matching only (no marker can satisfy an arbitrary base).
- Partial/wrap **virtual** methods record no type args and are not discriminated — graceful, documented degradation (their **abstract** members, which route through `HandleCallWithReturn`, *are* discriminated).

## Tests
New integration tests in `GenericTests.cs`:
- the zero-arg regression from the discussion;
- `AnyType` wildcard and exact-wins-over-wildcard;
- type-arg-aware verification;
- **multiple type parameters** — `(T1, T2)` order sensitivity, partial wildcards (one `AnyType` + one concrete), exact-wins-over-partial-wildcard, and multi-param verification with wildcard counts.

Plus updated generic-method snapshots.

**Green:** 1010 integration, 62 snapshot, 30 analyzer tests (net8.0 + net10.0); runtime builds across all TFMs + Roslyn variants.

## Note for reviewers
`FindMatchingSetup` keeps a separate no-default `(int, object?[])` overload. Collapsing it into a single defaulted-`Type[]?` overload makes 2-arg call sites silently bind to the generic `FindMatchingSetup<T1>` instead — which broke every wrap/partial mock during development. A comment marks the constraint.

### DIFF
--- a/TUnit.Mocks.SourceGenerator.Tests/Snapshots/Interface_FluentUI_Shape_Nullable_Warnings.verified.txt
+++ b/TUnit.Mocks.SourceGenerator.Tests/Snapshots/Interface_FluentUI_Shape_Nullable_Warnings.verified.txt
@@ -34,7 +34,7 @@ file sealed class IDialogReferenceMockImpl : global::IDialogReference, global::T
     {
         try
         {
-            var __result = _engine.HandleCallWithReturn<T?>(0, "GetReturnValueAsync", global::System.Array.Empty<object?>(), default, new global::System.Type[] { typeof(T) });
+            var __result = _engine.HandleCallWithReturn<T?>(0, "GetReturnValueAsync", global::System.Array.Empty<object?>(), default, global::TUnit.Mocks.TypeArguments.Of<T>.Value);
             if (global::TUnit.Mocks.Setup.RawReturnContext.TryConsume(out var __rawAsync))
             {
                 if (__rawAsync is global::System.Threading.Tasks.Task<T?> __typedAsync) return __typedAsync;
@@ -95,7 +95,7 @@ public static class IDialogReference_MockMemberExtensions
     public static global::TUnit.Mocks.MockMethodCall<T?> GetReturnValueAsync<T>(this global::TUnit.Mocks.Mock<global::IDialogReference> mock)
     {
         var matchers = global::System.Array.Empty<global::TUnit.Mocks.Arguments.IArgumentMatcher>();
-        return new global::TUnit.Mocks.MockMethodCall<T?>(global::TUnit.Mocks.MockRegistry.GetEngine(mock), 0, "GetReturnValueAsync", matchers, new global::System.Type[] { typeof(T) });
+        return new global::TUnit.Mocks.MockMethodCall<T?>(global::TUnit.Mocks.MockRegistry.GetEngine(mock), 0, "GetReturnValueAsync", matchers, global::TUnit.Mocks.TypeArguments.Of<T>.Value);
     }
 
     #if NET9_0_OR_GREATER
@@ -238,7 +238,7 @@ file sealed class IDialogServiceMockImpl : global::IDialogService, global::TUnit
     {
         try
         {
-            var __result = _engine.HandleCallWithReturn<global::IDialogReference?>(0, "UpdateDialogAsync", new object?[] { id, parameters }, default, static __behavior => global::IDialogReferenceMockFactory.CreateAutoMock(__behavior), new global::System.Type[] { typeof(TData) });
+            var __result = _engine.HandleCallWithReturn<global::IDialogReference?>(0, "UpdateDialogAsync", new object?[] { id, parameters }, default, static __behavior => global::IDialogReferenceMockFactory.CreateAutoMock(__behavior), global::TUnit.Mocks.TypeArguments.Of<TData>.Value);
             if (global::TUnit.Mocks.Setup.RawReturnContext.TryConsume(out var __rawAsync))
             {
                 if (__rawAsync is global::System.Threading.Tasks.Task<global::IDialogReference?> __typedAsync) return __typedAsync;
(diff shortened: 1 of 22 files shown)


thomhurst:
@
## Response to Round 6

**Internal cast → `InvalidCastException`** — Addressed in 44f3ea14a2, but not via the suggested default interface method: `TUnit.Mocks` targets `netstandard2.0` (`Library.props`), where DIMs do not compile (CS8701). Instead the hard cast is now a type test that falls back to the public, non-filtering `CreateVerification` overload — a custom `IMockEngineAccess` loses type-argument filtering for generic-method verification but never throws, which is the same graceful degradation the DIM would have provided.

**Wildcard vs. exact match order-dependent** — Not changing this; the analysis has the iteration direction inverted. `FindMatchingSetup` iterates **last-added-first** (`MockEngine.cs`: "Iterate last-added-first to implement \"last wins\" semantics"), so in the quoted example the call to `Greet<Class1>()` hits the exact setup at index 1 first and returns `"exact"`. Last-wins is the library-wide, documented, and tested semantic — for type arguments (`Generic_Method_Exact_Type_Setup_Wins_Over_Wildcard`) and equally for argument matchers (`Exact_Match_Overrides_Any_Match_Later_Wins`). Specificity scoring for type arguments alone would make them behave differently from `Arg.Any<T>()` matchers, which would be the real inconsistency.

**Mutable shared `Type[]`** — Declining. Converting to `IReadOnlyList<Type>`/`ReadOnlyMemory<Type>` ripples through `MethodSetup`, `CallRecord`, and `TypeArgumentMatching`, adding interface dispatch in the per-call matching hot path, to guard a `[EditorBrowsable(Never)]` field that only generated code references and whose doc comment states it must never be mutated. The hot-path cost is not worth the theoretical misuse.

**`FriendlyTypeName` drops namespace** — Declining for consistency: every other diagnostic in `TUnit.Mocks` uses simple `Type.Name` (`Mock.cs`, `MockEngine.cs`, `MockEngine.Typed.AutoMock.cs`). Namespace-qualifying only verification failure messages would be inconsistent and noisy for the common case.
@

thomhurst:
## Response to Round 7

Addressed in 0afbe71bb1:

**OrderedVerification drops type arguments** — Confirmed and fixed. `OrderedCallExpectation` now carries the type arguments, `RecordExpectation` threads them through, `FindMatchingCalls` applies `TypeArgumentMatching.Matches`, and ordered failure messages include the type arguments. Regression tests added (`Ordered_Verification_Discriminates_By_Type_Argument`, `Ordered_Verification_Fails_When_Type_Arguments_Out_Of_Order`).

**Mutable shared `Type[]`** — Implemented as suggested: type-argument arrays are `ImmutableArray<Type>` end-to-end (`TypeArguments.Of<T>.Value`, `MethodSetup`, `CallRecord`, engine dispatch, verification), with the arity-5+ generator fallback emitting `ImmutableArray.Create`. Indexed access on the hot matching path stays array-speed. All converted signatures were introduced on this branch, so no released surface changed.

**`CreateVerification()` duplication** — Deduped into `MockCallVerification.Create`, used by both `MockMethodCall` and `VoidMockMethodCall`.

**`FindMatchingSetup<T1..T8>` omits `TypeArgumentsMatch`** — Investigated; the suggested `Debug.Assert(setup.TypeArguments is null)` is wrong, and trying it proved it: `Generic_Abstract_Generic_Method_And_Event_Work` fails, because virtual/partial/wrap generic methods legitimately reach typed dispatch with type-argument-carrying setups under the documented graceful-degradation path. Since typed dispatch never carries call-side type arguments, `TypeArgumentsMatch` there would be a constant `true` — the omission is semantic, not an oversight. A comment on the family now documents both reachable paths and cross-references `ShouldGenerateTypedWrapper`.

**`ITypeArgumentMatcher` composability** — Declining for this PR as YAGNI: there is no user demand yet for predicate-style type-argument matching, and committing to a public matcher interface now constrains the API before the feature has usage feedback. The sentinel design does not block adding `ITypeArgumentMatcher` later non-breakingly (`AnyType`/`AnyValueType` would simply implement it).

**`eagerRegister` + `typeArguments` ctor gap** — Agreed it is not a current bug (property setters are non-generic); noting it as a known forward-compat point rather than adding an unused overload now.

Verified: 1015 mock tests + 62 generator snapshot tests pass, all TFMs build clean.

PR #6156, opened 2026-06-03
fix: dispose shared fixtures when only a subset of consuming tests runs

thomhurst:
## Problem

Discussion #6151: a `[ClassDataSource<T>(Shared = SharedType.PerTestSession)]` fixture implementing `IAsyncDisposable` never gets `DisposeAsync` called when running a single test from Rider (one of two `[Arguments]` cases).

## Root cause

Shared-object disposal is pure ref-counting (`ObjectTracker`): +1 per test that registers the object, -1 when a test finishes, dispose at 0.

The increment happened at **build time** (`TestBuilder.BuildTestAsync` -> `RegisterTestArgumentsAsync`), but filtering happens **after** building, and decrements only happen for **executed** tests. Any built-but-not-run test permanently inflated the count, so it never reached zero:

- A single `[Arguments]` case selected by an IDE uid filter (both cases share one metadata entry, so both are always built) - the reported repro
- An `[Explicit]` sibling matched-then-excluded by a filter
- Discovery-only requests (IDE server mode), which built and registered **all** tests - also meaning discovery constructed expensive fixtures (e.g. WebApplicationFactory) as a side effect

## Fix

1. **Move argument registration post-filter.** Removed the build-time `RegisterTestArgumentsAsync` call; `TestFilterService.RegisterTest` (which already performed the same call with identical failure semantics) is now the single registration point, running only for tests that will execute (incl. dependency-expanded ones).
2. **Discovery-only requests skip argument registration** (event receivers still run, so skip reasons / display-name formatters are unaffected). Discovery no longer creates shared fixtures at all.
3. **Dynamic tests register explicitly.** `AddDynamicTest` / `CreateTestVariant` bypass the discovery pipeline, so `TestRegistry` now registers built tests before enqueueing them (this also gives them `OnTestRegistered` receivers, which previously never fired for dynamic tests).
4. **Session-end sweep + cache reset.** At the end of each run session, any still-tracked objects are disposed (covers cancellation/miscounts) and `TestDataContainer` is cleared, so a subsequent run request in the same process (IDE server mode) creates fresh fixtures instead of receiving disposed ones. Runs after `After(TestSession)` hooks and static-property disposal - existing ordering preserved.

`ObjectTracker.ClearStaticTracking` previously had zero callers; the new sweep wires that lifecycle up properly.

## Tests (TDD - red first, then green)

- `TUnit.TestProject/Bugs/6151/Bug6151FilteredDisposalTests.cs` - PerTestSession fixture with a normal test + `[Explicit]` sibling (same built-but-not-run shape as the IDE uid-filter case, reproducible via `--treenode-filter`). A guarded `[After(TestSession)]` hook writes created/disposed counts to a marker file (opt-in via env var, inert otherwise).
- `TUnit.Engine.Tests/FilteredSharedFixtureDisposalTests.cs` - asserts `Created=1;Disposed=1` for the filtered-subset run (failed with `Disposed=0` before the fix) plus a direct-single-test sanity case.
- `InvokableTestBase.RunOptions` gained `WithEnvironmentVariable` to support the marker handshake.

## Verification

- New regression test red before fix (`Created=1;Disposed=0`), green after - both source-gen and reflection modes
- `TUnit.UnitTests`: 218/218
- Engine-test batch (dynamic, explicit, skip, retry, cancellation, data-source, property-injection, session/discovery hooks, disposal): 35/35 (AOT variants run in CI)
- TestProject disposal suites both modes: Bugs 1924 (660 tests), 5982, 2867, SharedDisposalTest, DisposalRegressionTests, NestedDisposalOrder - all green
- `TUnit.PublicAPI` snapshots updated for the two new `Clear()` members (`ScopedDictionary`, `ThreadSafeDictionary`)

Note: `ClassDataSourceTupleTests.Test_ClassDataSource_UnwrapsTuples` fails identically with and without this change (verified via stash) - pre-existing, unrelated.

Fixes #6151

### DIFF
--- a/TUnit.Core/Data/ScopedDictionary.cs
+++ b/TUnit.Core/Data/ScopedDictionary.cs
@@ -11,4 +11,9 @@ public class ScopedDictionary<TScope>
 
         return innerDictionary.GetOrAdd(type, factory);
     }
+
+    /// <summary>
+    /// Removes all scopes and their cached instances.
+    /// </summary>
+    public void Clear() => _scopedContainers.Clear();
 }
--- a/TUnit.Core/Data/ThreadSafeDictionary.cs
+++ b/TUnit.Core/Data/ThreadSafeDictionary.cs
@@ -103,4 +103,9 @@ public class ThreadSafeDictionary<TKey, [DynamicallyAccessedMembers(DynamicallyA
     public TValue this[TKey key] => _innerDictionary.TryGetValue(key, out var lazy)
         ? lazy.Value
         : throw new KeyNotFoundException($"Key '{key}' not found in dictionary");
+
+    /// <summary>
+    /// Removes all keys and values from the dictionary.
+    /// </summary>
+    public void Clear() => _innerDictionary.Clear();
 }
--- a/TUnit.Core/TestDataContainer.cs
+++ b/TUnit.Core/TestDataContainer.cs
@@ -29,4 +29,17 @@ internal static class TestDataContainer
     {
         return _keyContainer.GetOrCreate(key, type, func);
     }
+
+    /// <summary>
+    /// Clears all cached shared instances. Called at the end of a run session so that a
+    /// subsequent run request in the same process (e.g. IDE server mode) creates fresh
+    /// instances instead of reusing already-disposed ones.
+    /// </summary>
+    public static void Reset()
+    {
+        _globalContainer.Clear();
+        _classContainer.Clear();
+        _assemblyContainer.Clear();
+        _keyContainer.Clear();
+    }
 }
--- a/TUnit.Core/Tracking/ObjectTracker.cs
+++ b/TUnit.Core/Tracking/ObjectTracker.cs
@@ -37,6 +37,41 @@ internal class ObjectTracker(TrackableObjectGraphProvider trackableObjectGraphPr
     }
 
     /// <summary>
+    /// Disposes any objects still tracked with a positive reference count, then clears all
+    /// static tracking state. Call at the end of a run session: every executed test has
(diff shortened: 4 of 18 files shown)


thomhurst:
Addressed the review in d4d6cdf9b:

**Double-disposal of static properties — confirmed real, fixed.** `DisposeStaticPropertiesAsync` disposed directly via `Disposer`, leaving the +1 entry from `TrackStaticProperties` in `s_trackedObjects`, so the new session-end sweep would have disposed those objects a second time. It now goes through a new `ObjectTracker.UntrackObjectAsync` — decrement to zero, dispose once, entry removed. Verified with static-property suites in both modes.

**Instance method on static state** — documented: the method needs the injected `Disposer` (error logging), which only instances carry.

**`TestContext.Current = null` removal — no gap.** The only thing that set `TestContext.Current` in the build path was `RegisterTestArgumentsAsync` itself (`TestArgumentRegistrationService.cs:26`); with that call removed from `BuildTestAsync`, nothing in the build path sets it anymore, so there is nothing to reset. The post-filter `RegisterTest` call site has never reset it — identical to pre-PR behavior. (`InternalDiscoveredTest` assignment never touched `Current`.)

**`DiscoverTestsFullyStreamingAsync`** — it has zero callers today; added a warning comment that any future execution use must register tests with `isForExecution: true` first.

**Static counters not reset between runs** — fixed: the marker hook now resets `CreatedCount`/`DisposedCount` after writing, so a long-lived server process reports per-session counts.

**`RunOptions.With*` mutation** — left as-is, consistent with the existing `WithArgument` pattern as you noted.

PR #6158, opened 2026-06-03
fix(analyzers): decouple code fixers from Rules to prevent MissingFieldException in VS

### LINKED ISSUE #6157: [Bug]: MissingFieldException for Latest migration analyser in VS2026
### Description

At random parts of coding, I get an exception showing up in the VS IDE with the below error message:

```
MissingFieldException: Field not found 'TUnit.Analyzers.Rules.MSTestMigration'
    at TUnit.Analyzers.CodeFixers.MSTestMigrationCodeFixProvider.get_DiagnosticId()
    at TUnit.Analyzers.CodeFixers.Base.BaseMigrationCodeFixProvider.get_FixableDiagnosticIds()
...
```

I also get a similar error for NUnitMigration analyser too.

### Expected Behavior

No exceptions

### Actual Behavior

Very regular exceptions 

### Steps to Reproduce

Create normal tests and save file. 

### TUnit Version

Latest

### .NET Version

.net 9

### Operating System

Windows

### IDE / Test Runner

Visual Studio

```shell

```

### Additional Context

_No response_

### IDE-Specific Issue?

thomhurst:
Fixes #6157

## Root cause

`TUnit.Analyzers.CodeFixers.dll` ships in the version-agnostic `analyzers/dotnet/cs` folder while `TUnit.Analyzers.dll` ships per-Roslyn (`analyzers/dotnet/roslyn4.x/cs`), and the code fixers reference the analyzers assembly at runtime (ProjectReference, resolved by simple name — the assemblies are not strong-named).

Visual Studio cannot unload analyzer assemblies. After a TUnit package update without a VS restart (or with mixed TUnit versions in one VS session), the **new** code fixers bind against a **stale** `TUnit.Analyzers.dll`. VS evaluates `FixableDiagnosticIds` eagerly on every lightbulb pass, which JITs the field reference `Rules.MSTestMigration` / `Rules.NUnitMigration` — fields added in #3195 that the stale assembly lacks — throwing `MissingFieldException` "at random parts of coding".

The heavy migration machinery (`TUnit.Analyzers.Migrators.Base.*`) is only touched lazily inside `RegisterCodeFixesAsync`, which can only run when the matching (new) analyzer reported the diagnostic — so decoupling the eager members removes the failure mode entirely.

The same latent landmine existed in `TUnit.Assertions.Analyzers.CodeFixers`.

## Fix

- New `DiagnosticIds` classes with `public const string` per rule in `TUnit.Analyzers` and `TUnit.Assertions.Analyzers`; `Rules.cs` consumes the constants (single source of truth).
- All code fixers now use `DiagnosticIds.X` (consts are baked into the consuming IL at compile time) instead of `Rules.X.Id`; the two lazy `Rules.X.Title` usages replaced with literals. Both code fixer assemblies now carry **zero** references to the `Rules` type.
- IL-level regression tests (`System.Reflection.Metadata`) assert the built code fixer dlls contain no `TypeReference` to `Rules` — any future `Rules.X.Id` reintroduction fails the test.

No behavior change: same diagnostic IDs, same code-fix titles (copied from resx).

## Testing

- `TUnit.Analyzers.Tests`: 705 passed, 0 failed, 1 skipped (pre-existing skip)
- `TUnit.Assertions.Analyzers.CodeFixers.Tests`: 19 passed, 0 failed
- All three Roslyn variant projects (4.4 / 4.7 / 4.14) build clean with the source-linked `DiagnosticIds.cs`

## User-facing note

Existing affected sessions are environmental (stale assembly already loaded): restarting VS after updating TUnit clears it. This change makes future version-skew sessions benign.

### DIFF
--- a/SharedTestHelpers/RulesDecouplingVerifier.cs
+++ b/SharedTestHelpers/RulesDecouplingVerifier.cs
@@ -1,0 +1,61 @@
+using System.Collections.Generic;
+using System.IO;
+using System.Reflection;
+using System.Reflection.Metadata;
+using System.Reflection.PortableExecutable;
+
+namespace TUnit.Tests.Shared;
+
+/// <summary>
+/// Verifies that a code fixer assembly carries no IL reference to its analyzer project's
+/// <c>Rules</c> type. Guards against https://github.com/thomhurst/TUnit/issues/6157.
+/// </summary>
+/// <remarks>
+/// Code fixer assemblies ship in the version-agnostic <c>analyzers/dotnet/cs</c> folder while the
+/// analyzer assemblies ship per-Roslyn (<c>analyzers/dotnet/roslyn4.x/cs</c>), and the dependency
+/// resolves at runtime by simple name. Visual Studio cannot unload analyzer assemblies, so after a
+/// package update (or with mixed TUnit versions in one VS session) a new code fixer can bind
+/// against a stale analyzer assembly. Any IL reference to the <c>Rules</c> type — e.g.
+/// <c>Rules.X.Id</c> inside the eagerly-evaluated <c>FixableDiagnosticIds</c> — then throws
+/// <see cref="System.MissingFieldException"/> for rules the stale assembly doesn't have. Code
+/// fixers must use the compile-time-baked <c>DiagnosticIds</c> constants instead, which this
+/// helper enforces at the IL level: a <c>TypeReference</c> to <c>Rules</c> appears for any usage
+/// (field access, <c>typeof</c>, method call), so an empty result proves full decoupling.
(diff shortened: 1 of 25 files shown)


thomhurst:
Re follow-up review: confirmed the `TUnit0062` collision (`CancellationTokenMustBeLastParameter` vs `FactoryAccessedTooEarly`) is real and pre-existing — both `Rules.cs` files shipped the literal before this PR. Not renumbering here: published IDs, `.editorconfig` severities, and resx resource keys (`TUnit0062Title` etc.) all depend on the ID, so that is a user-facing change deserving its own PR. Added a NOTE comment at the AspNetCore constant documenting the collision and the shared-severity implication for projects referencing both analyzers.

PR #6176, opened 2026-06-07
fix(mocks): WasCalled/WasNeverCalled assertions via [GenerateAssertion] for all ICallVerification types

thomhurst:
## Summary
- Replace hand-written `WasCalledAssertion`/`WasNeverCalledAssertion` classes and `IAssertionSource<T>` extension overloads with generic `[GenerateAssertion]` methods constrained to `ICallVerification`
- Fixes `await Assert.That(mock.Method(Any())).WasCalled(Times.Once)` not compiling for generated mock call types (`MockMethodCall<T>`, `PropertyMockCall<T>` struct, `VoidMockMethodCall`, `PropertySetterMockCall<T>`)
- `WasCalled(Times)` and `WasNeverCalled` remain direct extension methods so the non-assertion form `mock.Add(1, 2).WasCalled(Times.Once)` keeps working

## Test coverage
New `MockAssertionExtensionsTests` (32 tests, all through the assertion pipeline):
- **Positive:** `WasCalled()` shorthand, all `Times` variants incl. boundaries (Once/Exactly/AtLeast/AtMost/Between/Never), `WasNeverCalled()`, arg-specific counting, `Any()` matcher, void methods, property getters (struct `PropertyMockCall<T>` — exercises the generic constraint path), property setters via `.Setter` and `.Set(value)`
- **Negative:** each `Times` variant with wrong count asserts `AssertionException` with `Mock verification failed` / `called N time(s)` message, `WasNeverCalled` after call, non-matching args, null `ICallVerification` target

Also cleaned up `VerificationTests.Greet_Returns_Configured_Value` (stale 'does not compile' comment removed; result now asserted).

## Test results
- `MockAssertionExtensionsTests`: 32/32 pass (net10.0)
- `VerificationTests`: 15/15 pass (net10.0)


thomhurst:
Addressed all three review items in 4df31cc:

- **Inner exception**: both catch blocks now call `AssertionResult.Failed(ex.Message, ex)`, matching the deleted assertion classes' behavior.
- **Nested `IGreeter`**: removed — verified the shadowing also affected the pre-existing `Verify_String_Method_On_Different_Interface` test; both tests now use the top-level `IGreeter`.
- **Expectation message**: `WasCalled()` now uses "to have been called at least once".

Ran `VerificationTests` + `MockAssertionExtensionsTests` locally: 47/47 passed in both source-gen and reflection modes.

PR #6187, opened 2026-06-07
feat(assertions): return typed value from IsAssignableTo<T> (#6184)

### LINKED ISSUE #6184: [Feature]: Make IsAssignableTo behave like IsTypeOf and return the typed value
### Problem Statement

IsTypeOf returnss the typed value
```csharp
IEnumerable<MyDto> returnedData = (await Assert.That(okResult.Value).IsTypeOf<IEnumerable<MyDto>>())!;
```

But the IsAssignableTo returns object instead of the typed value requiring 2 steps
```
await Assert.That(okResult.Value).IsAssignableTo<IEnumerable<MyDto>>();
IEnumerable<MyDto> returnedData = (IEnumerable<MyDto>)okResult.Value!;
```

### Proposed Solution

return the typed value casted to the specified type.

Would also be good if in general would be possible to handle the nullability without requiring `(await Assertion)!` like the `IsNotNull`

### Feature Category

Assertion

QoL

thomhurst:
## Summary

Closes #6184.

`IsTypeOf<T>()` returns the value cast to `T`, but `IsAssignableTo<T>()` previously returned the original (untyped) value, forcing a second manual cast:

```csharp
// Before
await Assert.That(okResult.Value).IsAssignableTo<IEnumerable<MyDto>>();
IEnumerable<MyDto> data = (IEnumerable<MyDto>)okResult.Value!;

// After
IEnumerable<MyDto> data = (await Assert.That(okResult.Value).IsAssignableTo<IEnumerable<MyDto>>())!;
```

`IsAssignableTo<T>()` now returns the value cast to `T`, matching `IsTypeOf<T>()`.

## Implementation

`IsAssignableToAssertion<TTarget, TValue>` now derives from `Assertion<TTarget>` (was `Assertion<TValue>`) and maps the context to the target type via `context.Map<TTarget>(value => value is TTarget c ? c : default)`, so awaiting it yields `TTarget?`.

Validation still runs against the retained **original** context, so the existing behaviour is fully preserved:
- dual-mode checks (value type vs. thrown-exception type, e.g. `.Throws<X>().And.IsAssignableTo<Y>()`),
- null handling,
- error messages (`"type X is not assignable to Y"`).

Both the mapped and original contexts share the same cached underlying evaluation, so the source is still evaluated exactly once. No extension-method signatures changed — the awaited type simply flips from `TValue?` to `TTarget?`.

## Scope

`IsAssignableTo` only. `IsAssignableFrom<TSource>` and the negated variants (`IsNotAssignableTo`/`IsNotAssignableFrom`/`IsNotTypeOf`) are intentionally unchanged — they have no meaningful typed return.

## Tests

- New `TUnit.Assertions.Tests/IsAssignableToTypedReturnTests.cs`: cast-to-interface (the issue scenario), cast-to-base, exact type, plus not-assignable and null failure cases.
- Full assertions suite green (2130 tests), including `TypeAssertionAmbiguityTests`, `TypeOfTests`, and `Throws*` regression suites.

### DIFF
--- a/TUnit.Assertions.Tests/IsAssignableToTypedReturnTests.cs
+++ b/TUnit.Assertions.Tests/IsAssignableToTypedReturnTests.cs
@@ -1,0 +1,69 @@
+using System.Text;
+
+namespace TUnit.Assertions.Tests;
+
+/// <summary>
+/// Tests for GitHub issue #6184: <c>IsAssignableTo&lt;T&gt;()</c> should return the value
+/// cast to <c>T</c> (parity with <c>IsTypeOf&lt;T&gt;()</c>), so the caller doesn't need a
+/// second manual cast.
+/// </summary>
+public class IsAssignableToTypedReturnTests
+{
+    private interface IElement { }
+    private class Element : IElement { }
+    private class DerivedElement : Element { }
+
+    [Test]
+    public async Task IsAssignableTo_ReturnsValueCastToInterface()
+    {
+        // The exact scenario from the issue: an object/base-typed value that we want back
+        // as the interface it implements.
+        object obj = new List<string> { "a", "b", "c" };
+
+        IEnumerable<string> result = await Assert.That(obj).IsAssignableTo<IEnumerable<string>>();
+
+        await Assert.That(result.Count()).IsEqualTo(3);
+        await Assert.That(result.First()).IsEqualTo("a");
+    }
+
+    [Test]
+    public async Task IsAssignableTo_ReturnsValueCastToBaseType()
+    {
+        DerivedElement derived = new DerivedElement();
+
+        Element result = (await Assert.That(derived).IsAssignableTo<Element>())!;
+
+        await Assert.That(result).IsSameReferenceAs(derived);
+    }
+
+    [Test]
+    public async Task IsAssignableTo_ExactType_ReturnsValue()
+    {
+        var sb = new StringBuilder("Test");
+        object obj = sb;
+
+        StringBuilder result = (await Assert.That(obj).IsAssignableTo<StringBuilder>())!;
+
+        await Assert.That(result.ToString()).IsEqualTo("Test");
+    }
+
+    [Test]
+    public async Task IsAssignableTo_NotAssignable_StillFails()
+    {
+        await Assert.ThrowsAsync<TUnit.Assertions.Exceptions.AssertionException>(async () =>
+        {
+            Element element = new Element();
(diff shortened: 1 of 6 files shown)


thomhurst:
Thanks for the thorough review. Addressed in 7f1c3d7:

**1. Missing Throws-chaining test coverage** — Good catch, added two direct tests plus a single-evaluation guard:
- `IsAssignableTo_ChainedAfterThrows_ChecksAndReturnsExceptionType` — `…Throws<InvalidOperationException>().And.IsAssignableTo<Exception>()`. In the Throws chain the asserted context value *is* the thrown exception (`ThrowsAssertion<TException> : Assertion<TException>`), so the map casts it and the awaited result is the exception typed as `Exception` (the test asserts `result!.Message == "boom"`). The exception branch of `CheckAsync` is now exercised directly.
- `IsAssignableTo_ChainedAfterThrows_Fails_WhenExceptionNotAssignable` — covers the failing exception path.
- `IsAssignableTo_EvaluatesSourceOnlyOnce` — counts delegate invocations to lock the caching guarantee (see point 2).

(Note: the API is `Throws<T>()`, not `ThrowsAsync<T>()`.)

**2. Double-evaluation / coupling to caching** — Now covered by `IsAssignableTo_EvaluatesSourceOnlyOnce`, which fails if the source delegate runs more than once, so a future re-evaluation change can't silently regress it.

I'd push back on the suggested closure alternative (capturing the cast result into a field from inside the `Map` lambda): the map runs lazily during evaluation, so mutating assertion state from it is a worse coupling than reading the original context, and it still wouldn't supply the original *type* needed for the "type X is not assignable to Y" message on a failed cast (the mapped value is `null` there). `_sourceContext` is required specifically for that message and is genuinely needed, not just convenient.

**3. `_passedTask` vs `Passed`** — Intentional, not changing it. The sync sibling overrides return `Task<AssertionResult>` directly, so `_passedTask` (a cached, pre-allocated `Task`) avoids a `Task.FromResult(...)` allocation on every call. `AssertionResult.Passed` is correct only because my overrides became `async`. Converting the sync siblings would *add* allocations, so the current mix is the efficient choice rather than an inconsistency to clean up.

thomhurst:
Addressed the remaining nit in 897671f — renamed the unused `CheckAsync` parameter to `_` with a short comment explaining the intentional bypass (validation runs against the original context for the failure message). Build green, all 8 tests still passing.

PR #6195, opened 2026-06-08
fix(engine): run lifecycle hooks before test class construction (#6192)

### LINKED ISSUE #6192: [Bug]: Test class constructors are called before 'BeforeAssembly' and 'BeforeClass' events
### Description

It is my understanding, that the documentation states, that lifecycle events are called before test class instantiation.

https://tunit.dev/docs/writing-tests/lifecycle

### Expected Behavior

1) TUnitTestHooks.BeforeEveryAssembly is invoked
2) TUnitTestHooks.BeforeEveryClass is invoked
3) The MyTests constructor is invoked twice
4) MyTests.TestB is invoked
5) MyTests.TestA is invoked
6) TUnitTestHooks.AfterEveryClass is invoked
7) TUnitTestHooks.AfterEveryAssembly is invoked

### Actual Behavior

1) The MyTests constructor is invoked twice
2) TUnitTestHooks.BeforeEveryAssembly is invoked
3) TUnitTestHooks.BeforeEveryClass is invoked
4) MyTests.TestB is invoked
5) MyTests.TestA is invoked
6) TUnitTestHooks.AfterEveryClass is invoked
7) TUnitTestHooks.AfterEveryAssembly is invoked

### Steps to Reproduce

```
namespace YourTestProjectNameHere;

internal static class TUnitTestHooks
{
    [BeforeEvery(Assembly)]
    public static Task BeforeEveryAssembly(AssemblyHookContext context)
    {
        return Task.CompletedTask;
    }

    [BeforeEvery(Class)]
    public static Task BeforeEveryClass(ClassHookContext context)
    {
        return Task.CompletedTask;
    }

    [AfterEvery(Assembly)]
    public static Task AfterEveryAssembly(AssemblyHookContext context)
    {
        return Task.CompletedTask;
    }

    [AfterEvery(Class)]
    public static Task AfterEveryClass(ClassHookContext context)
    {
        return Task.CompletedTask;
    }
}

public class MyTests
{
    public MyTests()
    {
    }

    [Test]
    public Task TestA()
    {
        return Task.CompletedTask;
    }

    [Test]
    public Task TestB()
    {
        return Task.CompletedTask;
    }
}
```

### TUnit Version

1.51.0

### .NET Version

net10.0

### Operating System

Windows

### IDE / Test Runner

Visual Studio

```shell

```

### Additional Context

_No response_

### IDE-Specific Issue?

thomhurst:
Fixes #6192.

## Problem

The documented lifecycle contract says hooks run **before** test-class instantiation, but the constructor was running **before** `BeforeEveryAssembly`/`BeforeEveryClass` (and `Before(Assembly)`/`Before(Class)`).

**Root cause:** `TestCoordinator` called `CreateInstanceAsync()` (the constructor) *before* `TestExecutor.ExecuteAsync()`, and the before-hooks live inside `ExecuteAsync` (`GetOrCreateBeforeAssemblyTask`/`GetOrCreateBeforeClassTask`). So construction always preceded the hooks. The repro has no data sources, so this is purely execution-time ordering — the "constructor twice" in the report is just the two tests, each constructed once before the hooks.

## Fix

The before-hook tasks are already cached/idempotent in `BeforeHookTaskCache`, so they can be awaited earlier without re-running. New `TestExecutor.EnsureClassAndAssemblyHooksExecutedAsync` runs the cached `Before(TestSession/Assembly/Class)` tasks, and `TestCoordinator` calls it immediately before `CreateInstanceAsync` on both the no-retry fast path and the retry path. `ExecuteAsync`'s subsequent hook awaits become no-ops; the non-cached `FirstTestIn*` event receivers and `RegisterAfter*` pairs stay exactly where they were.

The helper replays the `RestoreExecutionContext` chain between hook awaits (mirroring `ExecuteAsync`) so AsyncLocals still flow `BeforeTestSession` → `BeforeAssembly` → `BeforeClass`.

## Tests

- New self-validating regression test `Bugs/6192/Bug6192Tests.cs`: the constructor throws if `BeforeEvery(Assembly)`/`BeforeEvery(Class)` have not yet run. Fails before the fix, passes after.
- Verified no regression across the lifecycle/hook suites (`Bugs/1914` AsyncLocal flow, `BeforeTests/*`, `Bugs/4432` constructor injection, `Bugs/_2804` hook cleanup), in **both** source-generated and reflection modes.

Engine-only change — no source-generator or snapshot changes.

### DIFF
--- a/TUnit.Engine/Services/TestExecution/TestCoordinator.cs
+++ b/TUnit.Engine/Services/TestExecution/TestCoordinator.cs
@@ -133,6 +133,11 @@ internal sealed class TestCoordinator : ITestCoordinator
                     // constructor (#6192). Hook tasks are cached, so ExecuteAsync's later awaits are no-ops.
                     await _testExecutor.EnsureClassAndAssemblyHooksExecutedAsync(test, cancellationToken).ConfigureAwait(false);
 
+                    // Flow AsyncLocals captured by Before(Assembly)/Before(Class) hooks into the
+                    // constructor. Must run here (not inside the gate above) so the restored ambient
+                    // ExecutionContext is still in effect when CreateInstanceAsync runs the constructor.
+                    test.Context.ClassContext.RestoreExecutionContext();
+
                     test.Context.Metadata.TestDetails.ClassInstance = await test.CreateInstanceAsync().ConfigureAwait(false);
 
                     // Drop the cached eligible-objects list so any later consumer rebuilds it with the new ClassInstance included — the initial list was built before the instance existed.
@@ -358,6 +363,11 @@ internal sealed class TestCoordinator : ITestCoordinator
         // Run Before(TestSession/Assembly/Class) hooks (incl. BeforeEvery) before constructing the
         // instance so hooks precede the constructor (#6192). Cached, so ExecuteAsync re-awaits are no-ops.
         await _testExecutor.EnsureClassAndAssemblyHooksExecutedAsync(test, cancellationToken).ConfigureAwait(false);
+
+        // Flow AsyncLocals captured by Before(Assembly)/Before(Class) hooks into the constructor.
+        // Must run here (not inside the gate above) so the restored ambient ExecutionContext is still
+        // in effect when CreateInstanceAsync runs the constructor.
+        test.Context.ClassContext.RestoreExecutionContext();
 
         test.Context.Metadata.TestDetails.ClassInstance = await test.CreateInstanceAsync().ConfigureAwait(false);
 
--- a/TUnit.Engine/Services/TestExecution/TestCoordinator.cs
+++ b/TUnit.Engine/Services/TestExecution/TestCoordinator.cs
@@ -128,6 +128,11 @@ internal sealed class TestCoordinator : ITestCoordinator
                 }
                 else
                 {
+                    // Run Before(TestSession/Assembly/Class) hooks (incl. BeforeEvery) before constructing
+                    // the instance so the documented lifecycle contract holds — hooks precede the
+                    // constructor (#6192). Hook tasks are cached, so ExecuteAsync's later awaits are no-ops.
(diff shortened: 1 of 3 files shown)


thomhurst:
Addressed the review feedback in c12d654:

**Missing `ClassContext.RestoreExecutionContext()` (real bug, fixed).** Confirmed: the gate ran the Before(Class) hook but the constructor then ran without the class ExecutionContext restored, so AsyncLocals a class-level hook opts into via `AddAsyncLocalValues()` reached the test body (`ExecuteAsync` line 157) but **not** the constructor.

Restoring at the end of `EnsureClassAndAssemblyHooksExecutedAsync` does **not** work — the ambient ExecutionContext is reset when that async method returns to its caller, so the restore is lost before `CreateInstanceAsync` runs. I verified this: a regression test asserting the AsyncLocal in the constructor still failed with the restore inside the gate method, and passed once moved. The restore now lives in `TestCoordinator` immediately before `CreateInstanceAsync`, on both the fast path and the retry path.

**Test coverage gaps (fixed).** `Bug6192Tests` now also covers `[Before(Class)]`/`[Before(Assembly)]` (not just `BeforeEvery`), and a new `Bug6192RetryTests` covers the retry path. Both assert AsyncLocal flow into the constructor, which is what guards the fix above.

**`Task` vs `ValueTask` (fixed).** `EnsureTestSessionHooksExecutedAsync` now returns `ValueTask`.

**`volatile` probe fields (fixed).** The static bools are now `volatile`.

**After-hook registration coupling (no change).** The `RegisterAfterTestSessionHook` omission is intentional and already documented in the XML comment on `EnsureClassAndAssemblyHooksExecutedAsync`. The hypothetical regression requires a new skip path inserted between the gate and `ExecuteAsync` that doesn't exist today; adding a guard for it now is speculative. Left as-is.

**Comment on the duplicate `GetOrCreateBeforeTestSessionTask` call site (no change).** The method's XML doc already states the matching awaits in `ExecuteAsync` are no-ops because the tasks are cached. An extra inline comment would be redundant.

PR #6196, opened 2026-06-08
feat(assertions): inference-friendly pinned overload for covariant [AssertionExtension] with own generic (#5922)

thomhurst:
## Summary

Resolves the call-site ergonomics issue in #5922 (the follow-up to the #5918 emit fix).

When an `[AssertionExtension]` class declares its **own** generic parameter **and** derives from `Assertion<TConcrete>` for a non-sealed concrete (covariance-candidate) receiver, the generator emitted only a covariant overload:

```csharp
public static MatchesAssertion<T> Matches<TActual, T>(this IAssertionSource<TActual> source, Func<T, bool> predicate, ...)
    where TActual : System.Exception
```

Because `IAssertionSource<T>` is **invariant**, the covariant `TActual` is genuinely required so `Assert.That(subclass)` can bind. But C# forbids partial type-argument specification, so a caller naming the class's own (non-inferable) argument had to repeat the receiver type:

```csharp
Assert.That(ex).Matches<Exception, MyPayload>(p => p.IsValid);   // verbose
```

## Change

The generator now **additionally** emits an inference-friendly *pinned-receiver* overload — but only for this exact shape (covariance-candidate receiver **and** own generic params):

```csharp
public static MatchesAssertion<T> Matches<T>(this IAssertionSource<System.Exception> source, Func<T, bool> predicate, ...)
```

So the common exact-receiver call site only needs the class's own type argument, while the covariant overload is kept for subclass receivers:

```csharp
Assert.That(ex).Matches<MyPayload>(p => p.IsValid);              // exact type — clean
Assert.That(argEx).Matches<ArgumentException, MyPayload>(...);   // subclass — still explicit
```

No `OverloadResolutionPriority` is needed: C#'s "more specific parameter types" rule prefers the concrete-receiver overload when both apply (the same way `M(int)` beats `M<T>(T)`). Dropping the attribute also avoids a polyfill dependency in the generated code.

`MethodAssertionGenerator` is intentionally left unchanged — it already suppresses covariance for own-generic methods (documented there, citing CS1929).

## Files

- `AssertionExtensionGenerator.cs` — add `pinnedReceiver` path; emit the extra overload only for the covariance-candidate + own-generics shape (via a small local `EmitMethod` helper).
- `ConcreteReceiverWithExtraGeneric.*.verified.txt` — updated snapshots (all TFMs); strengthened the generator test to pin both overload signatures alongside the compile-clean gate.
- `CovariantGenericExtensionInferenceTests.cs` (new) — compile-time guard proving overload resolution: pinned binds for exact type, covariant binds for subclass, and no CS0121 ambiguity when both apply.

## Verification

- Generator tests: **267/267**
- PublicAPI snapshots: **18/18** (no shipping assertion surface changed)
- Inference-proof runtime tests: **3/3** on net8.0 + net10.0


thomhurst:
Pushed a follow-up (`04954d2`) that **narrows when the pinned overload is emitted**, after a fair challenge that it rarely helps in practice.

### The problem with always emitting it
The pinned overload only earns its place when the caller is *forced* to name a type argument — i.e. when an own generic parameter is **not inferable** from the constructor's value arguments (it appears only inside a `Func<T, bool>` / `Action<T>` / `Expression<...>`, where a lambda can't fix it). When every own generic **is** inferable (e.g. a plain `T tag`), the caller writes no type arguments at all, the covariant overload binds on its own, and the pinned overload is pure dead weight — extra API/IntelliSense surface for zero benefit.

### Change
Emission is now gated on `CovarianceHelper.OwnGenericsAreInferable(constructor, ownTypeParameters)`:
- **non-inferable** own generic (e.g. `Func<T,bool>`) → pinned overload emitted (the real ergonomic win).
- **inferable** own generic (e.g. `T tag`) → pinned overload skipped.
- Detection is conservative: anything not provably inferable is treated as non-inferable, so it keeps the prior behaviour rather than risk an ergonomic regression.

A nice side effect: with the inferable case excluded, the pinned (`<T...>`) and covariant (`<TActual, T...>`) overloads are **arity-disjoint and never both applicable**, so the "more specific receiver" tiebreaker is no longer exercised and no `OverloadResolutionPriority` is involved — which moots review concern #2.

### Review feedback addressed
- **#1 (closure capture):** `EmitMethod` is now an explicit `private static` method; dependencies are passed in rather than captured.
- **#2 (ORP comment):** moot — overloads are arity-disjoint, documented in the comment block.
- **#3 (nullable path):** confirmed non-issue — `GenerateExtensionMethod` is never called with `isNullableOverload: true` in this generator, so pinned + nullable can't co-occur.
- **#4 (always-passing test predicate):** added inline comments on the compile-guard fixtures.

### Tests
- New snapshot fixture `ConcreteReceiverWithInferableGenericAssertion` (`T tag`) proves the pinned overload is **not** emitted; existing `ConcreteReceiverWithExtraGeneric` (`Func<T,bool>`) snapshots are unchanged.
- Generator tests **52/52**, compile-guard inference tests **9/9** (net8/9/10), PublicAPI **18/18**.

thomhurst:
Thanks for the second pass. Went through all four findings (commit `94d3737`):

**#2 — ORP comment near `effectivePriority`** → ✅ Applied. Added a comment at the `effectivePriority` calculation explaining that the pinned and covariant overloads intentionally share the same priority, since they're arity-disjoint and the "more specific receiver" rule is the tiebreaker (not ORP).

**#3 — delegate-inside-collection conservatism** → ✅ Applied. Documented the known conservatism in `AppearsInInferablePosition`'s summary: a `T` reachable only via `IEnumerable<Func<T,bool>>` is reported non-inferable even though the compiler could infer it; the only consequence is one redundant overload (still resolved by the more-specific-receiver rule), so it shouldn't be "fixed" into an over-aggressive heuristic.

**#1 — `IsGenericType && TypeParameters.Length > 0` "always-true"** → Pushing back: it's not redundant. `INamedTypeSymbol.IsGenericType` walks containing types — it returns `true` if the type *or any containing type* has type parameters. So a non-generic assertion nested in a generic outer class (`class Outer<T> { class Inner : Assertion<SomeConcrete> }`) has `IsGenericType == true` but `TypeParameters.Length == 0`. The `&& TypeParameters.Length > 0` is what restricts `hasOwnGenerics` to the type's *own* type parameters. Removing it would set `hasOwnGenerics` true for that nested case with an empty `ownTypeParameters`. Leaving as-is.

**#4 — no negated-pinned compile-guard test** → Pushing back: the negated path is the *same* emit code — `negated` only flips the method name (`methodName = negated ? data.NegatedMethodName : data.MethodName`); the generic-parameter handling, `pinnedReceiver` logic, and `emitPinned` value are identical for the positive and negated calls. So a negated call-site guard would exercise the exact same overload-resolution machinery the existing positive `Pinned_Overload_Exact_Receiver_Only_Needs_Own_Type_Argument` test already covers — it'd be duplicate coverage rather than new protection. Happy to add one if you'd still prefer the explicit belt-and-braces.

Also noting your correction on prior item #1 (the `EmitMethod` closure) — yep, it's now an explicit static method, so that's resolved.

PR #6197, opened 2026-06-08
feat: add DeferEnumeration to defer data-source expansion to runtime (#5833)

### LINKED ISSUE #5833: [Feature]: xunit's `DisableDiscoveryEnumeration`
### Problem Statement

It would be useful to have an option similar to xUnit’s `DisableDiscoveryEnumeration`, where test cases are not discovered during discovery but instead executed at runtime.

This could help reduce IDE overhead when there are thousands of test cases.

https://api.xunit.net/v3/2.0.0/v3.2.0.0-Xunit.TheoryAttribute.DisableDiscoveryEnumeration.html

### Proposed Solution

```cshap
public static IEnumerable<int> TestData() => Enumerable.Range(0, 10000);
[Test]
[MethodDataSource(nameof(TestData), DisableDiscoveryEnumeration = true)]
public async Task MyTest(int input)
{
    await Assert.That(input).IsBetween(0, int.MaxValue);
}
```

### Alternatives Considered

_No response_

### Feature Category

Data-Driven Testing (Arguments, DataSources)

### Additional Context

_No response_

### Contribution

thomhurst:
## What

Adds `DeferEnumeration` — TUnit's equivalent of xUnit's `DisableDiscoveryEnumeration` (closes #5833).

Setting `DeferEnumeration = true` on a data source tells TUnit **not** to enumerate it during discovery. The test appears as a **single placeholder node**, and the data rows are enumerated into individual cases only when the test is **run** (each reported nested under the placeholder). This avoids the IDE/test-explorer overhead of expanding a data source that produces thousands of cases.

```csharp
public static IEnumerable<int> ManyCases() => Enumerable.Range(0, 10_000);

[Test]
[MethodDataSource(nameof(ManyCases), DeferEnumeration = true)]
public async Task MyTest(int input)
{
    await Assert.That(input).IsGreaterThanOrEqualTo(0);
}
```

With the flag set, discovery shows **one** node for `MyTest` instead of 10,000; running it expands and reports each case.

## How

Deferral only suppresses *discovery* enumeration — at run time, enumerating is fine. So:

- **Flag**: `bool DeferEnumeration` added to `IDataSourceAttribute` (honored by all data sources). `[Arguments]` and the internal empty source hardcode it to `false` (a single row — nothing to defer).
- **Discovery**: `TestBuilder` emits one `DeferredEnumerationExecutableTest` placeholder when any data source on the test defers (gated by a new `TestBuildingContext.IgnoreDeferral`). Built in both the discovery and execution builds so it matches the IDE's UID filter.
- **Execution**: `TestSessionCoordinator.ExecuteTests` calls a new `DeferredTestExpander` **before** counters/scheduling. It re-runs the builder with `IgnoreDeferral` to produce the real cases, registers them, reports the placeholder, and adds the children to the scheduled list — so the children flow through the **normal pipeline** with correct hooks and lifecycle counting. The placeholder itself never runs as a test (its `Create`/`Invoke` throw as a safety guard).
- **Nesting**: `TUnitMessageBus` now surfaces `TestContext.ParentTestId` as the MTP `parentTestNodeUid`, so the rows nest under the placeholder. (This also improves nesting for existing runtime test variants.)
- **Source generator**: threads the `DeferEnumeration` named argument into the emitted `MethodDataSource` factory.

Works in **both** source-gen and reflection modes; no new reflection (AOT-safe).

## Tradeoffs / limitations (documented)

- Individual rows can't be selected/filtered from the IDE or targeted by another test's `[DependsOn]` — they don't exist until runtime.
- The placeholder is reported once as a passed container (counts as +1 per deferred test). If the data source throws while enumerating, the error surfaces as a failed result at run time instead of failing discovery for the whole assembly.

⚠️ **Breaking-ish**: `IDataSourceAttribute` gains a non-default `DeferEnumeration` member, so anyone implementing the interface directly must add it. This mirrors the existing `SkipIfEmpty` member (a default interface member isn't safe given net472 support).

## Verification

Verified locally in both source-gen and reflection modes:

- Discovery collapses deferred tests to 1 node each (including a `[Repeat(2)]` variant).
- Run all → full expansion; run a single deferred test via filter → expands all its children (filter-bypass works).
- Throwing data source → surfaces as a failed result, no discovery crash.

Tests/snapshots added/updated:

- Samples in `TUnit.TestProject/DeferEnumerationTests/`
- Engine tests `TUnit.Engine.Tests/DeferEnumerationTests.cs` (dual-mode; AOT runs in CI)
- Source-gen snapshot `DeferEnumerationTests.Test.verified.txt`
- PublicAPI snapshots regenerated (all 4 TFMs)
- Docs: `docs/docs/writing-tests/defer-enumeration.md` + overview table + sidebar

Existing `TUnit.UnitTests` (219), source-gen snapshot tests (118), and variant engine tests pass unchanged.


thomhurst:
Thanks for the review — addressed in 43443ff.

**Issue 2 (count inflation + misleading green placeholder) — fixed.** The placeholder is now treated as a true container: on successful expansion it is no longer reported as its own result, so it doesn't count toward totals and can't show green while its children fail. The children carry the real results (nested via `ParentTestId`). A result is reported only if the *expansion itself* throws, so that failure stays visible. (Per-row data errors still surface as their own failed child via the standard data-generation-error path.) Engine tests now assert exactly the case count (10 / 30), and the docs note explains the container semantics. Verified in both modes: `Deferred` → 10, all → 40, throwing source → `total 1, failed 1` (just the failed child, no green placeholder).

**Observation (FsCheck live setter) — fixed.** `FsCheckPropertyAttribute.DeferEnumeration` is now hardcoded `false` (`get => false; set {}`) like `[Arguments]`, since FsCheck has no discovery-time rows to defer.

**Issue 1 (read-only property) — partially declined, with a fix.** A read-only `bool DeferEnumeration { get; }` isn't viable: attribute named-argument syntax (`[MethodDataSource(..., DeferEnumeration = true)]`) requires the property to be settable, and the member is shared via the interface — so making it get-only would make the feature uncompilable on the attributes that legitimately use it. Throwing in the setter is also risky: the source generator round-trips named args into generated attribute initializers, so a throw would run at module/attribute construction and could break discovery. So I kept the no-op `false` on the single-row sources (intentional per the design — there's genuinely nothing to defer for one row) and documented it on the interface instead. An analyzer warning for `DeferEnumeration = true` on a single-row source is a reasonable future addition if this proves to be a footgun in practice.

**Issue 3 (`_Deferred` suffix collision) — kept, with rationale.** This matches the existing `GenerateFailedTestId` convention (`_DataGenerationError` suffix), which carries the identical theoretical risk. A real collision needs a contrived clash — a method literally named `Foo_Deferred(int x)` *and* a non-deferred `Foo(int x)` with matching parameters. Diverging only this one suffix to a non-identifier character would make it inconsistent with the sibling method; I'd rather change both together (or neither) in a follow-up if this is a concern.

thomhurst:
Thanks — both minor notes addressed in 52fd97f (no behavior change; counts still 40 passing / error case total 1, failed 1):

- **O(N·M) placeholder removal** → now a single `testList.RemoveAll(static t => t is DeferredEnumerationExecutableTest)` pass (O(N+M)) before the expansion loop.
- **Error-path message order** → `InProgress` is now emitted before `EndTime`/`SetResult`, so the sequence is `InProgress` → `Failed` with `Start ≤ End`.

`_Deferred` suffix left as-is per the agreed rationale (consistency with `_DataGenerationError`).

thomhurst:
Good catch — running the placeholder in the IDE showed *"Inconclusive: Test has not run"* because, after the previous fix, the placeholder node produced no result (its children were the only results). Fixed in ee04fed.

The placeholder is now reported as a **container with an aggregate result**:
- `InProgress` when expansion starts (so the node the user ran shows as running and the cases nest under it);
- after the cases execute, a final result that aggregates them — **passed** only if every case passes, **failed** if any case fails, **skipped** if all were skipped.

This resolves the "not run" state without the misleading-green problem (reporting it always-passed would hide child failures). The cost is that the placeholder counts as one extra entry in flat reporters (TRX/console) — a 10-row deferred source reports 11 (10 cases + 1 container) — which is now documented. Engine tests updated (11 / 31; error case = failed case + failed container = 2).

@thomhurst if you re-pull and rebuild, the placeholder node should now resolve to pass/fail in the IDE rather than "not run" — could you confirm it behaves as expected on your end? I can verify console/TRX counts here but can't drive Rider directly.

thomhurst:
Thanks — addressed in 30500cc.

**`IsForExecution: false` during expansion — verified correct, no change.** `IsForExecution` is referenced in exactly one place in the build path (`TestBuilder.cs:120`): `if (buildingContext.IsForExecution && buildingContext.Filter != null)` — the pre-filter optimisation. Since the expander passes `Filter: null`, that branch is skipped regardless of the flag's value, so `false` vs `true` is behaviorally identical here. It's not what gates event receivers — execution registration is done explicitly via `RegisterTestsAsync(isForExecution: true)`. `false` also matches the existing runtime-built-test precedent (`TestRegistry.ProcessPendingDynamicTests`, `TestBuilderPipeline.BuildTestsAsync`). Added a comment documenting this.

**Class-data deferral coverage — added.** New `DeferEnumerationClassDataTests` sample (`[MethodDataSource(..., DeferEnumeration = true)]` on the class) + engine test, exercising the `metadata.ClassDataSources` branch: discovery shows 1 node, run expands to 6 (5 instances + placeholder).

**Count inflation — documented prominently.** Now called out in the Trade-offs warning block of the docs, not just the info note.

**`_Deferred` collision — strengthened the proof, kept the suffix.** I dug into the exact id shapes and the collision can't actually occur: every real id from `GenerateTestId` is `...(ctorParams).{classIdx}.{classLoop}.Method(params).{methodIdx}.{methodLoop}.{repeat}` — the method name is always wrapped in numeric data-index segments. The placeholder id is `...Method(params)_Deferred` with none of those segments, so the shapes can never be equal. The scenario in the review (`MyTest_Deferred` vs `MyTest` deferred) produces `...MyTest_Deferred(int).i.j.k` vs `...MyTest(int)_Deferred` — different. I added this as a `<remarks>` proof rather than switching to a non-identifier delimiter, since a non-ASCII/control char in test ids risks breaking TRX serialization and treenode filters for a collision that can't happen. Happy to revisit if you'd still prefer the delimiter change.

PR #6230, opened 2026-06-12
feat(mocks): setup/verify on secondary interfaces of multi-type mocks

thomhurst:
Closes the gap reported in discussion #4981 (comment 17280965): `Mock.Of<DbContext, IInfrastructure<IServiceProvider>>()` had no way to configure the secondary interface, and casting the object to it executed the real getter.

## What users get

```csharp
var dbContext = Mock.Of<DbContext, IInfrastructure<IServiceProvider>>();
dbContext.Instance.Returns(serviceProvider);   // secondary member, directly on the mock
dbContext.Instance.WasCalled();

((IInfrastructure<IServiceProvider>)dbContext.Object).Instance; // => serviceProvider
```

- Secondary members surface as extensions on `Mock<T1>` — setup, verify, raise, `VerifyInOrder`, unified `Invocations`, strict mode and `Reset` all work through the existing single engine.
- Name collisions across interfaces get a short interface prefix (`mock.IBar_Tag`); signature-identical members dedupe onto the primary surface and intercept calls through every cast.
- Misuse throws: using a secondary extension on a mock not created with that interface raises `InvalidOperationException` with the correct `Mock.Of<...>` hint, instead of a silently-never-matching setup.
- Class primaries (the DbContext case) now work end-to-end: previously the multi factory was never registered (`Mock.Of<Class, IExtra>()` threw) and the impl never implemented the secondary interface. Explicit/non-virtual class implementations are intercepted via explicit re-implementation. Ctor args supported: `Mock.Of<T1, T2>(params object[])` up to T4.

## Design notes

The same `(T1, Tn)` pair can appear in multiple combos (`Mock.Of<T1,T2>()` and `Mock.Of<T1,T2,T3>()`), where the member gets **different** union member IDs. The extension surface is therefore generated once per pair (deduped pipeline model) with **local ordinals**, and each combo factory registers a `typeof(Tn) + localId -> unionId` map on the engine (`MockEngine.RegisterSecondaryInterface`/`TryGetSecondaryMemberId` — new uniquely-named methods, no overload changes). This keeps one engine (no split Invocations/Reset semantics) while making the shared surface correct for every combo layout.

The single- and multi-type member walks were unified into one collector, making the primary-members-get-identical-IDs invariant structural instead of two hand-synced copies (which had already diverged for class primaries).

## Testing

- 17 new runtime tests (`MultipleInterfaceSecondarySetupTests`): secondary setup/verify/reset/strict, cross-combo pair reuse, dedup, collision prefixing, wrong-mock guard, class primary with explicit impl / non-virtual blocker / ctor args, generic secondary interface.
- 3 new generator snapshot tests (interface-primary, class-primary with explicit impl, name collision).
- Full suites green: TUnit.Mocks.Tests 1106 (net10) / 1105 (net9) / 1090 (net8), generator snapshots 69/69, Mocks.Http 54, Mocks.Logging 31. Roslyn 4.14/4.4/4.7 generator variants build clean.

Docs: multi-interface section in `docs/writing-tests/mocking/setup.md` updated.


thomhurst:
Addressed the follow-up review in 32d921e8d:

**Setter-slot / unmapped-member error** — fixed via your Option B. `HasSecondaryInterface` is back (and now actually called by generated code): `__Id` distinguishes "mock lacks the interface" (keeps the `Mock.Of<T1, Tn>()` hint) from "interface registered but slot unmapped" (honest "no setup mapping on this mock instance" message). Worth noting the unmapped path is currently unreachable through generated extensions — signature-identical members (including the read-write-on-getter-only case, since the property key is name+type) are excluded from the pair surface and configured through the primary extension instead — but the split keeps the error honest if emission ever changes.

**Rename/signature-key comment** — added at the rename site in `BuildPairModel`: rename touches `Name` only; `ComputeMemberIdMap` correlates by the original signature key, so the map is unaffected.

**Collection expressions in generated code** — declining this one: generated code compiles inside consumer projects whose `LangVersion` we don't control, and `[2, 3]` requires C# 12. Every other emission site in the generator sticks to `new T[] { ... }` / `Array.Empty<T>()` for the same reason, so this also keeps the emitted style uniform.

PR #6286, opened 2026-06-19
perf(sourcegen): consolidate per-file ModuleInitializers into merged .cctor (#6226)

### LINKED ISSUE #6226: Startup JIT: consolidate remaining per-file ModuleInitializers into merged partial-class .cctor pattern
## Background

The test/hook registration path already minimizes startup JIT: per-file generated output contributes static field initializers to partial classes (`TUnit.Generated.TUnit_TestRegistration` / `TUnit_HookRegistration`), the compiler merges them into a single `.cctor` (one JIT-compiled method per assembly), triggered by the single per-assembly `[ModuleInitializer]` in `InfrastructureGenerator`. Registrations only store lazy factories, so heavy metadata construction is deferred to discovery (where it JITs in parallel).

Several generators still emit a **separate `[ModuleInitializer]` method per file/class**, each costing a type load + a serial JIT at module load:

- `TUnit.Core.SourceGenerator/Generators/PropertyInjectionSourceGenerator.cs` — 4 emit sites (~lines 589, 762, 800, 886). Worse: the `InitializerPropertyRegistry.Register` initializer constructs the full `InitializerPropertyInfo[]` (including delegates) **eagerly inline in module init**, not lazily.
- `TUnit.Core.SourceGenerator/Generators/AotConverterGenerator.cs:656`
- `TUnit.Core.SourceGenerator/CodeGenerators/StaticPropertyInitializationGenerator.cs:211`
- `TUnit.Core.SourceGenerator/CodeGenerators/DynamicTestsGenerator.cs:105` (minor — usually few dynamic sources)

## Proposal

Apply the existing consolidation pattern to these generators:

1. Emit per-file **static field initializers** on shared partial shell classes (e.g. `TUnit.Generated.TUnit_PropertyRegistration`), declared once by `InfrastructureGenerator` and triggered via `RuntimeHelpers.RunClassConstructor` from the existing single module initializer.
2. Register **lazy factories** (`Func<...>` with `static` lambdas) instead of constructing registration payloads eagerly — particularly the `InitializerPropertyInfo[]` arrays in PropertyInjectionSourceGenerator.

This is free aggregation via the C# compiler's partial-class merge — **no `.Collect()` fan-in is added to the incremental pipeline**, so incremental compilation behavior is unchanged.

## Expected impact

N per-file module-initializer methods (type load + serial tier-0 JIT each) collapse into one merged `.cctor` per concern. For large suites with many property-injected/initialized classes this removes a linear serial JIT cost from module load.

## Notes

- Keep generated registration bodies loop-free (loops in module-init/.cctor paths can trigger OSR tier-1 compilation of large methods).
- Dual-mode rule applies: source-gen consolidation only; reflection mode unaffected.
- Snapshot tests will need re-verification.

thomhurst:
Fixes #6226.

## Problem
Several generators still emitted a **separate `[ModuleInitializer]` method per file/class**, each costing a type load + a serial tier-0 JIT at module load:
- `PropertyInjectionSourceGenerator` (4 emit sites; the InitializerProperties site also built the `InitializerPropertyInfo[]` array eagerly inline in module init)
- `AotConverterGenerator`
- `StaticPropertyInitializationGenerator`
- `DynamicTestsGenerator` (one ModuleInitializer per dynamic-test method)

## Fix
Apply the existing consolidation pattern (already used by test/hook registration via `TUnit_TestRegistration`/`TUnit_HookRegistration`):

1. `InfrastructureGenerator` declares **4 new empty partial shells** in `TUnit.Generated` and triggers them from its single module initializer via `RuntimeHelpers.RunClassConstructor`:
   `TUnit_PropertyRegistration`, `TUnit_ConverterRegistration`, `TUnit_StaticPropertyRegistration`, `TUnit_DynamicTestRegistration`.
2. Each generator now contributes `static readonly int _r_X = Registry.Register(...);` fields to the matching shell. The compiler merges all contributions into **one `.cctor` per concern** — N per-file ModuleInitializer methods collapse into 1 merged method.
3. Five registry `Register`/`RegisterDynamic` methods now return `int` (dummy `0`) so they can be used as static field initializers — matching the existing `RegisterEntries`/`RegisterHook` convention.

This is free aggregation via the compiler's partial-class merge — **no `.Collect()` fan-in is added** to the incremental pipeline, so incremental compilation behavior is unchanged.

## Eager vs. lazy
Registrations are emitted **eager-inline with no factory lambdas**. A `static () => …` factory lambda is a new compiler-generated method (= a new JIT call) per type, so it's only worth it when it defers genuinely heavy, parallelizable work. The four consolidated concerns here are cheap (`new Source()`, a small `InitializerPropertyInfo[]`, `new Converter()`, `new DynamicSource()`), so they go straight into the merged `.cctor`.

The **test/hook generators are intentionally left as-is**: they already consolidate into a merged `.cctor`, already batch per-class work into shared named methods (`__CreateInstance`/`__Invoke`/`__Attributes`), and keep a single per-class `() => Source.Entries` lazy factory — that one lambda is the deferral that pushes heavy `ClassMetadata`/`MethodMetadata`/`Entries` construction off the serial module-init path onto parallel discovery JIT.

## Dual-mode
Source-gen consolidation only. Registry **consumer** methods (`GetSource`, `GetProperties`, `GetRegisteredProperties`, `DiscoverInjectableProperties`, `TryConvert`) are byte-identical, and reflection mode never calls `Register`. Reflection mode is unaffected.

## Validation
- Builds clean: `TUnit.Core`, source generator, `TUnit.TestProject` (0 errors).
- **480** source-generator snapshot tests pass. Only Infrastructure + AotConverter output is snapshot-tested, so just 5 `.verified.txt` changed (4 `DuplicateTypeName…Infrastructure…` + the `[Skip]`'d `AotConverterGeneratorTests.GeneratesCode`).
- **4** `TUnit.PublicAPI` Core snapshots updated for the `void`→`int` surface change and pass on all TFMs.
- Runtime (SourceGenerated): property injection + static properties, 29 dynamic tests, AOT converters and basics all green — confirming each merged `.cctor` fires and registers correctly.


thomhurst:
Thanks for the review — addressed in 106a84276.

**Issue 1 (SafeName collision):** Fixed. `_r_` static-property fields now append a stable per-build hash of the fully-qualified type + property name, so types differing only in `.` vs `_` no longer collide in the merged `TUnit_StaticPropertyRegistration` `.cctor`. (Worth noting: the prior `registeredFields` guard would have *silently dropped* the second registration rather than erroring, so this is a real correctness fix, not just compile-safety.)

**Issue 2 (dynamic overloads):** Fixed, and a bit deeper than the field name — the generated **file name** `Dynamic_{class}_{method}.g.cs` collided too (it would throw at `AddSource` before the field ever mattered), as did two same-minimal-name classes in different namespaces. Both the file name and the field now carry a stable hash of the fully-qualified type + method + line, so overloaded/colliding dynamic sources now generate distinct files/fields and **work** instead of failing.

**Minor (void→int API delta):** Acknowledged — intentional, matches the existing `RegisterEntries`/`RegisterHook` convention (they already return `int` for exactly this static-field-initializer use), source-compatible for call sites, and tracked in the updated `TUnit.PublicAPI` Core snapshots.

Validation after the fix: generator builds (Debug+Release), TestProject compiles, 29 dynamic + static-property runtime tests green, full 480 source-gen snapshot suite green (static/dynamic output isn't snapshotted, so no snapshot churn from the hashes).

thomhurst:
Rebased onto latest `main` and addressed the remaining review findings:

- **PropertyInjectionSourceGenerator field-name collision (Finding A):** Confirmed already resolved on the branch (commit `85d1f7f5ae`) — all four merged-partial field names carry the deterministic FNV-1a `stableHash` suffix (`FileNameHelper.GetStableHashCode(...).ToString("x8")`), matching `StaticPropertyInitializationGenerator`. No collision risk remains.
- **AotConverterGenerator namespace (Finding B):** Converted the file-scoped `namespace TUnit.Generated;` to the explicit block form `namespace TUnit.Generated { ... }` so it matches the shell in `InfrastructureGenerator` and the other generators, wrapping both the converter classes and the `TUnit_ConverterRegistration` partial.

Validation: generator builds clean; source-gen snapshot suite 476 passed / 4 pre-existing skips / 0 failed (no `.received.txt` produced); TUnit.TestProject compiles.

thomhurst:
Addressed the new follow-up finding: `StaticPropertyInitializationGenerator` was deduping the generated *initializer method* by simple type name (`Initialize_{ContainingType.Name}_{Property.Name}`) while the registration fields used an FQN stable hash — so two same-simple-named types in different namespaces, each with a same-named static property, collapsed to one method (hardcoding the first type's FQN) and the second type's property was silently initialized against the wrong type.

Fix: method names now carry the same FQN `stableHash` suffix the field names use (hoisted into shared `GetStableHash`/`GetInitializerMethodName` helpers and threaded through `GenerateIndividualPropertyInitializer`), so method generation and the `InitializerAsync` reference are identical and FQN-unique. Getter/setter bodies already targeted each property's own FQN.

Validation: generator builds clean; source-gen snapshot suite 476 passed / 4 skipped / 0 failed (no `.received.txt`); TestProject compiles.

thomhurst:
Applied the cheap consolidation suggestion: moved the `SafeName` helper out of `StaticPropertyInitializationGenerator` into `FileNameHelper` (alongside `GetStableHashCode`) as a `public static`, and switched the generator to call it. Logic is byte-identical, so generated output is unchanged — snapshot suite still 476 passed / 4 skipped / 0 failed with no `.received.txt` produced.

Deliberately left `PropertyInjectionSourceGenerator.GetSafeClassName` alone — it has materially different logic (symbol-based, strips `global::`, removes spaces, chained replaces), so consolidating it would risk changing generated identifiers.

Deferring the other suggestion (adding snapshot tests for the PropertyInjection/StaticProperty generators) — it's a pre-existing coverage gap rather than something this PR introduced, so it's better as a separate follow-up than expanding this PR's scope.

PR #6293, opened 2026-06-21
feat(aspire): actionable diagnostics when resources fail to start

thomhurst:
## Why

When an Aspire resource fails to start under TUnit, the failure was hard to diagnose: the fixture only read `State.Text`, so a user got an opaque `"Resource 'x' failed to start"` / timeout plus raw logs — no exit code, no health-check reason, no clue *which* dependency was blocking, and the state-transition timeline went only to stderr and was lost.

This makes the failure itself answer **what's wrong and what to do next**, and routes the detail into the test report.

## What changed (all in `TUnit.Aspire.Core/AspireFixture.cs`)

- **Snapshot-driven diagnosis** — on failure/timeout, read the live `CustomResourceSnapshot` (`ExitCode`, `HealthReports`, state) and classify the cause: container-runtime-down, image-pull failure, port-in-use, OOM (exit 137), non-zero exit, health-check-failing, never-started, crashed-no-code. Each class maps to **one actionable hint**.
- **State distinction at timeout** — `"Running but not Healthy (health 'ready' Unhealthy: connection refused)"` vs `"still Starting (never reached Running)"` vs `"FailedToStart, exit code 137 (OOM)"`, derived from each resource's current snapshot.
- **WaitFor dependency chain** — turns the opaque `"Waiting"` into `"api waiting on 'migrations' (Exited, exit code 1)"` via `WaitAnnotation`.
- **Startup timeline** — the background monitor records state/health transitions and folds them into the diagnostics; its lifetime now spans **both** `StartAsync` and the resource-wait phase so health-wait hangs are covered.
- **Concise exception + attached artifact** — the exception message stays short (diagnosis + hints + pointer); the full timeline and **untruncated** per-resource logs are written to an attached `Artifact` (on the failing test, with `TestSessionContext` fallback), so it shows up in trx/HTML/IDE.

### Incidental fixes
- Add `Exited` to the fail-fast watch — a container that crash-exits cleanly never reports `FailedToStart`, so previously we waited the full timeout for an already-dead resource.
- Lock `LogProgress` — the background monitor, the main init thread, and multiple fixture instances all write the shared static stderr stream; without a lock, lines splice.
- Fix `LogLine` rendering — `$"  {line}"` emitted the noisy generated record `ToString()` (`LogLine { LineNumber = .. }`); now uses `.Content` (also in `WatchResourceLogs`).

## Notes
- No public API changes — every addition is `private`/`internal`. Both source-gen and reflection modes; Native-AOT safe (public property reads only, no member reflection).
- Every Aspire 13.4.6 API used was verified by reflection-probing the actual assembly. Key gotcha encoded in the design: `HealthStatus` is `null` unless state == `Running`, so diagnosis reads `HealthReports` directly.

## Testing
- **18 pure no-Docker unit tests** in `TUnit.Aspire.Tests/AspireDiagnosticsTests.cs` (classification, hints, descriptions, log-signature scan, timeline formatting) — all pass.
- Builds clean on **net8.0 / net9.0 / net10.0**.
- Docker-backed integration fixtures (bad image tag, exit-1, never-healthy) are a sensible follow-up; not included here.

### DIFF
--- a/TUnit.Aspire.Core/AspireFixture.cs
+++ b/TUnit.Aspire.Core/AspireFixture.cs
@@ -1,10 +1,12 @@
 using System.Collections.Concurrent;
+using System.Collections.Immutable;
 using System.Diagnostics;
 using System.Text;
 using Aspire.Hosting;
 using Aspire.Hosting.ApplicationModel;
 using Aspire.Hosting.Testing;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
 using Microsoft.Extensions.Hosting;
 using TUnit.Core;
 using TUnit.Core.Interfaces;
@@ -29,6 +31,11 @@ public class AspireFixture<TAppHost> : IAsyncInitializer, IAsyncDisposable
 {
     private DistributedApplication? _app;
     private OtlpReceiver? _otlpReceiver;
+
+    // Captured state-transition timeline (recorded by the background monitor), folded into the
+    // exception when startup fails so the failure tells the whole story in one place.
+    private readonly ConcurrentQueue<StateTransition> _timeline = new();
+    private Stopwatch? _timelineClock;
 
     /// <summary>
     /// The running Aspire distributed application.
@@ -81,7 +88,7 @@ public class AspireFixture<TAppHost> : IAsyncInitializer, IAsyncDisposable
                 {
                     foreach (var line in batch)
                     {
-                        testContext.Output.WriteLine($"[{resourceName}] {line}");
+                        testContext.Output.WriteLine($"[{resourceName}] {line.Content}");
                     }
                 }
             }
@@ -97,6 +104,7 @@ public class AspireFixture<TAppHost> : IAsyncInitializer, IAsyncDisposable
     // --- Configuration hooks (virtual) ---
 
     private static readonly Stream StdErr = Console.OpenStandardError();
+    private static readonly object StdErrGate = new();
 
     /// <summary>
     /// Logs progress messages during initialization. Override to route to a custom logger.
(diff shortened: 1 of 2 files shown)


thomhurst:
Thanks for the review — addressed in 6b36d0a. Verified each against the code; all were valid:

- **High — `Exited` false-positives:** switched the fail-fast watcher to the predicate `WaitForResourceAsync` overload; `Exited` now only counts as a failure with a **non-zero** exit code (`IsFailureState`), so a one-shot resource that exits cleanly (code 0) is no longer mis-reported.
- **Medium — sequential log collection:** logs for all problem resources are now gathered with `Task.WhenAll`; output order stays deterministic.
- **Medium — only first dependency shown:** `SummarizeDependencies` now skips already-ready deps and reports the first *unsatisfied* one (the full chain remains in the artifact via `AppendDependencyChain`).
- **Low — timeline cap dropped the startup story:** now keeps head **and** tail (first 80 + last 120), dropping only the middle churn.
- **Low — `FormatTimeline` hid single transitions:** guard relaxed to `== 0`; a lone immediate-failure transition now renders.
- **Low — `DateTime.Now`:** → `DateTime.UtcNow` for the artifact filename.
- **Low — bare `catch {}`:** now `catch (Exception ex)` and logs the artifact-write failure via `LogProgress` instead of swallowing it.
- **Nice-to-have — `HealthSignature`:** now a per-check `name:status` signature, so independent checks flipping in opposite directions still register and the timeline shows which check moved.

19 unit tests pass; builds on net8.0/net9.0/net10.0.

thomhurst:
Second review pass addressed in d21fab6 — all four were valid:

- **Issue 1 (filename collision):** added `-{Guid.NewGuid():N}` to the artifact filename, so same-`TAppHost` fixtures timing out in the same millisecond can't overwrite each other.
- **Issue 2 (double `TryGetCurrentState`):** the snapshot is now read once per resource in `BuildDiagnosticsAndAttachAsync` and the `ResourceEvent` is threaded through `DiagnoseResource` and `AppendResourceDiagnosticBlock` — no more re-querying live state mid-render.
- **Issue 3 (timeline read/drain race):** good catch — `BuildDiagnosticsAndAttachAsync` runs while the monitor is still recording (it's stopped in `InitializeAsync`'s `finally`). Switched `_timeline` from `ConcurrentQueue` to a `List` guarded by a lock; `RecordTransition` and `FormatTimeline` both take the lock, so a read can no longer observe a transient empty/partial queue mid-rebuild.
- **Issue 4 (untested trimming):** extracted the compaction into a pure `internal static TrimTimeline(list, head, tail)` and added two tests (400→200 keeping head+tail; under-cap left unchanged).

21 unit tests pass; builds on net8.0/net9.0/net10.0.

Issue #6276, opened 2026-06-18
[Feature]: Count() and ContainsKey() not available in net8 ?


vpenades:
### Problem Statement

TUnit 1.55.2

For backwards compatibility and LTS, we maintain tests for both net8 and net 10

I'm surprised to find that Count() and ContainsKey() are only available in net10 ? why?

### Proposed Solution

support Count() and ContainsKey() in net8

### Feature Category

Test Discovery / Attributes

### How important is this feature to you?

None

### Contribution

- [ ] I'm willing to submit a pull request for this feature

Issue #5720, opened 2026-04-24
[Feature]: IsEqualTo() & co with implicitly convertible wrappers


Timovzl:
### Problem Statement

I'm doing assertions on wrapper Value Objects, i.e. single-valued Value Objects wrapping a primitive. For example, we might have a `ProductCode` wrapping `string`. These have an _implicit_ conversion _to the primitive_, and an _explicit_ conversion _to the wrapper_.

Before 1.39.0, the following worked for string wrappers:

```cs
await Assert.That(updatedStockItem.ProductCode).IsEqualTo("Example");
```

As of 1.39.0, the `string` overload is gone, and a `ProductCode` is required. Since the conversion in that direction is (rightfully) explicit, we have to make it manually, or resolve `ProductCode.Value`.

I considered filing a bug, but realized that the wrapper scenario only worked with `string` wrappers. An `int` wrapper, for example, already did not benefit. As such, a feature request seems more suitable.

### Proposed Solution

```cs
await Assert.That(updatedStockItem.ProductCode).IsEqualTo("Example"); // T = ProductCode
await Assert.That(updatedStockItem.SomeWrappedNumber).IsEqualTo(1); // T = MyIntWrapper
```

Could such equality checks (and perhaps also other comparisons) be made flexible enough to work, for wrappers?

For example, if the `T` in `Assert.That(T)` has an implicit conversion operator to some type `X`, it would be fantastic if `IsEqualTo` and other comparison overloads were made available accepting an `X`. In that case, the left hand would be converted to `X` and the result compared to the right hand's `X`.

### Alternatives Considered

```cs
await Assert.That(updatedStockItem.ProductCode.Value).IsEqualTo("Example");
await Assert.That(updatedStockItem.ProductCode).IsEqualTo((ProductCode)"Example");
```

Since we type these _very_ often, the simplification has merit.

### Feature Category

Assertions

### How important is this feature to you?

Nice to have - would improve my experience

### Contribution

- [ ] I'm willing to submit a pull request for this feature

Timovzl:
@thomhurst I want to express my heartfelt compliments and thanks, for both TUnit and your willing and proactive expansion of it. My coworkers and I enjoy it immensely.

Timovzl:
[...]

PR #6313, opened 2026-06-24
fix: resolve CS0121 IsEqualTo ambiguity on .NET 8 SDK (#6296)

### LINKED ISSUE #6296: [Bug]: Compile error CS0121 on .NET 8
### Description

We compile our product in a .NET 8.0 environment and get `error CS0121: The call is ambiguous between the following methods or properties:` with the latest version of TUnit.

### Expected Behavior

No compilation error

### Actual Behavior

Compilation error CS0121

### Steps to Reproduce

global.json
```JSON
{
  "sdk": {
    "version": "8.0.422",
    "rollForward": "latestFeature"
  }
}
```

```csharp
[Test]
public async Task Guid_IsEqualTo_IsAmbiguous_OnSdk8(CancellationToken cancellationToken)
{
    var expected = Guid.Parse("11111111-1111-1111-1111-111111111111");
    var actual = Guid.Parse("11111111-1111-1111-1111-111111111111");

    // CS0121: ambiguous between IsEqualTo<TValue, TOther> and IsEqualTo<TValue>
    await Assert.That(actual).IsEqualTo(expected);
}
```

### TUnit Version

1.56.25

### .NET Version

.NET 8.0

### Operating System

Linux

### IDE / Test Runner

```shell
error CS0121: The call is ambiguous between the following methods or properties: 'EqualsAssertionExtensions.IsEqualTo<TValue, TOther>(IAssertionSource<TValue>, TOther?, string?)' and 'EqualsAssertionExtensions.IsEqualTo<TValue>(IAssertionSource<TValue>, TValue?, string?)'
```

### Additional Context

_No response_

### IDE-Specific Issue?

thomhurst:
Fixes #6296.

## Problem

Consumers building with the **.NET 8 SDK** (pinned via `global.json`) get a compile error on a plain same-type equality assertion:

```csharp
await Assert.That(actual).IsEqualTo(expected); // actual & expected are both Guid
// error CS0121: ambiguous between IsEqualTo<TValue, TOther>(...) and IsEqualTo<TValue>(...)
```

The cross-type `IsEqualTo<TValue, TOther>` / `IsNotEqualTo<TValue, TOther>` overloads (added in #5720 for value-object ergonomics) are equally applicable to the source-generated same-type overloads when the expected value is the same type as the source (`Guid` vs `Guid`). They were deprioritized **only** by `[OverloadResolutionPriority]`.

## Root cause (the key finding)

`[OverloadResolutionPriority]` honoring tracks the **Roslyn compiler version that ships with the build SDK, NOT `LangVersion`.** Reproduced with the exact SDK from the issue:

| Build SDK | effective `LangVersion` | result |
|---|---|---|
| **8.0.422** | `latest` (props imported ✓) | **CS0121** |
| 10.0.301 | `12.0` | builds |

Same project, same package (`TUnit.Assertions` 1.56.25). The `.NET 8` SDK's Roslyn (4.11) simply ignores the attribute, so the `LangVersion=latest` props bump from #6282 is a **no-op** for this scenario — `rollForward: latestFeature` keeps the consumer on the 8.0.4xx SDK band, which never rolls to a 9/10 Roslyn. Any fix must therefore **not depend on ORP**.

## Fix

Give the two cross-type overloads a trailing `params object[]` parameter, making them applicable only in **expanded form**. A same-type call then binds to the normal-form same-type overload via the C# "normal beats expanded" tie-break (§12.6.4.2), which **every** Roslyn version honors — including the .NET 8 SDK's. `[CallerArgumentExpression]` capture is preserved; ORP is kept as a redundant signal on modern compilers. Empty `params` compiles to `Array.Empty<object>()` (no allocation), and this is the value-object path, not a hot path.

Genuine cross-type calls (`Assert.That(productCode).IsEqualTo("X")`) are unaffected — only the cross-type overload is applicable there.

## Validation

- **End-to-end**: packed the fix and consumed it from a project pinned to **SDK 8.0.422**. Both `Assert.That(guid).IsEqualTo(guid)` (was CS0121) **and** `Assert.That(productCode).IsEqualTo("X")` now compile and pass at runtime.
- Existing implicit-conversion / overload-resolution suites: **93/93 pass** (`Issue5720`, `ImplicitStringOperator`, `CollectionOverloadResolution`).
- Added `Issue6296Tests` (5 tests; compilation is the assertion).
- PublicAPI snapshots regenerated for all 4 TFMs (the only change is the added `params object[] _`).

## Known follow-up (not in this PR)

The broader `Assert.That(...)` typed surface (`.All`/`.Count`/`.ContainsKey`, #6276/#6280) is disambiguated the same way (ORP) and is therefore **also** broken for .NET 8 SDK-pinned consumers. The `LangVersion` props fix does not address it for the same reason. Worth a separate pass.

### DIFF
--- a/TUnit.Assertions.Tests/Bugs/Issue6296Tests.cs
+++ b/TUnit.Assertions.Tests/Bugs/Issue6296Tests.cs
@@ -13,12 +13,15 @@ namespace TUnit.Assertions.Tests.Bugs;
 /// solely by the Roslyn that ships with the .NET 9 SDK and later — NOT by LangVersion. The
 /// .NET 8 SDK's compiler ignores it, so the call stayed ambiguous regardless of LangVersion.
 ///
-/// The fix gives the cross-type overloads a trailing <c>params object[]</c>, making them
-/// applicable only in expanded form so they lose to the normal-form same-type overload via
-/// the compiler-version-independent "normal beats expanded" tie-break. The body of each test
-/// below would not compile before the fix when built with the .NET 8 SDK — successful
-/// compilation IS the regression assertion; the runtime assertions confirm the same-type
-/// overload (default equality), not the implicit-conversion overload, is the one selected.
+/// The fix gives the cross-type overloads a trailing <c>params CrossTypeOverloadMarker[]</c>,
+/// making them applicable only in expanded form so they lose to the normal-form same-type
+/// overload via the compiler-version-independent "normal beats expanded" tie-break. The marker
+/// element type (rather than <c>object</c>) has no accessible constructor, so a stray trailing
+/// argument — e.g. <c>IsEqualTo("x", StringComparer.Ordinal)</c> on a value-object source — is a
+/// compile error instead of being silently swallowed and discarded (PR #6313 review). The body
+/// of each test below would not compile before the fix when built with the .NET 8 SDK —
+/// successful compilation IS the regression assertion; the runtime assertions confirm the
+/// same-type overload (default equality), not the implicit-conversion overload, is selected.
 /// </summary>
 public class Issue6296Tests
 {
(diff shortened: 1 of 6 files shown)


thomhurst:
Addressed the 🔴 finding #1 (comparer silently swallowed by `params object[] _`) in 01da30acf.

**Fix:** the trailing `params` element type is now a dedicated `CrossTypeOverloadMarker` with no accessible constructor instead of `object`. The empty-expanded-form tie-break that fixes #6296 is unchanged (callers still pass zero trailing args), but because no caller can ever construct a `CrossTypeOverloadMarker`, a stray trailing argument no longer converts into the params array — it's a compile error again, exactly as before #6296. So:

```csharp
Assert.That(productCode).IsEqualTo("x", StringComparer.OrdinalIgnoreCase); // CS1503 again, not swallowed
```

Verified on the .NET 8 SDK's Roslyn (8.0.4xx): same-type tie-break still binds the same-type overload, genuine cross-type calls still compile, and the footgun call is now `CS1503: cannot convert from 'System.StringComparer' to '...CrossTypeOverloadMarker'`.

Added two regression tests confirming the same-type **with-comparer** overload still binds and the comparer is actually applied (the case that would have been stolen by `object[]`). PublicAPI snapshots regenerated for all 4 TFMs.

Re finding #2 (the param is permanent ABI): still true, and now the marker type is too — but it's `[EditorBrowsable(Never)]` and the trade-off (compile-time safety vs. a hidden public type) seems worth it for a test framework where a silently-ignored comparer = false pass. Re finding #3 (cross-type overloads newly visible on net4.7): unchanged and confirmed intentional.

PR #6314, opened 2026-06-24
fix(mocks): generate mock for qualified-name X.Mock() calls (#6298)

### LINKED ISSUE #6298: [Bug]: Mocking a qualified name doesn't generate the mock class
### Description

If you only reference an interface using a qualified name then the mock class isn't generated.

If something else generates the mock class then everything still works as expected.

### Expected Behavior

Mock class is generated.

### Actual Behavior

Mock class isn't generated.

### Steps to Reproduce

```c#
using Namespace;
using ITestInterfaceFromUsing = Namespace.ITestInterface;

namespace Namespace
{
    public interface ITestInterface
    {
        string Method();
    }
}

public class Tester
{
    [Test]
    public async Task HowYouReferenceMatters()
    {
        var mock1 = Namespace.ITestInterface.Mock();
        mock1.Method().Returns("value");

        // Not only will mock2 work, but this also fixes mock1
        //var mock2 = ITestInterfaceFromUsing.Mock();
        //mock2.Method().Returns("value");
    }
}
```

### TUnit Version

1.56.25

### .NET Version

.NET 10

### Operating System

Windows

### IDE / Test Runner

> error CS1061: 'Mock<ITestInterface>' does not contain a definition for 'Method' and no accessible extension method 'Method' accepting a first argument of type 'Mock<ITestInterface>' could be found (are you missing a using directive or an assembly reference?)

### Additional Context

_No response_

### IDE-Specific Issue?

thomhurst:
Fixes #6298.

## Problem

Invoking `.Mock()` through a fully-qualified type reference didn't generate the mock:

```csharp
var mock = Namespace.ITestInterface.Mock();
mock.Method().Returns("value"); // error CS1061: Mock<ITestInterface> does not contain a definition for 'Method'
```

A simple-name reference (`ITestInterface.Mock()`) or a `using` alias worked — and once *anything* triggered generation, the qualified-name call worked too.

## Root cause

`MockTypeDiscovery.IsMockExtensionInvocation` gated the left-hand side of `X.Mock()` on syntax node kind, accepting `IdentifierNameSyntax`, `GenericNameSyntax`, `QualifiedNameSyntax`, and `AliasQualifiedNameSyntax`.

In **expression** position a dotted name like `Namespace.ITestInterface` parses as a `MemberAccessExpressionSyntax`, **not** a `QualifiedNameSyntax` (that node only appears in type/`using` contexts). So the qualified form fell through the gate and the mock surface was never generated.

## Fix

Drive discovery off the resolved **symbol** rather than the syntactic shape. The predicate now matches any `<expr>.Mock()`, and the existing semantic transform (`TransformMockExtensionInvocation`) already bails unless the LHS resolves to an `INamedTypeSymbol` — which naturally filters out variables, fields, and namespaces. Loosening the cheap syntax gate is therefore safe.

## Tests

- Snapshot: `MockGeneratorTests.Static_Extension_Discovery_Via_Qualified_Name`
- Runtime regression: `TUnit.Mocks.Tests/Issue6298Tests.cs` (mirrors the issue repro)

All snapshot (69) and `TUnit.Mocks.Tests` (1158) tests pass locally on net10.0.


thomhurst:
Rebased onto latest `main` (clean) and investigated the review finding about the `using Issue6298Namespace;` directive in `Issue6298Tests.cs`.

Removing the `using` does **not** compile — the generated mock's setup/proxy members are emitted into the interface's own namespace, so the `using` is genuinely required for the test to build (`CS1061` otherwise). Per the reviewer's fallback suggestion, kept the `using` and added a comment documenting that it's intentional/required, clarifying that the thing actually under test is the fully-qualified `Issue6298Namespace.IIssue6298Interface.Mock()` call form. The snapshot test `Static_Extension_Discovery_Via_Qualified_Name` (no `using`) remains the authoritative regression guard.

All green on net8/9/10.

PR #6323, opened 2026-06-29
feat(aspire): dump resource error logs on test failure + warn on missing telemetry wiring

thomhurst:
## What

Two opt-out DX features so a failing Aspire test surfaces SUT logs without any manual wiring.

### 1. Dump resource error logs on test failure (default on)
`AspireFixture<T>` now implements `ITestEndEventReceiver`. When a test that consumes the fixture fails, each waited-on resource's recent **stderr** lines are appended to that test's output.

- Reuses the existing `ResourceLoggerService.WatchAsync` pattern; collected in parallel with a short per-resource timeout.
- Resource-scoped, **not** request-correlated — on a shared (session/class) fixture the log buffer spans the whole run, so output can include lines from earlier tests. Precise per-request correlation already flows live via the OTLP receiver (trace-context based). This is the coarse fallback for resources that don't export OpenTelemetry logs.
- Knobs: `DumpResourceLogsOnFailure` (default `true`), `MaxFailureLogLinesPerResource` (50), `FailureLogCollectionTimeout` (2s).

### 2. Warn on missing telemetry wiring (default on)
`OtlpReceiver` now records the `service.name` of every incoming OTLP **log** record (before the trace-id filter — presence means OTLP logging reached us at all). At session end the fixture emits a one-time hint for any project resource that produced console output but never sent correlated OTLP logs — the usual symptom of an SUT missing OpenTelemetry log export (e.g. Aspire ServiceDefaults not wired). Turns a silent gap into an actionable nudge.

- Emitted after the receiver drain (seen-set complete) but before `StopAsync` (app still up to probe console output); skipped on a run-abort (incomplete data).
- Name match relies on the fixture-injected `OTEL_SERVICE_NAME=<resourceName>`; an SUT that overrides its own service name may produce a false hint, hence the soft wording. Knob: `WarnOnMissingTelemetry` (default `true`).

## Why

For a SUT wired with standard ServiceDefaults, correlated per-request logs already reach the owning test live. These two changes cover the two remaining gaps with zero user code: failure visibility for uncorrelated **console** logs, and a hint when correlation silently can't work because OTLP log export isn't on.

## Tests

- New `Receiver_LogWithServiceName_RecordsSeenService` (hand-built OTLP logs protobuf, unregistered trace id, asserts case-insensitive `HasSeenLogsFrom` + `SeenLogServiceNames`).
- `OtlpReceiverIngestionTests` 8/8; pure Aspire `TraceRegistryTests` / `ResourcesToRemoveTests` pass; clean build on net10.0.
- Docker-backed integration path is not covered here (no container runtime in this environment) — same gap as the existing Aspire startup-diagnostics tests.

## Notes / landmines for reviewers

- `TestContext.Result` is internal; the public read path is `context.Execution.Result?.State`.
- `OtlpReceiver.ProcessLogs` changed from `static` to an instance method (needs the seen-services field).
- Event receivers fire for `[ClassDataSource]`-injected fixtures because `BuildEligibleEventObjects` scans test class arguments / injected properties.

### DIFF
--- a/TUnit.Aspire.Core/AspireFixture.cs
+++ b/TUnit.Aspire.Core/AspireFixture.cs
@@ -218,6 +218,11 @@ public class AspireFixture<TAppHost> : IAsyncInitializer, IAsyncDisposable, ITes
     /// Time budget for collecting a resource's buffered logs on test failure. The log stream
     /// never completes on its own (it tails live output), so this bounds the wait after the
     /// backlog replays. Default: 2 seconds.
+    /// <para>
+    /// Also bounds the per-resource console-output probe used by the missing-telemetry hint
+    /// (<see cref="WarnOnMissingTelemetry"/>) at session end, so raising it for slow log
+    /// collection also lengthens that teardown probe for any resource not seen exporting OTLP logs.
+    /// </para>
     /// </summary>
     protected virtual TimeSpan FailureLogCollectionTimeout => TimeSpan.FromSeconds(2);
 
@@ -451,6 +456,18 @@ public class AspireFixture<TAppHost> : IAsyncInitializer, IAsyncDisposable, ITes
     {
         var app = _app;
         if (!DumpResourceLogsOnFailure || app is null || context.Execution.Result?.State != TestState.Failed)
+        {
+            return;
+        }
+
+        // OnTestEnd fires once per retry attempt — the end-event receivers run inside the retried
+        // test body (see RetryHelper.ExecuteWithRetry), and TestContext.Output is shared across all
+        // attempts. Dumping on a non-final attempt would leak that attempt's failure logs into the
+        // output of a test that ultimately passes on retry. Only dump on the final attempt. (A custom
+        // retry predicate that declines mid-way terminates as Failed before the limit; in that rarer
+        // case we forgo the dump rather than risk leaking it — the default per-exception retry, which
+        // always runs to the limit on failure, dumps exactly once on the terminal attempt.)
+        if (context.Execution.CurrentRetryAttempt < context.Metadata.TestDetails.RetryLimit)
         {
             return;
         }
@@ -697,6 +714,7 @@ public class AspireFixture<TAppHost> : IAsyncInitializer, IAsyncDisposable, ITes
             // an SUT using the default service name matches by resource name here; one that overrides
             // its service name may produce a false hint — hence the soft wording below.
             var candidates = model.Resources.OfType<ProjectResource>()
+                .Where(ShouldWaitForResource)
                 .Where(p => !receiver.HasSeenLogsFrom(p.Name))
                 .ToList();
             if (candidates.Count == 0)
--- a/TUnit.Aspire.Core/AspireFixture.cs
+++ b/TUnit.Aspire.Core/AspireFixture.cs
@@ -26,7 +26,7 @@ namespace TUnit.Aspire;
 /// or subclass to customize behavior via the virtual configuration hooks.
 /// </para>
 /// </remarks>
-public class AspireFixture<TAppHost> : IAsyncInitializer, IAsyncDisposable
+public class AspireFixture<TAppHost> : IAsyncInitializer, IAsyncDisposable, ITestEndEventReceiver
     where TAppHost : class
 {
     private DistributedApplication? _app;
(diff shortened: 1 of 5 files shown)


thomhurst:
Addressed the open items from the follow-up review in `90a36ff0a`:

- **#2 (memory):** `CollectResourceLogLinesAsync` now keeps a fixed-size sliding window (`Queue<string>` capped at `max`) instead of buffering the full backlog and tail-truncating. Memory stays proportional to `MaxFailureLogLinesPerResource` regardless of how noisy the resource is. The unbounded diagnostics path (no `max`) is unchanged.
- **#4 (alias):** dropped the leftover `var diag = _diagnostics` in `ProcessLogs` (a remnant of the old static signature); the method uses the field directly.
- **FYI (hardcoded timeout):** the telemetry-hint console probe now takes `FailureLogCollectionTimeout` rather than a baked-in 2s, so extending the dump timeout extends the probe too.

`OtlpReceiverIngestionTests` 8/8 green, clean build on net10.0.

thomhurst:
Thanks — the deep pass caught two real bugs. Addressed in `061197f72`:

**#1 (false-positive hint) — fixed.** Confirmed: `OtlpLogParser.ParseLogRecord` returns `null` for any record without a trace id *before* `ProcessLogs` runs, so the per-record `_seenLogServiceNames.TryAdd` only ever fired for records that already had a trace id — exactly inverting the intent. A correctly-wired SUT emitting only untraced logs (startup/background, no active `Activity`) would have triggered the very hint it was meant to avoid. `Parse` now takes an `onResourceSeen` callback invoked ahead of the trace-id filter; the receiver records via a cached delegate (no per-request alloc). The record-emission contract is unchanged (untraced records still excluded from results — two existing parser tests assert that). New test `Parse_UntracedLogRecord_StillReportsSeenService` locks it in.

**#2 (retry leak) — fixed.** Verified against `RetryHelper.ExecuteWithRetry` → `ExecuteTestLifecycleAsync` → `TestExecutor.ExecuteAsync`, whose `finally` fires the end-event receivers *inside* the retried body, and `RetryHelper` clears `Result`/`TestStart`/`Timings` but not `Output`. So a failed attempt-1 dump persisted into an eventually-passing test's output. Gated the dump on `CurrentRetryAttempt >= RetryLimit` (final attempt only). Caveat noted in a comment: a custom retry predicate that declines mid-way terminates as Failed before the limit, and in that rarer case we forgo the dump rather than risk a leak; the default per-exception retry always runs to the limit on failure, so it dumps exactly once on the terminal attempt.

**#4 (filter consistency) — fixed.** `EmitTelemetryWiringHintsAsync` now applies `ShouldWaitForResource` before probing, matching the on-failure dump.

**#6 (doc) — fixed.** `FailureLogCollectionTimeout`'s XML doc now discloses its reuse as the session-end probe timeout.

**#3 (tracking on shared `OtlpReceiver`) — keeping as-is, with reasoning.** `OtlpReceiver` is the component that *receives* OTLP, and it already owns received-signal observability (`Diagnostics`, `RequestCount`) as internal surface; `SeenLogServiceNames` is the same category, not Aspire-specific bookkeeping. `AutoReceiver` carries one extra `ConcurrentDictionary` that stays empty until a log arrives — negligible. On the static/instance asymmetry: `ProcessLogs` needs instance state regardless now (it passes the instance `_recordSeenLogService` to the parser), and `ProcessTraces` has no equivalent seen-set, so the asymmetry reflects a real difference between the two signals rather than a broken convention. A delegate-extraction layer would add indirection for marginal purity benefit. Happy to revisit if you'd still prefer the wrapper.

**#5 (dedup the two `WatchAsync` helpers) — keeping separate, with reasoning.** They share ~20 lines of scaffolding but have opposite termination semantics: the probe returns on the *first* line (early-exit), while the failure dump must drain the *full* backlog until the timeout to keep the *most recent* `max` lines — it can't early-exit on `max` or it'd capture the oldest lines instead. The proposed early-exit-on-`max` path would therefore change the dump's behavior, not just speed it up, so a clean merge isn't available for the lines saved. Left as two focused methods.

`OtlpReceiverIngestionTests` 8/8, `OtlpLogParserTests` (+1 new) and `OtlpReceiverTests` all green on net10.0.

PR #6465, opened 2026-07-21
perf: remove EnumerableAsyncProcessor dependency

thomhurst:
## Summary

Removes the `EnumerableAsyncProcessor` package dependency from `TUnit.Engine` (one fewer transitive dependency for every TUnit user) and replaces it with an internal, allocation-light `ParallelMap` helper purpose-built for the discovery hot path.

### The replacement (`ParallelMap`)

- Bounded-parallelism async map: a fixed set of workers (`min(DOP, count)`) pulls items via an interlocked cursor — no partitioner, no channels, no per-item `Task`, one results array, one shared closure.
- Results preserve source order (written by index).
- Sets below 8 items run sequentially, skipping task scheduling overhead entirely (previously only one of the three callsites had this fallback).
- On a selector exception the cursor is parked so sibling workers stop pulling remaining items; `Task.WhenAll` surfaces the fault.
- Works uniformly across all TFMs including `netstandard2.0` (where `Parallel.ForEachAsync` is unavailable).

### Cleanups enabled by the rewrite

- `TestBuilderPipeline`'s three build paths (sequential small-set branch, parallel branch, "streaming" branch) collapsed onto one shared per-metadata method — the streaming path was fake-streaming (fully buffered before yielding) and its dynamic-test branch was a ~120-line copy of `GenerateDynamicTests`.
- Dead `CreateFailedTestForGenericResolutionError` + unreachable resolution-error scaffolding deleted.
- Build results are flattened eagerly with exact list capacity instead of a lazy `SelectMany` that `TestRegistry` enumerated twice.
- `UploadToNuGetModule` push loop rewritten as a plain sequential `for` (same one-at-a-time semantics).

Net: **-138 lines** (163+ / 301-).

### Behavior notes

- The former streaming path was *unbounded* parallel; it is now bounded at `Environment.ProcessorCount`, consistent with the non-streaming path.
- Dynamic-test repeats (2–7) now build sequentially via the shared threshold — they previously paid `Task.Run` worker overhead.

## Test plan

- [x] `TUnit.Engine` + `TUnit.Pipeline` build clean on all TFMs, 0 warnings
- [x] `TUnit.UnitTests`: 259/259 pass
- [x] `TUnit.TestProject` dynamic-test filter: pass in **both** source-gen and reflection modes
- [x] `TUnit.TestProject` repeat-attribute tests: pass
- [x] Repo-wide grep: zero remaining `EnumerableAsyncProcessor` references

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -12,7 +12,6 @@
     <PackageVersion Include="BenchmarkDotNet.Annotations" Version="0.15.8" />
     <PackageVersion Include="coverlet.collector" Version="10.0.1" />
     <PackageVersion Include="CliWrap" Version="3.10.2" />
-    <PackageVersion Include="EnumerableAsyncProcessor" Version="3.8.4" />
     <PackageVersion Include="FakeItEasy" Version="9.0.1" />
     <PackageVersion Include="FluentValidation.DependencyInjectionExtensions" Version="12.1.1" />
     <PackageVersion Include="FsCheck" Version="3.3.3" />
--- a/src/TUnit.Engine/Building/TestBuilderPipeline.cs
+++ b/src/TUnit.Engine/Building/TestBuilderPipeline.cs
@@ -52,17 +52,6 @@ internal sealed class TestBuilderPipeline
         return testBuilderContext;
     }
 
-    [UnconditionalSuppressMessage("Trimming", "IL2026", Justification = "Reflection mode is not used in AOT/trimmed scenarios")]
-    [UnconditionalSuppressMessage("AOT", "IL3050", Justification = "Reflection mode is not used in AOT scenarios")]
-    public async Task<IEnumerable<AbstractExecutableTest>> BuildTestsAsync(string testSessionId)
-    {
-        var collectedMetadata = await _dataCollector.CollectTestsAsync(testSessionId).ConfigureAwait(false);
-
-        // For this method (non-streaming), we're not in execution mode so no filter optimization
-        var buildingContext = new TestBuildingContext(IsForExecution: false, Filter: null);
-        return await BuildTestsFromMetadataAsync(collectedMetadata, buildingContext).ConfigureAwait(false);
-    }
-
     /// <summary>
     /// Collects all test metadata without building tests.
     /// This is a lightweight operation used for dependency analysis.
@@ -88,7 +77,7 @@ internal sealed class TestBuilderPipeline
     /// <param name="buildingContext">Context for test building</param>
     /// <param name="metadataFilter">Optional predicate to filter which metadata should be built (null means build all)</param>
     /// <param name="cancellationToken">Cancellation token</param>
-    public async Task<IEnumerable<AbstractExecutableTest>> BuildTestsStreamingAsync(
+    public async Task<IEnumerable<AbstractExecutableTest>> BuildTestsAsync(
         string testSessionId,
         TestBuildingContext buildingContext,
         Func<TestMetadata, bool>? metadataFilter = null,
--- a/src/TUnit.Engine/Building/TestBuilderPipeline.cs
+++ b/src/TUnit.Engine/Building/TestBuilderPipeline.cs
@@ -1,6 +1,4 @@
-using System.Collections.Concurrent;
 using System.Diagnostics.CodeAnalysis;
-using EnumerableAsyncProcessor.Extensions;
 using Microsoft.Testing.Platform.Requests;
 using TUnit.Core;
 using TUnit.Core.Helpers;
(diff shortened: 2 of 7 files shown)


thomhurst:
## Before/after performance

**End-to-end discovery** — `TUnit.TestProject.exe --list-tests` (18,138 tests, net10.0, Release, Windows x64). Baseline = `main` (ede5377e33) built in a separate worktree; interleaved A/B runs after warmup:

| Mode | Before (median) | After (median) | Delta |
|---|---|---|---|
| Source-gen (12 runs each) | 1847 ms | 1271 ms | **-576 ms (-31%)** |
| Reflection (8 runs each) | 2032 ms | 1486 ms | **-546 ms (-27%)** |

Source-gen std dev ~60 ms per variant — the gap is ~9 sigma, well outside noise. Wall-clock includes constant process startup/JIT, so the discovery-phase improvement is larger than the percentage suggests.

**Primitive micro-benchmark** — `EnumerableAsyncProcessor` (`SelectAsync().ProcessInParallel(ProcessorCount)`) vs `ParallelMap.SelectParallelAsync`, same DOP, mean of 20 iterations after warmup, allocations via `GC.GetTotalAllocatedBytes`:

| N | Selector | EAP time | EAP alloc | ParallelMap time | ParallelMap alloc | Speedup | Alloc reduction |
|---|---|---|---|---|---|---|---|
| 100 | sync-completing | 0.069 ms | 44.5 KB | 0.020 ms | 10.7 KB | 3.5x | 4.2x |
| 100 | `Task.Yield` | 0.177 ms | 86.7 KB | 0.030 ms | 15.3 KB | 6.0x | 5.7x |
| 1,000 | sync-completing | 1.021 ms | 432 KB | 0.076 ms | 77.5 KB | 13.4x | 5.6x |
| 1,000 | `Task.Yield` | 1.527 ms | 847 KB | 0.166 ms | 104 KB | 9.2x | 8.2x |
| 10,000 | sync-completing | 7.161 ms | 4.15 MB | 0.622 ms | 745 KB | 11.5x | 5.7x |
| 10,000 | `Task.Yield` | 16.960 ms | 8.26 MB | 1.661 ms | 983 KB | 10.2x | 8.6x |

The end-to-end win comes from the primitive swap plus deleting the fake-streaming path (per-metadata async-iterator state machines and triple list copies).

thomhurst:
Addressed all three review points in 5b4004e:

**1. ParallelMap is now the single bounded-parallelism strategy.** `InvokePostResolutionEventsInParallelAsync` routes through a new `ParallelMap.ForEachParallelAsync` overload (takes `Func<T, ValueTask>` so sync-completing event receivers don't allocate a `Task` per test, and skips the results array entirely). The `#if NET8_0_OR_GREATER` / `Parallel.ForEachAsync` split is gone — including the netstandard2.0 `Task.WhenAll` fallback that was unbounded. The threshold check at the callsite is gone too; the helper owns that policy.

**2. Dedicated `ParallelMapTests` added** (13 test cases): source-order preservation under staggered concurrent completion, `SequentialThreshold` boundary counts (7/8/9), exactly-once invocation, DOP cap never exceeded (tracked via interlocked high-water mark), selector exception propagation, fail-fast cursor parking (verifies siblings stop pulling — invoked count stays well below N=10,000 when item 0 throws), pre-cancelled token, and the `ForEachParallelAsync` equivalents.

**3. Naming fixed.** `TestBuilderPipeline.BuildTestsStreamingAsync` → `BuildTestsAsync`; the old zero-caller `BuildTestsAsync(string testSessionId)` overload turned out to be dead code and is deleted. `TestBuilder.BuildTestsStreamingAsync` (genuinely `IAsyncEnumerable`-based) keeps its name unambiguously.

Verified: engine builds clean on all TFMs, `TUnit.UnitTests` 272/272 (259 + 13 new), TestProject dynamic-test smoke passes.

thomhurst:
Re-profiled at head 5b4004e (post-review changes: `ForEachParallelAsync` convergence, rename, tests) — results hold:

| Measurement | Before (main) | After (5b4004e) | Delta |
|---|---|---|---|
| Discovery, source-gen (median, 12 interleaved runs) | 1936 ms | 1281 ms | **-34%** |
| Discovery, reflection (median, 10 runs) | 2016 ms | 1476 ms | **-27%** |

(Noisier machine this session — two spikes hit both variants equally thanks to interleaving; medians robust.)

Primitive micro-benchmark unchanged within noise: ~9–12x faster, ~4–9x fewer allocations vs `EnumerableAsyncProcessor` at N=1,000–10,000 (e.g. N=10,000 `Task.Yield`: 14.3 ms / 8.25 MB → 1.6 ms / 984 KB). The new `ForEachParallelAsync` shares the same worker loop.

PR #6473, opened 2026-07-23
fix(mocks): omit 'struct' constraint when type parameter has 'unmanaged' constraint

### LINKED ISSUE #6471: [TUnit.Mocks] Generated mock for interface with 'unmanaged' type constraint does not compile (emits 'where T : struct, unmanaged')
### Description

`Mock.Of<T>()` over a generic interface whose type parameter has an `unmanaged` constraint produces generated code that does not compile. The generator emits `where T : struct, unmanaged` on the mock classes, but C# forbids combining `struct` and `unmanaged` (CS0449). Because the constraint clause is invalid, `T` is then not recognized as satisfying the interface's `unmanaged` constraint, which cascades into ~30 CS8377 errors across the generated files.

### Minimal reproduction

```csharp
public interface ISnapshotSource<T> where T : unmanaged
{
    void Fill(ref T snapshot);
}

public sealed class ReproTests
{
    [Test]
    public async Task Mock_OfUnmanagedConstrainedInterface_Compiles()
    {
        var sourceMock = Mock.Of<ISnapshotSource<int>>();

        await Assert.That(sourceMock).IsNotNull();
    }
}
```

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>net10.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
    <IsTestProject>true</IsTestProject>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="TUnit" Version="1.61.29" />
    <PackageReference Include="TUnit.Mocks" Version="1.61.29" />
  </ItemGroup>
</Project>
```

### Actual behavior

`dotnet build` fails with 33 errors in the generated sources. The generated mock class is declared as:

```csharp
public sealed class ISnapshotSource_T_Mock<T> : global::TUnit.Mocks.Mock<global::ISnapshotSource<T>>, global::ISnapshotSource<T> where T : struct, unmanaged
```

Representative errors:

```
ISnapshotSource_T__Mock.g.cs(5,148): error CS0449: The 'class', 'struct', 'unmanaged', 'notnull', and 'default' constraints cannot be combined or duplicated, and must be specified first in the constraints list.
ISnapshotSource_T__Mock.g.cs(5,21): error CS8377: The type 'T' must be a non-nullable value type, along with all fields at any level of nesting, in order to use it as parameter 'T' in the generic type or method 'ISnapshotSource<T>'
```

The same pattern (`where T : struct, unmanaged, ...`) is emitted in `ISnapshotSource_T__MockImplFactory.g.cs` and fails the same way.

### Expected behavior

The generated mock compiles, emitting `where T : unmanaged` only (the `unmanaged` constraint already implies a non-nullable value type, so `struct` must be omitted).

### Environment

- TUnit / TUnit.Mocks 1.61.29
- .NET SDK 10.0.302, `net10.0`
- macOS (arm64), also reproduced in a larger solution with `where T : unmanaged, ISomeInterface`

thomhurst:
## Summary

Fixes #6471.

`Mock.Of<T>()` over a generic interface whose type parameter has an `unmanaged` constraint generated code that does not compile. Roslyn reports `HasValueTypeConstraint == true` for `unmanaged`-constrained type parameters, so `GetGenericConstraints` emitted `where T : struct, unmanaged` — invalid C# (CS0449), cascading into ~30 CS8377 errors across the generated mock, impl, factory, and extension files.

## Fix

In `MethodSymbolExtensions.GetGenericConstraints` (the single constraint-emission helper, used for both interface-level and method-level type parameters), only emit `struct` when the type parameter does **not** also have the `unmanaged` constraint. `unmanaged` already implies a non-nullable value type.

## Tests

- New snapshot test `Interface_With_Unmanaged_Constrained_Type_Parameter` covering `where T : unmanaged` (with the issue's `ref T` repro shape) and `where T : unmanaged, IDisposable`.
- New compile-guard test `Interface_With_Unmanaged_Constrained_Type_Parameter_Compiles` asserting the generated output produces no CS0449/CS8377.
- Updated `Interface_With_Generic_Method_Constraints_On_Explicit_Impl.verified.txt`: the pre-existing `GetUnmanaged<T>()` case now emits `where T : unmanaged` instead of `where T : struct, unmanaged` (only change in the snapshot).

All 80 TUnit.Mocks.SourceGenerator.Tests and 1160 TUnit.Mocks.Tests pass locally (net10.0).


thomhurst:
Follow-up for the review comment above (sibling generators with the same `unmanaged` constraint bug): #6474 — fixes CollectionShapeAssertionGenerator (CS0449), PropertyInjectionSourceGenerator (silent downgrade), and the four emitters that dropped `unmanaged` entirely. Shared-helper consolidation left as a separate refactor.

PR #6501, opened 2026-07-28
fix(mocks): report TM006 instead of CS1729 for unsubclassable classes

### LINKED ISSUE #6493: [TUnit.Mocks] Mock() fails for types with no accessible parameterless constructor (e.g. Azure.Messaging.ServiceBus.Administration.*RuntimeProperties): CS1729
## Description

`T.Mock()` fails to compile for `Azure.Messaging.ServiceBus.Administration.QueueRuntimeProperties`, `SubscriptionRuntimeProperties`, and `TopicRuntimeProperties` (from `Azure.Messaging.ServiceBus` 7.20.2) — none of these types have an accessible parameterless constructor, but the generated mock implementation factory unconditionally emits `new T()`.

## Repro

```xml
<PackageReference Include="Azure.Messaging.ServiceBus" Version="7.20.2" />
```

```csharp
using Azure.Messaging.ServiceBus.Administration;
using TUnit.Mocks;

var props = QueueRuntimeProperties.Mock(); // fails
```

Note: merely mocking `ServiceBusAdministrationClient` (whose `GetQueueRuntimePropertiesAsync` etc. return `Response<QueueRuntimeProperties>`) does **not** trigger this on its own — the failure only appears once `QueueRuntimeProperties.Mock()` (or the analogous Subscription/Topic type) is actually invoked, e.g. to construct a `Response<T>` return value for a setup. Confirmed with a minimal, isolated repro.

## Environment

- Azure.Messaging.ServiceBus: 7.20.2
- TFM: net10.0
- OS: Windows 11

## Actual error (one of three, identical pattern for all three types)

```
...MockGenerator\Azure_Messaging_ServiceBus_Administration_QueueRuntimeProperties_MockImplFactory.g.cs(15,18):
  error CS1729: 'QueueRuntimeProperties' does not contain a constructor that takes 0 arguments
```

Reflecting over the type confirms it genuinely has no public or non-public parameterless constructor:

```
QueueRuntimeProperties ctors: (String) [non-public]
SubscriptionRuntimeProperties ctors: (String, String) [non-public]
TopicRuntimeProperties ctors: (String) [non-public]
```

## Suspected cause

The mock-impl generator assumes every mockable concrete type has (or can synthesize) a parameterless constructor to build the backing instance, and doesn't fall back to an alternative construction strategy (e.g. `RuntimeHelpers.GetUninitializedObject`, or simply declining to support "as a return value" mocking for such types with a clear diagnostic) when no such constructor exists.

## Impact

Any test that needs a `Response<QueueRuntimeProperties>` / `Response<SubscriptionRuntimeProperties>` / `Response<TopicRuntimeProperties>` value from a mocked `ServiceBusAdministrationClient` setup cannot use `T.Mock()` for that return value at all.

## Context

Found while migrating an OSS health-checks library (dailydevops/healthchecks) from NSubstitute to TUnit.Mocks across ~65 test files.

thomhurst:
Fixes #6493

## Problem

`T.Mock()` on a class whose constructors are all inaccessible produced:

```
QueueRuntimeProperties_MockImplFactory.g.cs(15,18): error CS1729:
'QueueRuntimeProperties' does not contain a constructor that takes 0 arguments
```

`MemberDiscovery.DiscoverConstructors` only collects *accessible* constructors, so the model ended up with an empty `Constructors` array and `MockFactoryBuilder.GenerateConstructorDispatch` fell through to `new XMockImpl(engine)`. The impl derives from the target, and its implicit `base()` call has nothing to bind to. The error pointed into generated code with no link back to the call site.

Such a type genuinely cannot be mocked: every subclass constructor must chain to a base constructor, so with all of them private (or internal in another assembly) no subclass can be declared at all. The right outcome is a clear diagnostic, not generated code that cannot compile.

## Fix

- `MockTypeModel.LacksAccessibleConstructor` — true for a class target with no chainable constructor.
- `MockGenerator` skips impl/factory/members for those targets and emits only the static `Mock()` entry point, so the call site still binds and the diagnostic below is the single error the user sees.
- New `TM006` + `InaccessibleConstructorMockAnalyzer`, reported at the call site for `Mock.Of<T>()`, `Mock.Wrap<T>()` and the generated `T.Mock()` form. The message points at the library's own factory (e.g. Azure's `ServiceBusModelFactory`) as the way to build such a value.

Protected and protected-internal constructors remain chainable — the generated impl derives from the target — so abstract classes are unaffected. Sealed and value types are left to TM001/TM002 rather than double-reported.

## Tests

`tests/TUnit.Mocks.Analyzers.Tests/InaccessibleConstructorMockAnalyzerTests.cs` — 9 cases: private-only ctor via all three entry points, plus non-reporting cases for public / implicit parameterless / protected / same-assembly internal constructors, interfaces, and sealed types.

Analyzer tests: 39 passed. `TUnit.Mocks.Tests`: 1160 passed. Generator snapshots: 80 passed, unchanged.

### DIFF
--- a/src/TUnit.Mocks.Analyzers/InaccessibleConstructorMockAnalyzer.cs
+++ b/src/TUnit.Mocks.Analyzers/InaccessibleConstructorMockAnalyzer.cs
@@ -80,11 +80,8 @@ public class InaccessibleConstructorMockAnalyzer : DiagnosticAnalyzer
                 : null;
         }
 
-        // Generated per-type entry point: `T.Mock()`. The generated static extension class always
-        // lives in namespace TUnit.Mocks; its members bind through a compiler-synthesised nested
-        // extension type, so match on the namespace rather than the immediate containing type.
-        if (methodSymbol is not { Name: "Mock", IsStatic: true }
-            || !IsTUnitMocksNamespace(methodSymbol.ContainingNamespace)
+        // Generated per-type entry point: `T.Mock()`.
+        if (!IsGeneratedStaticEntryPoint(methodSymbol)
             || invocation.Expression is not MemberAccessExpressionSyntax memberAccess)
         {
             return null;
@@ -92,6 +89,32 @@ public class InaccessibleConstructorMockAnalyzer : DiagnosticAnalyzer
 
         return context.SemanticModel.GetSymbolInfo(memberAccess.Expression, context.CancellationToken).Symbol
             as INamedTypeSymbol;
+    }
+
+    /// <summary>
+    /// Matches the <c>Mock()</c> the generator emits: a static member of a C# 14
+    /// <c>extension(T)</c> block inside a <c>*_MockStaticExtension</c> class in namespace
+    /// <c>TUnit.Mocks</c>. Roslyn reports the member's immediate containing type as the
+    /// synthesised, unnamed extension type, so the check walks out to the declaring class —
+    /// matching on the namespace alone would claim any static <c>Mock()</c> a consumer happens to
+    /// declare there.
+    /// </summary>
+    private static bool IsGeneratedStaticEntryPoint(IMethodSymbol method)
+    {
+        if (method is not { Name: "Mock", IsStatic: true } || !IsTUnitMocksNamespace(method.ContainingNamespace))
+        {
+            return false;
+        }
+
+        for (var containing = method.ContainingType; containing is not null; containing = containing.ContainingType)
+        {
+            if (containing.Name.EndsWith("_MockStaticExtension", System.StringComparison.Ordinal))
+            {
+                return true;
+            }
+        }
+
+        return false;
     }
 
     private static bool IsTUnitMocksNamespace(INamespaceSymbol? ns)
--- a/src/TUnit.Mocks.Analyzers/InaccessibleConstructorMockAnalyzer.cs
+++ b/src/TUnit.Mocks.Analyzers/InaccessibleConstructorMockAnalyzer.cs
@@ -75,7 +75,11 @@ public class InaccessibleConstructorMockAnalyzer : DiagnosticAnalyzer
 
         if (IsMockEntryPointMethod(methodSymbol))
         {
-            return methodSymbol.TypeArguments.Length == 1
+            // The multi-type overloads — Of<T1, T2>() through Of<T1, T2, T3, T4>() — return
+            // Mock<T1>: T1 is the type the impl subclasses, T2..T4 are interfaces layered on it,
+            // and MockTypeDiscovery reuses T1's constructors for the multi-type model. So the
+            // first type argument is the constructor-bearing target for every overload.
+            return methodSymbol.TypeArguments.Length >= 1
                 ? methodSymbol.TypeArguments[0] as INamedTypeSymbol
                 : null;
         }
--- a/src/TUnit.Mocks.Analyzers/InaccessibleConstructorMockAnalyzer.cs
+++ b/src/TUnit.Mocks.Analyzers/InaccessibleConstructorMockAnalyzer.cs
@@ -105,8 +105,12 @@ public class InaccessibleConstructorMockAnalyzer : DiagnosticAnalyzer
     }
 
     /// <summary>
-    /// Mirrors the generator's constructor discovery: a generated subclass can chain to any
-    /// constructor that is not private, and not internal-without-access from another assembly.
+    /// Mirrors the generator's constructor discovery — <c>MemberDiscovery.DiscoverConstructors</c>
+    /// → <c>IsMemberAccessible</c> → <c>AreMemberSignatureTypesAccessible</c>. A generated subclass
+    /// can chain to a constructor only when the constructor itself is reachable AND every one of
+    /// its parameter types is: the generator drops a constructor whose signature mentions an
+    /// inaccessible type, and a target left with none is exactly the case this rule reports.
+    /// Keep the two in step; they live in separate assemblies with no shared project.
     /// Protected (and protected internal) constructors are reachable precisely because the
     /// generated impl derives from the target.
     /// </summary>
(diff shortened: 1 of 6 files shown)


thomhurst:
@claude @codex could we use UnsafeAccessor to allow mocking these? It would be good if we could still actually mock them - unsafe accessory gives us reflection-like access, but should also work for aot?

thomhurst:
Thanks — finding 1 is a real divergence and is fixed in 2985b13. One correction on the worked example, and a push-back on the extraction.

**1. Divergence fixed; the specific repro isn't constructible though.** `IsChainable` now applies the same signature-type rule as `AreMemberSignatureTypesAccessible` / `IsTypeAccessible` (arrays, generic type arguments and pointers included), so the analyzer and `DiscoverConstructors` can no longer disagree.

I did try to pin it with the scenario you described — a public constructor whose parameter type is internal in another assembly — and it can't exist: the declaring assembly won't compile it.

```
Library.cs(9,16): error CS0051: Inconsistent accessibility: parameter type 'Capability' is less
accessible than method 'GatedService.GatedService(Capability)'
```

C# requires a constructor to be no more accessible than its parameter types, so for constructors the parameter-type gate can only fire where the constructor itself is already gated (`internal`, `private protected`) and therefore already reported. I kept the check anyway — it costs nothing, removes the divergence, and the rule is shared with member signatures where it does bite — but I couldn't write a failing test for it, so the test I added instead is the cross-assembly `internal` constructor: the actual Azure `QueueRuntimeProperties` shape, which the previous same-assembly-private tests could only approximate. Plus a public-constructor-with-public-parameter case guarding the new check against false positives.

**2. On the real `T.Mock()` extension shape.** Agreed it's the weakest point, and I'd take real coverage if it were available. It isn't within this PR: the analyzer test harness resolves `Microsoft.CodeAnalysis.CSharp.Workspaces` 4.12.0 / `Microsoft.CodeAnalysis.Common` 4.7.0 from `Directory.Packages.props`, which is repo-wide and also the floor the shipped analyzer and generator compile against. Bumping it to a Roslyn that parses `extension(...)` blocks raises the minimum Roslyn for consumers, which I don't think belongs in a bug-fix PR. Running generator output through the analyzer in `TUnit.Mocks.SourceGenerator.Tests` hits the same pinned Roslyn, and the only real-SDK project (`TUnit.Mocks.Tests`) can't host a TM006 trigger because TM006 is an error and would break the build.

What I do have is a real-SDK verification: with the branch built into `TUnit.Mocks.Tests`, both entry points report and the CS1729 is gone —

```
Issue6493Repro.cs(17,20): error TM006: Cannot mock 'NoAccessibleCtor' ...   // NoAccessibleCtor.Mock()
Issue6493Repro.cs(18,21): error TM006: Cannot mock 'NoAccessibleCtor' ...   // Mock.Of<NoAccessibleCtor>()
```

That's also what drove the namespace-based matching in the first place: matching on `ContainingType.Name.EndsWith("_MockStaticExtension")` did *not* fire against the real extension-block symbol, only the namespace check did. Happy to open a follow-up issue for a newer-Roslyn analyzer test harness if you'd like that tracked.

Analyzer tests 41 passed, `TUnit.Mocks.Tests` 1160, generator snapshots 80.

thomhurst:
> could we use UnsafeAccessor to allow mocking these? ... unsafe accessor gives us reflection-like access, but should also work for aot?

I tested both halves of this against the real Azure package. Short answer: it can't unlock **mocking**, but it can unlock a genuinely useful **stub/value-builder** — and yes, it's AOT-safe.

**It doesn't help mocking, for two independent reasons.**

*1. The blocker is subclassing, not construction.* TUnit.Mocks intercepts by emitting a subclass, and every derived constructor must chain to an accessible base constructor. `UnsafeAccessorKind.Constructor` produces a factory-shaped method (it `newobj`s and returns an instance) — it can't stand in for a `base(...)` initializer, and the generator emits C#, which always emits that call. So the impl still won't compile:

```
Sub.cs(7,12): error CS1729: 'QueueRuntimeProperties' does not contain a constructor that takes 0 arguments
```

*2. Even with a subclass there'd be nothing to intercept.* Reading the metadata for `QueueRuntimeProperties`:

```
TYPE QueueRuntimeProperties attrs=Public, BeforeFieldInit
  methods: virtual=0 nonvirtual=23
  properties: Name:Assembly, TotalMessageCount:Assembly, ActiveMessageCount:Assembly, ... (all internal setters)
```

Zero virtual members. A subclass could override nothing, so `.Returns(...)` would have no slot to attach to regardless of how we got an instance.

**Where it does pay off.** What people actually want from these types is a *value* to hand back from a mocked `GetQueueRuntimePropertiesAsync`. UnsafeAccessor does that cleanly — internal constructor plus internal setters, no backing-field name guessing:

```csharp
[UnsafeAccessor(UnsafeAccessorKind.Constructor)]
static extern QueueRuntimeProperties NewProps(string name);

[UnsafeAccessor(UnsafeAccessorKind.Method, Name = "set_ActiveMessageCount")]
static extern void SetActive(QueueRuntimeProperties props, long value);

var p = NewProps("my-queue");
SetActive(p, 42);
// prints: OK Name=my-queue ActiveMessageCount=42
```

That ran clean against `Azure.Messaging.ServiceBus` 7.20.2. It's AOT-safe — the runtime/ILC binds these at link time, which is exactly why it's the sanctioned replacement for private reflection.

Caveats worth pricing in before committing to it: it needs `#if NET8_0_OR_GREATER` (TUnit.Mocks also targets netstandard2.0, and consumer-side TFM guards matter here — see the net8.0 issue on #6503); it deliberately bypasses accessibility, so a library renaming an internal member breaks consumers at *runtime* rather than compile time; and the result is a stub, not a mock — no verification, no setups.

So I'd frame it as a separate feature — something like `T.Stub()` returning a configurable instance of an otherwise-unconstructible type — rather than something that changes TM006. TM006 is still the right answer for `T.Mock()` on these types, because mocking them genuinely cannot work. Happy to open an issue for the stub feature with the above as the feasibility evidence if you want to pursue it.

thomhurst:
Fixed in 9c17356. The multi-type gap is real — verified end to end rather than taken on the review's word:

- `Mock.Of<T1, T2>()` … `Of<T1, T2, T3, T4>()` all return `Mock<T1>` and route to `OfMulti<T1>` (`src/TUnit.Mocks/Mock.cs:123-181`).
- `MockTypeDiscovery` builds the multi-type model with `Constructors = singleTypeModel.Constructors` (`Discovery/MockTypeDiscovery.cs:192`), i.e. T1's constructors verbatim — so `LacksAccessibleConstructor` fires on exactly the same condition as the single-type model and generation is skipped.
- The analyzer's `TypeArguments.Length == 1` gate therefore let precisely those calls through with no diagnostic, converting a compile-time error into a runtime `InvalidOperationException`.

`ResolveMockTarget` now takes `TypeArguments[0]` whenever there's at least one type argument, with a comment recording why the first argument is the constructor-bearing target for every overload. Added `Multi_Type_Of_With_Inaccessible_Primary_Constructor_Reports_TM006` covering `Mock.Of<RuntimeProperties, ISomeInterface>()`; the suite is 12/12 green.

On the marker-attribute suggestion for `IsGeneratedStaticEntryPoint`: agreed in principle, but it's a cross-project contract change (the generator would have to stamp a marker `TUnit.Mocks` declares, and the analyzer resolve it by symbol) that touches every generated static extension and its snapshots. That's a wider change than this fix warrants — leaving the convention match, which is exercised by `Unrelated_Static_Mock_Method_Does_Not_Report`.

Also rebased onto main to clear the conflict (`Rules.cs` — TM006 and TM007 landed in the same slot; both descriptors kept).

PR #6502, opened 2026-07-28
fix(mocks): skip interfaces with inaccessible abstract members

### LINKED ISSUE #6491: [TUnit.Mocks] Mock() fails for IDocumentStore (RavenDB.Client): asymmetric-accessor member CS0535/CS0548/CS0551
## Description

`T.Mock()` fails to compile when `T` is `Raven.Client.Documents.IDocumentStore` (from `RavenDB.Client` 7.2.5). The generated mock implementation does not correctly handle a nested interface member, `Raven.Client.Json.Serialization.ISessionBlittableJsonConverter.MissingProperties`, resulting in 9 compiler errors.

## Repro

```xml
<PackageReference Include="RavenDB.Client" Version="7.2.5" />
```

```csharp
using Raven.Client.Documents;
using TUnit.Mocks;

var store = IDocumentStore.Mock();
```

Confirmed minimal, isolated repro (single-file console app, no other mocked types in the compilation) — this is not usage-order-dependent or masked by other errors.

## Environment

- RavenDB.Client: 7.2.5
- TFM: net10.0
- OS: Windows 11

## Actual errors

```
...MockGenerator\Raven_Client_Json_Serialization_ISessionBlittableJsonConverter_MockImplFactory.g.cs(7,64):
  error CS0535: 'ISessionBlittableJsonConverterMockImpl' does not implement interface member
  'ISessionBlittableJsonConverter.MissingProperties.get'

...MockImplFactory.g.cs(7,64):
  error CS0535: 'ISessionBlittableJsonConverterMockImpl' does not implement interface member
  'ISessionBlittableJsonConverter.MissingProperties.set'

...MockImplFactory.g.cs(76,133):
  error CS0548: 'ISessionBlittableJsonConverterMockImpl.MissingProperties': property or indexer
  must have at least one accessor

...Raven_Client_Json_Serialization_ISessionBlittableJsonConverter_Mock.g.cs(49,197):
  error CS0122: 'ISessionBlittableJsonConverter.MissingProperties' is inaccessible due to its
  protection level

...Mock.g.cs(7,160):
  error CS0535: 'ISessionBlittableJsonConverterMock' does not implement interface member
  'ISessionBlittableJsonConverter.MissingProperties.get'

...Mock.g.cs(7,160):
  error CS0535: 'ISessionBlittableJsonConverterMock' does not implement interface member
  'ISessionBlittableJsonConverter.MissingProperties.set'

...Mock.g.cs(49,197):
  error CS0548: 'ISessionBlittableJsonConverterMock.ISessionBlittableJsonConverter.MissingProperties':
  property or indexer must have at least one accessor

...Mock.g.cs(49,197):
  error CS0551: Explicit interface implementation
  'ISessionBlittableJsonConverterMock.ISessionBlittableJsonConverter.MissingProperties' is missing
  accessor 'ISessionBlittableJsonConverter.MissingProperties.get'

...Mock.g.cs(49,197):
  error CS0551: Explicit interface implementation
  'ISessionBlittableJsonConverterMock.ISessionBlittableJsonConverter.MissingProperties' is missing
  accessor 'ISessionBlittableJsonConverter.MissingProperties.set'
```

## Suspected cause

`ISessionBlittableJsonConverter.MissingProperties` appears to be a member with mismatched/asymmetric accessor accessibility (e.g. one accessor is `internal`/non-public while the other is public, or the property is declared with only one accessor at a differing visibility than the generator assumes). The generator emits both a get and set accessor for this member regardless, producing a member that does not satisfy the interface contract and is also inaccessible from the generated assembly.

## Impact

Any type whose interface graph transitively includes this member (in this case reached via `IDocumentStore` → ... → `ISessionBlittableJsonConverter`) cannot be mocked with `T.Mock()` at all — there's no per-member opt-out to work around it from the calling code.

## Context

Found while migrating an OSS health-checks library (dailydevops/healthchecks) from NSubstitute to TUnit.Mocks across ~65 test files. Isolated with a minimal single-file repro to rule out cross-file interference.

thomhurst:
Fixes #6491

## Problem

`Mock.Of<Raven.Client.Documents.IDocumentStore>()` produced 9 errors on generated code:

```
ISessionBlittableJsonConverter_MockImplFactory.g.cs(7,64): error CS0535: 'ISessionBlittableJsonConverterMockImpl'
  does not implement interface member 'ISessionBlittableJsonConverter.MissingProperties.get'
..._Mock.g.cs(49,197): error CS0122: 'ISessionBlittableJsonConverter.MissingProperties' is inaccessible
  due to its protection level
```

Confirmed against the shipped assembly — `MissingProperties`'s accessors carry `Assembly` accessibility:

```
TYPE Raven.Client.Json.Serialization.ISessionBlittableJsonConverter attrs=Public, ClassSemanticsMask, Abstract
  PROP MissingProperties
    get=Assembly, Virtual, HideBySig, Abstract, SpecialName
    set=Assembly, Virtual, HideBySig, Abstract, SpecialName
```

Since C# 8 an interface may declare non-public abstract members, and no type outside the declaring assembly can implement them. `MemberDiscovery` correctly filtered the inaccessible member out of the model, so the generated impl never declared it — but the impl still had to satisfy the interface, hence CS0535.

The user never asked to mock that interface: transitive auto-mock discovery reached it from `IDocumentStore`, so there was no way to opt out from the call site.

## Fix

- `InterfaceImplementability.CanBeImplemented` — an interface is mockable only when every abstract member it (or a base interface) requires an implementer to provide is accessible from the consuming compilation. Members with a default implementation are ignored; they are not the implementer's problem.
- `BuildSingleTypeModel` returns null for those interfaces, so they are dropped from every discovery path — direct, `[GenerateMock]`, and the transitive walk.
- `GetAutoMockFactoryMethod` applies the same rule, so a member returning such an interface falls back to a plain default rather than referencing a factory that was deliberately not generated (which would be CS0400).
- New `TM007` + `InaccessibleInterfaceMemberMockAnalyzer` reports at the call site when the interface is mocked directly, naming the member responsible — replacing the CS0535 cascade with one actionable error.

## Tests

- `tests/TUnit.Mocks.SourceGenerator.Tests/Issue6491Tests.cs` — the reported shape end-to-end: an external assembly whose `IDocumentStore` exposes an interface with an `internal` abstract member. Asserts no CS0535/CS0548/CS0551/CS0122 in the generated compilation, that the unimplementable interface gets no impl, and that a sibling public interface still gets its auto-mock factory.
- `tests/TUnit.Mocks.Analyzers.Tests/InaccessibleInterfaceMemberMockAnalyzerTests.cs` — 8 cases covering inaccessible property / method / inherited member, both entry points, and the non-reporting cases (all-public members, default-implemented non-public member, same-assembly `internal`, class targets).

Analyzer tests 38 passed, generator snapshots 82 passed, `TUnit.Mocks.Tests` 1160 passed, Http 54, Logging 31.


thomhurst:
Both findings confirmed and fixed in 4bb8a2b.

**1. `protected` false positive — you're right, and it was a regression.** I probed it with a two-assembly compilation before changing anything:

```
PROT_IMPL_ERRORS: <none>          // class ProtImpl : ExternalLib.IProt with explicit impl, compiles clean
IProt.Hidden acc=Protected vsAssembly=False vsImplType=True
IInt.Hidden  acc=Internal  vsAssembly=False vsImplType=False
```

So `IsSymbolAccessibleWithin(member, assembly)` was asking the wrong question, and as written the PR would have dropped every interface with a `protected` abstract member out of generation.

Rather than special-casing the accessibility kinds inline, `InterfaceImplementability` now delegates to `MemberDiscovery.IsMemberAccessible` (made `internal`). That is the rule that already decides which members reach the model, so the predicate is exact by construction: a required member discovery drops is precisely a member the impl will fail to implement. It also picks up the signature-type check for free — an abstract member whose parameter type is inaccessible has the same failure mode and wasn't covered before.

**2. Multi-type combos — confirmed.** `Mock.Of<T1, T2>()` now bails out when any type argument is unimplementable (matching what `BuildSingleTypeModel` returning null already did for the primary), and `ResolveMockTargets` in the analyzer walks all type arguments, so TM007 names the offending interface instead of leaving a raw CS0535.

**3. Tests.** You were right that the old `protected`-based tests couldn't stand in for the real thing — and after fix #1 they'd have been asserting the opposite of correct behaviour. Added `VerifyAnalyzerWithLibraryAsync` (Microsoft.CodeAnalysis.Testing `AdditionalProjects`) so every reporting case now uses a genuinely separate compilation with `internal` members, plus explicit non-reporting cases for cross-assembly `protected` and `protected internal`. Generator side gained `Interface_With_Protected_Members_Is_Still_Mockable`.

**On extracting the shared accessibility logic:** done on the generator side (single call into `MemberDiscovery.IsMemberAccessible`). The analyzer still carries its own ~12-line copy, because the analyzer and generator ship as separate assemblies with no shared project — the generator's sources are also linked into three Roslyn-version variants, so introducing a shared compilation unit is a build-layout change I'd rather not land inside a bug-fix PR. It's called out in an XML doc on both sides, and the two protected/protected-internal cases now pin the behaviour from the analyzer's end. Happy to do the extraction as a follow-up if you'd prefer it tracked.

Analyzer tests 41 passed, generator tests 84 passed, `TUnit.Mocks.Tests` 1160, Http 54, Logging 31.

thomhurst:
CI note — the macOS failure on ad22d9b5 was a latent fragility in the snapshot harness, not the diagnostic logic. Fixed in 5e42fac.

Only the two `[assembly: GenerateMock]`-driven tests failed (`KiotaIRequestAdapter_ConstrainedGenericMethod_EmitsDefaultConstraint` and `GenerateMock_Attribute_With_Concrete_Class`); the other 82 passed, and all 84 pass locally on Windows in both Debug and Release. The reference set is the only platform-varying input to those tests, and it had two order-dependent sources:

- `AppDomain.CurrentDomain.GetAssemblies()` only returns assemblies the runtime has already faulted in, so whether `TUnit.Mocks` is referenced depends on which test ran first — and this PR adds four tests, which changes that order.
- the `ref/TUnit.Mocks.dll` copy comes from a `<None Include>` whose glob expands at *evaluation* time, so a clean build that evaluates this project before `TUnit.Mocks`/netstandard2.0 has produced output ships without it.

With no `TUnit.Mocks` reference, `GenerateMockAttribute` doesn't bind, the generator emits nothing, and it surfaces as a snapshot mismatch instead of a missing reference — which is why it read as a logic regression.

`LoadReferences` now falls back to the assembly the project actually references, and throws naming the real cause if neither source yields one. Verified by deleting the `ref` copy and re-running: 84 passed with and without it.

The 6504 windows failure was separate and unrelated to mocks — `RunEngineTestsModule`/`ExpectedStateTests` with `TaskCanceledException`; neither `TUnit.Engine.Tests` nor `TUnit.TestProject` references TUnit.Mocks. Re-run in flight.

PR #6503, opened 2026-07-28
feat(mocks): accept an async factory in Returns() on async members

### LINKED ISSUE #6495: [TUnit.Mocks] No way to make an async member's .Returns() genuinely pending (only synchronous Func<T> supported) - breaks timeout-race tests
## Description

`.Returns()` only accepts a **synchronous** `Func<T>` for computed/delayed values on an async-returning member — there doesn't appear to be any way to have a mocked async call return a `Task<T>`/`ValueTask<T>` that stays genuinely pending and completes only later (e.g. after a real `await Task.Delay(...)`). This makes it impossible to write a test that races a mocked async call against an external timeout, which is a common pattern for testing timeout-handling logic.

## Repro

```csharp
using TUnit.Mocks;

var mock = ISomeInterface.Mock(); // where SomeMethodAsync() : Task<int>

_ = mock.SomeMethodAsync().Returns(async () =>
{
    await Task.Delay(1000);
    return 42;
});
```

```
error CS4010: Cannot convert async lambda expression to delegate type 'Func<int>'. An async lambda
expression may return void, Task or Task<T>, none of which are convertible to 'Func<int>'.
```

The `.Returns()` overload set apparently only has a `Func<T>` (synchronous) computed-value overload — there is no `Func<Task<T>>`/`Func<ValueTask<T>>` overload that would let the returned task itself be a real, still-pending task.

## What we tried instead (doesn't work for this use case)

Replacing the async delay with a **synchronous blocking** delay:

```csharp
_ = mock.SomeMethodAsync().Returns(() =>
{
    Thread.Sleep(1000);
    return 42;
});
```

This compiles, but changes the semantics in a way that breaks timeout-racing tests: production code that does something like `await Task.WhenAny(mock.SomeMethodAsync(), Task.Delay(timeout))` never gets a chance to race, because the mocked call itself doesn't return an (incomplete) task until *after* the full synchronous delay has already elapsed — by the time `Task.WhenAny` receives the "actual call" task, it's already complete, so the timeout branch can never win regardless of how short `timeout` is.

## Environment

## Impact / ask

Is there a supported way to configure a mocked async member to return a task that only completes after a real asynchronous delay (so external code racing it against a timeout/cancellation behaves correctly)? If not, could `.Returns()` gain an overload accepting `Func<Task<T>>` / `Func<ValueTask<T>>` used as-is (not further wrapped), so the delay is genuinely asynchronous rather than synchronously blocking the caller?

## Context

Found while migrating an OSS health-checks library (dailydevops/healthchecks) from NSubstitute to TUnit.Mocks across ~65 test files. Several of our health checks have a "the underlying client is too slow, so we should time out and report Degraded/Unhealthy" test, and NSubstitute's `.Returns(async _ => { await Task.Delay(x); return y; })` pattern was used throughout for this. We ended up keeping NSubstitute for one such test (RabbitMQ's timeout test) specifically because of this gap.

thomhurst:
Fixes #6495

## Problem

```csharp
mock.SomeMethodAsync().Returns(async () => { await Task.Delay(1000); return 42; });
// error CS4010: Cannot convert async lambda expression to delegate type 'Func<int>'
```

`Returns` on an async member is generated over the *unwrapped* return type, so its factory overload is `Func<int>` — an async lambda has nowhere to bind. The shape that does compile, `Returns(() => { Thread.Sleep(1000); return 42; })`, blocks the caller and hands back an already-completed task, so `await Task.WhenAny(mock.SomeMethodAsync(), Task.Delay(timeout))` can never let the timeout branch win. That makes timeout-handling tests unwritable — the reporter kept NSubstitute for one test because of it.

`ReturnsAsync(Func<Task<T>>)` already did the right thing, but nothing at the `Returns` call site pointed there.

## Fix

Emit a `Returns` overload alongside every existing `ReturnsAsync` factory overload — the parameterless form in `EmitReturnsAsyncOverloads` and the typed-parameter form in `GenerateTypedReturnsAsyncOverload` — routing through the same `ReturnsRaw` path. The factory's task is handed back as-is, so it stays pending until it completes.

Covers `Task<T>`, `ValueTask<T>`, `Task` and `ValueTask` members. The synchronous `Func<T>` overload is untouched; `ReturnsAsync` is unchanged.

## Tests

`tests/TUnit.Mocks.Tests/Issue6495Tests.cs` — 8 cases: the timeout race (asserting the returned task is incomplete and that a 50ms timeout beats it), value propagation, per-call invocation, the typed-parameter form, `ValueTask<T>` and `Task` members, plus guards that the synchronous `Returns` factory and `ReturnsAsync` still behave as before.

`TUnit.Mocks.Tests` 1168 passed. Generator snapshots regenerated (8 files, additive only) and passing on net8.0/net9.0/net10.0. Analyzers 30, Http 54, Logging 31.

### DIFF
--- a/src/TUnit.Mocks.SourceGenerator/Builders/MockMembersBuilder.cs
+++ b/src/TUnit.Mocks.SourceGenerator/Builders/MockMembersBuilder.cs
@@ -720,9 +720,11 @@ internal static class MockMembersBuilder
         }
 
         // Returns alias, so an `async (a, b) => ...` lambda binds here rather than failing against
-        // the synchronous typed overload with CS4010. See issue #6495.
+        // the synchronous typed overload with CS4010. Deprioritised for the same reason as the
+        // parameterless alias — see EmitReturnsAsyncOverloads. See issue #6495.
         writer.AppendLine();
         writer.AppendLine("/// <summary>Configure a typed computed async return value using the actual method parameters. The returned task is handed back as-is, so an async factory stays pending until it completes.</summary>");
+        writer.AppendLine(PriorityMinusOneAttribute);
         using (writer.Block($"public {wrapperName} Returns({funcType} factory)"))
         {
             writer.AppendLine($"EnsureSetup().ReturnsRaw(args => (object?)factory({castArgs}));");
@@ -1791,7 +1793,14 @@ internal static class MockMembersBuilder
         // Same factory, spelled Returns — so an `async () => ...` lambda binds without the caller
         // having to know about ReturnsAsync (the synchronous Func<T> overload rejects it with
         // CS4010). The returned task is handed back as-is, so it stays pending. See issue #6495.
+        //
+        // Deprioritised against the synchronous Returns(Func<T>) sibling: when T is a reference
+        // type, a lambda whose body pins nothing — Returns(() => null), Returns(() => throw ...) —
+        // converts equally well to Func<T> and Func<Task<T>>, which would be CS0121. The priority
+        // breaks that tie back to the pre-existing synchronous meaning. A genuine async lambda is
+        // unaffected: it is not convertible to Func<T> at all, so it is the only candidate.
         writer.AppendLine($"/// <summary>Return a {taskLabel} from a factory, invoked on each call. The {taskLabel} is returned as-is, so an async factory stays pending until it completes.</summary>");
+        writer.AppendLine(PriorityMinusOneAttribute);
         writer.AppendLine($"public {wrapperName} Returns(global::System.Func<{taskType}> taskFactory) {{ EnsureSetup().ReturnsRaw(() => (object?)taskFactory()); return this; }}");
     }
 
--- a/src/TUnit.Mocks.SourceGenerator/Builders/MockMembersBuilder.cs
+++ b/src/TUnit.Mocks.SourceGenerator/Builders/MockMembersBuilder.cs
@@ -720,9 +720,11 @@ internal static class MockMembersBuilder
         }
 
         // Returns alias, so an `async (a, b) => ...` lambda binds here rather than failing against
-        // the synchronous typed overload with CS4010. Deprioritised for the same reason as the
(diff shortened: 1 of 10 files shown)


thomhurst:
Both reviews are right — confirmed the ambiguity before fixing it. With the new overloads and no tie-break, three of the new reference-typed cases fail to compile:

```
Issue6495Tests.cs(134,29): error CS0121: The call is ambiguous between
  '...GetNameAsync_M0_MockCall.Returns(System.Func<string?>)' and
  '...GetNameAsync_M0_MockCall.Returns(System.Func<System.Threading.Tasks.Task<string?>>)'
```

Fixed in bb4843d by putting `[OverloadResolutionPriority(-1)]` on the async factory overloads (both the parameterless and typed-parameter forms) rather than narrowing which members get the alias. The priority only breaks ties among *applicable* candidates, so:

- `Returns(() => null)` / `Returns(() => throw ...)` — both overloads applicable, priority sends it back to the synchronous `Func<T>` and its pre-existing meaning (null is the *value*; the member still returns a completed task).
- `Returns(async () => ...)` — an async lambda is not convertible to `Func<T>` at all, so the task overload is the only candidate and the feature still works.

This keeps the fix uniform across all async members instead of restricting it to non-nullable value types, which would have left the reference-typed case — the more common one for a mocked service — still needing `ReturnsAsync`.

`[OverloadResolutionPriority]` ranks only within a single containing type, which is satisfied here: both overloads are emitted into the same generated call wrapper. TUnit.Mocks already requires C# 14 (TM004), so the Roslyn ≥4.12 floor for honouring it is guaranteed.

Added to `Issue6495Tests.cs` against a new `IReferenceResultService` (`Task<string?>` members): null-returning lambda, throwing lambda, null-returning typed lambda, and an async lambda on a reference-typed result asserting it still stays pending. `TUnit.Mocks.Tests` 1172 passed; snapshots regenerated and green on net8.0/net9.0/net10.0.

PR #6504, opened 2026-07-28
fix(mocks): emit the setup surface into the globally-imported namespace

### LINKED ISSUE #6494: [TUnit.Mocks] Generated mock wrapper type names collide across namespaces (e.g. two different ICluster interfaces) - one silently loses its members
## Description

`T.Mock()` generates a wrapper type named after the mocked type's *short name only* (e.g. `IClusterMock` for an interface named `ICluster`), without namespace-qualifying the generated type name. When two distinct types with the same short name (but different namespaces) are both mocked in the same compilation, this causes a collision: one of the two generated wrappers ends up missing the real interface's members entirely, with no compiler error pointing at the actual cause.

## Repro

```xml
<PackageReference Include="CassandraCSharpDriver" Version="3.22.0" />
<PackageReference Include="CouchbaseNetClient" Version="3.9.4" />
```

```csharp
using TUnit.Mocks;
using CassandraDriver = global::Cassandra;

var couchbaseCluster = Couchbase.ICluster.Mock();

var cluster = CassandraDriver.ICluster.Mock();
var session = CassandraDriver.ISession.Mock();
_ = cluster.ConnectAsync().Returns(session); // fails
```

Both `Cassandra.ICluster` and `Couchbase.ICluster` are interfaces named `ICluster` in different namespaces. Mocking either one alone works fine. Mocking **both** in the same compilation breaks `Cassandra.ICluster`'s generated wrapper (in this ordering) — its `ConnectAsync` setup method is simply absent from the wrapper, even though the interface itself does declare it and mocking it in isolation exposes it correctly.

## Environment

- CassandraCSharpDriver: 3.22.0
- CouchbaseNetClient: 3.9.4
- TFM: net10.0
- OS: Windows 11

## Actual error

```
Program.cs(8,13): error CS1061: 'IClusterMock' does not contain a definition for 'ConnectAsync',
and no accessible extension method 'ConnectAsync' accepting a first argument of type 'IClusterMock'
could be found (are you missing a using directive or an assembly reference?)
```

Note there is no error/diagnostic pointing at a name collision — the symptom looks exactly like "this interface member isn't supported," which is misleading; the interface is fine in isolation.

## Suspected cause

The source generator's naming scheme for generated mock wrapper types appears to use only the mocked type's simple name (`ICluster` → `IClusterMock`) rather than a namespace- or assembly-qualified name (e.g. `Cassandra_ICluster_Mock` / `Couchbase_ICluster_Mock`, matching the naming convention already used for the generated *file* names, e.g. `Cassandra_ICluster_Mock.g.cs`). With two same-named interfaces mocked in one compilation, this produces two conflicting partial/duplicate type declarations, and one silently loses.

## Impact

Any project that mocks two same-named types (interfaces or classes) from different libraries/namespaces in the same test assembly will have one of them silently broken, with an error message that looks like an unrelated "member not supported" issue rather than a naming collision — likely to cost other users significant debugging time, as it did us.

## Context

Found while migrating an OSS health-checks library (dailydevops/healthchecks) from NSubstitute to TUnit.Mocks across ~65 test files — we mock both `Cassandra.ICluster` (Cassandra/CassandraHealthCheckTests.cs) and `Couchbase.ICluster` (Couchbase/CouchbaseHealthCheckTests.cs) in the same test project.

thomhurst:
Fixes #6494

## Problem

```
error CS1061: 'IClusterMock' does not contain a definition for 'ConnectAsync', and no accessible
extension method 'ConnectAsync' accepting a first argument of type 'IClusterMock' could be found
```

The report attributed this to a name collision between `Cassandra.ICluster` and `Couchbase.ICluster`. It isn't — it reproduces with a single mock and no collision:

```csharp
namespace Some.Library { public interface ICluster { string Bucket { get; } } }

// no `using Some.Library;`
var b = Some.Library.ICluster.Mock();   // binds — Mock() lives in namespace TUnit.Mocks
b.Bucket.Returns("bucket");             // CS1061
```

The `_MockMemberExtensions` / `_MockEventsExtensions` classes were emitted beside the mocked type. Extension members are only found when their containing namespace is imported, and the `Mock()` entry point is in `TUnit.Mocks` (a global using) so the call site compiles — leaving a mock whose entire setup/verify surface is invisible. The resulting error reads as "this member isn't supported", which is what cost the reporter debugging time.

## Fix

Emit the extension classes, the call wrappers they return, and the `Events` surface into `TUnit.Mocks.Generated` — added as a global using by `TUnit.Mocks.targets`. Those type names are already derived from the fully qualified type name, so two mocked types sharing a short name across namespaces remain distinct in the flat namespace.

Unchanged: the impl, factory, bridge and `IFooMock` wrapper types keep their current placement beside the mocked type, as do the out/ref setter delegates (named from the short name and referenced by the impl through `GetGlobalMockNamespacePrefix`), which now emit in their own namespace block.

Docs note added for projects that set `TUnitMockImplicitUsings=disable`.

## Tests

`tests/TUnit.Mocks.Tests/Issue6494Tests.cs` — member setup, verification and event raising against interfaces in namespaces the test deliberately does not import, including the reported two-`ICluster` shape.

`TUnit.Mocks.Tests` 1164 passed. Generator snapshots regenerated (59 files — namespace wrapper and re-indentation only) and passing on net8.0/net9.0/net10.0. Analyzers 30, Http 54, Logging 31.


thomhurst:
Addressed the snapshot-coverage nit in 1fbc04d. Since the delegate emission is the exact path this PR split out of the members loop, it was worth closing rather than deferring: `Interface_With_NonSpan_RefStruct_Out_Ref_Parameters` pins both halves of the split — the `OutSetter`/`RefSetter` delegates staying in the mocked type's namespace (where the impl references them by short name) and the setup surface landing in `TUnit.Mocks.Generated`. Snapshots now 81, green on net8.0/net9.0/net10.0.

thomhurst:
On the flat-namespace collision finding (`A_B.IFoo` vs `A.B.IFoo`): the collision is real but pre-existing, and not a regression from this PR.

`SanitizeIdentifier` isn't injective, and its output is used for the `AddSource` hint names — so the two types collide at the hint-name level before either surface is emitted. Generation dies the same way on `main` as on this branch, and it dies *silently*:

```
PROBE: generator produced 1 files, no exception       // only the post-init TUnit.Mocks.Generated file
```

Mocking either type alone yields 5 files. Same result on `main` and on this branch, so moving the surface into `TUnit.Mocks.Generated` doesn't change it — the duplicate-type error the finding predicts can't happen, because the second file is never added.

Filed as #6505 rather than folded in here: the fix is to make the identifier mapping injective (escape `_` before replacing separators), which renames generated types and files across the board and needs a full snapshot refresh — not something to bury in a namespace-placement fix. The silent-failure aspect is the more interesting half of that issue.

thomhurst:
Rebased onto main to clear the conflict. All 8 conflicts were generated-code snapshots: #6503 (async `Returns` factories) added overloads to the same call-wrapper types this branch moves into `TUnit.Mocks.Generated`, so the two edits collided inside the same generated blocks.

Regenerated rather than hand-merged them (426c1b1) — the diff against the pre-rebase snapshots is additions-only and entirely #6503's `Func<Task<T>>`/`Func<ValueTask<T>>` `Returns` overloads, now emitted inside the relocated types. Snapshot suite 85/85 and `TUnit.Mocks.Tests` 1182/1182 green locally on net10.0.

PR #6507, opened 2026-07-28
fix(mocks): make generated identifier sanitization injective

### LINKED ISSUE #6505: [TUnit.Mocks] Two mocked types whose sanitized names collide silently generate nothing (e.g. A_B.IFoo and A.B.IFoo)
## Description

`MockImplBuilder.SanitizeIdentifier` maps every non-alphanumeric character to `_` and collapses runs of underscores, so it is not injective: `A_B.IFoo` and `A.B.IFoo` both sanitize to `A_B_IFoo`.

That string is used for the `AddSource` hint names (`GetSafeFileName` → `GetCompositeSafeName`), so mocking both types in one compilation produces duplicate hint names. The result is **silent**: no mock is generated for *either* type, and no diagnostic is reported.

## Repro

```csharp
using TUnit.Mocks;

namespace A_B { public interface IFoo { void Go(); } }
namespace A.B { public interface IFoo { void Stop(); } }

public class Test
{
    public void Run()
    {
        var one = Mock.Of<A_B.IFoo>();
        var two = Mock.Of<A.B.IFoo>();   // both silently produce nothing
    }
}
```

Running the generator over that source yields only the post-initialization `TUnit.Mocks.Generated` namespace file — 1 file total. Mocking either type alone yields 5. Confirmed against `main`.

The user-visible symptom is whatever the missing surface causes downstream (CS1061 on every setup, or CS0117 on `T.Mock()`), with nothing pointing at the cause.

## Suspected fix

Make the identifier mapping injective — e.g. escape existing underscores before replacing separators (`_` → `__`, then `.` → `_`), so `A.B.IFoo` → `A_B_IFoo` and `A_B.IFoo` → `A__B_IFoo`. This touches every generated type and file name, so it needs a snapshot refresh; only fixtures whose type or namespace names already contain `_` actually change.

A cheaper stopgap would be to detect the duplicate hint name and report a diagnostic, so the failure stops being silent.

## Context

Found while reviewing #6494 / PR #6504 — flagged there as a possible regression from moving the setup surface into a flat namespace, but it reproduces identically on `main`, so it is pre-existing and independent of that change.

thomhurst:
Fixes #6505

## Problem

`SanitizeIdentifier` mapped every non-alphanumeric character to `_` and collapsed runs, so it was not injective: `A_B.IFoo` and `A.B.IFoo` both sanitized to `A_B_IFoo`.

That string is the `AddSource` hint name (via `GetSafeFileName` → `GetCompositeSafeName`), so mocking both types in one compilation produced duplicate hint names and Roslyn dropped the generated sources for **both** — with no diagnostic. The user only saw the downstream symptom (CS1061 on every setup, or CS0117 on `T.Mock()`), with nothing pointing at the cause.

## Fix

Double literal underscores before mapping separators (`_` → `__`, separators → `_`), so a single `_` in the result always came from a separator:

- `A.B.IFoo` → `A_B_IFoo`
- `A_B.IFoo` → `A__B_IFoo`

Runs of separators still collapse (`IFoo<T>` → `IFoo_T_`); only the separator/underscore distinction has to survive.

The two byte-identical copies of the routine (`MockImplBuilder` and `TypeSymbolExtensions`) now delegate to one implementation in `IdentifierEscaping`, so they can't drift.

## Snapshot churn

Only names that already contain an underscore change. Three snapshots were refreshed — all multi-interface mocks, whose composite hint name is built by joining with `_` before sanitizing and so now gains a doubled underscore at the join. The **file contents are identical**; only the hint-name sort order of the emitted files moved. (Keeping the join pre-sanitize is deliberate: joining already-sanitized parts with a plain `_` would let `A.B` + additional interface `C` collide with the type `A.B.C`.)

## Tests

New `Issue6505Tests` — verified to fail on `main` and pass here:
- both colliding types generate their member surfaces
- mocking both drops no generated file (count equals the sum of the individual runs, minus the shared post-init stub)

Full runs, all green: `TUnit.Mocks.SourceGenerator.Tests` (87), `TUnit.Mocks.Tests` (1182), `TUnit.Mocks.Http.Tests` (54), `TUnit.Mocks.Logging.Tests` (31), `TUnit.Mocks.Analyzers.Tests` (56).


thomhurst:
Confirmed — `A_.B.IFoo` and `A._B.IFoo` both sanitize to `A___B_IFoo`. Good catch.

That case can't be fixed by a better escape, though: the output alphabet is `[A-Za-z0-9_]`, and if both a separator and a literal underscore render as runs of `_`, a run only carries a *sum*, not an order — three underscores is `__` + `_` or `_` + `__` either way. Making it injective means escaping separators too (`A.B.IFoo` → `A__B__IFoo`), which uglifies every generated file and extension-class name for a namespace shape nobody writes.

So instead the residual case is now detected and reported rather than dropped (pushed in f44a8fa):

- The dedupe step already sees every model, so it groups them by generated name and flags collisions before anything is written.
- Colliding types are skipped and **TM008** is reported, naming both types and the name they share.
- Unrelated mocks in the same compilation keep generating — the old failure mode was a duplicate hint name aborting the generator, which killed *every* mock and reported only a CS8785 warning against the generator.

Grouping keys on the secondary-surface flag (a multi-interface combo and its pair setup surface deliberately share a composite name and are told apart by the hint-name suffix) and ignores models that are the same target mocked in different modes, so neither is a false positive.

Three new tests cover it: the sanitizer's separator/underscore distinction, TM008 on the adjacent-token case, and an unrelated mock surviving alongside a collision. Full suites still green — SourceGenerator 90, Mocks 1182, Http 54, Logging 31, Analyzers 56.

Issue #6514, opened 2026-07-29
[TUnit.Mocks] Get<T>() can't be configured for T inaccessible to the calling assembly (unlike NSubstitute's runtime-proxy auto-recursion)


samtrion:
## Summary

When migrating from NSubstitute to `TUnit.Mocks`, I hit a case where a mocked type's generic method is called *internally* by a third-party SDK with a type argument that our assembly cannot name — an `internal` type in that SDK. NSubstitute can still handle this transparently; `TUnit.Mocks` structurally cannot, because it's a compile-time source generator running in the consuming assembly rather than a runtime proxy.

## Context

`Microsoft.Azure.Functions.Worker.FunctionContext.Features` is an `IInvocationFeatures`, whose `Get<T>()` is called *inside the Worker SDK itself* (via `context.GetInvocationResult()`) with `T = Microsoft.Azure.Functions.Worker.Context.Features.IFunctionBindingsFeature` — an `internal` type. Our test assembly has no `InternalsVisibleTo` grant to that SDK assembly, so we cannot even write `IFunctionBindingsFeature` in our code, let alone configure a mock for it.

## What NSubstitute does that works

```csharp
using NSubstitute;

var features = Substitute.For<IInvocationFeatures>();
// We never configure Get<IFunctionBindingsFeature>() at all - we can't even name the type.
// But when the Worker SDK internally calls features.Get<IFunctionBindingsFeature>(),
// NSubstitute's Castle DynamicProxy transparently returns an auto-generated substitute
// for ANY T, regardless of its accessibility to our assembly - because Castle's proxy
// assembly (DynamicProxyGenAssembly2) is one of the assemblies the Worker SDK grants
// InternalsVisibleTo, while our own assembly is not.
context.Features.Returns(features);
```

This works because NSubstitute's substitution happens at *runtime*, inside an assembly the target SDK already trusts.

## What TUnit.Mocks does instead

```csharp
using TUnit.Mocks;

var features = IInvocationFeatures.Mock();
context.Features.Returns(features);

// Later, when Microsoft.Azure.Functions.Worker's own extension method internally calls:
//   features.Get<IFunctionBindingsFeature>()
// TUnit.Mocks generated Get<T>() overloads only for T's our compile-time code actually
// asked to configure. For any other T - especially one we structurally cannot name because
// it's internal to another assembly - the call falls through to a plain unconfigured
// default (null), and there is no API to register "whatever T is asked for, return X"
[...]

PR #6520, opened 2026-07-30
feat(mocks): experimental compile-time internals access (#6514 Tier 2)

thomhurst:
## Summary

The "Tier 2" follow-up to #6519's runtime auto-stubs for #6514, now fully productionized. Behind `TUnitMocksExperimentalInternalsAccess`, internal types of selected referenced assemblies become **first-class mockable** — nameable in test code, source-generated typed mocks, setups, matchers, verification — with **zero `InternalsVisibleTo` required** from the target assembly. That exceeds runtime-proxy libraries: NSubstitute/Moq can auto-substitute such types (when the SDK grants IVT to Castle) but can never let the test *configure* one, because the test cannot write the type's name.

```xml
<PropertyGroup>
  <TUnitMocksExperimentalInternalsAccess>true</TUnitMocksExperimentalInternalsAccess>
</PropertyGroup>
<ItemGroup>
  <TUnitMocksInternalsAccess Include="Microsoft.Azure.Functions.Worker.Core" />
</ItemGroup>
```

```csharp
var bindings = IFunctionBindingsFeature.Mock();          // internal to the SDK
features.Get<IFunctionBindingsFeature>().Returns(bindings.Object);
features.Get<IFunctionBindingsFeature>().WasCalled(Times.Once);
```

## Mechanism

The established publicizer pattern (Krafs.Publicizer / IgnoresAccessChecksToGenerator), wired into the TUnit.Mocks package:

1. `PublicizeAssemblyReferences` (Mono.Cecil MSBuild task) rewrites each selected reference so internals are public, preserving assembly identity (name/version/public key). The **implementation** assembly is the rewrite source — Roslyn ref assemblies strip internal members when no IVT exists (internal ctors vanish), which surfaced as TM006 on internal classes during test hardening.
2. The swap targets `ReferencePathWithRefAssemblies` — the compiler's view and nothing else. `ReferencePath`, copy-local, and `deps.json` keep the originals; the runtime loads the real assembly, where the task-emitted `IgnoresAccessChecksTo` makes the compiled IL (generated mock classes implementing internal interfaces included) valid.
3. **Zero source-generator changes** — through the publicized reference the types simply look public.

## Productionization

- **Packaging**: task multi-targets net472 (VS MSBuild) + net8.0 (dotnet msbuild), selected via `$(MSBuildRuntimeType)`; ships with Mono.Cecil under `tasks/<tfm>/`; `TUnit.Mocks.InternalsAccess.targets` packed to `buildTransitive/<tfm>/` and imported by `TUnit.Mocks.targets` — consumers need only the opt-in. Fully inert otherwise (full `TUnit.Mocks.Tests` suite unaffected: 1186/1186).
- **Tests** (10/10, Debug and Release/NoBuild CI-mode): end-to-end pipeline against a **strong-named** stand-in SDK with zero IVT grants — internal interface, closed-generic internal interface, internal class **partial mock** (internal ctor), typed setups/matchers/verification of calls made *inside* the SDK, manual implementations (type-load proof) — plus unit tests of the task itself: publicizing + identity preservation down to the public-key token, incrementality, generated-source content, `TUMIA001` for unresolved names.
- **Package-level verification**: packed the real nupkg, extracted it, and consumed the `buildTransitive` targets from a standalone console implementing an internal interface — works, **including under `PublishTrimmed` (full trim mode)**.
- **CI**: projects in `TUnit.CI.slnx`; `RunMockInternalsAccessTestsModule` runs the suite.
- **Docs**: new *Internals Access (experimental)* section in the mocking docs.

## Remaining caveats

- `IgnoresAccessChecksToAttribute` is CoreCLR-honored but not a documented contract (foundation of several long-lived OSS packages; low-but-nonzero risk).
- .NET Framework test targets: inert with warning `TUMIA002` (the runtime doesn't honor the attribute).
- Native AOT unverified (no MSVC linker on the dev machine); trimmed publish verified.
- Publicizer scope: types + methods (fields untouched; explicit interface impls stay private as IL requires).

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -54,6 +54,7 @@
     <PackageVersion Include="Microsoft.Testing.Platform.MSBuild" Version="2.3.3" />
     <PackageVersion Include="Imposter" Version="0.1.9"/>
     <PackageVersion Include="Mockolate" Version="3.4.0" />
+    <PackageVersion Include="Mono.Cecil" Version="0.11.6" />
     <PackageVersion Include="Moq" Version="4.20.72" />
     <PackageVersion Include="System.Threading.Channels" Version="9.0.0" />
     <PackageVersion Include="ModularPipelines.DotNet" Version="3.2.8" />
--- a/TUnit.CI.slnx
+++ b/TUnit.CI.slnx
@@ -27,6 +27,7 @@
     <Project Path="src/TUnit.Mocks/TUnit.Mocks.csproj" />
     <Project Path="src/TUnit.Mocks.Assertions/TUnit.Mocks.Assertions.csproj" />
     <Project Path="src/TUnit.Mocks.Http/TUnit.Mocks.Http.csproj" />
+    <Project Path="src/TUnit.Mocks.InternalsAccess.Tasks/TUnit.Mocks.InternalsAccess.Tasks.csproj" />
     <Project Path="src/TUnit.Mocks.Logging/TUnit.Mocks.Logging.csproj" />
     <Project Path="src/TUnit.Playwright/TUnit.Playwright.csproj" />
   </Folder>
@@ -76,6 +77,8 @@
     <Project Path="tests/TUnit.Engine.Tests/TUnit.Engine.Tests.csproj" />
     <Project Path="tests/TUnit.Mocks.Analyzers.Tests/TUnit.Mocks.Analyzers.Tests.csproj" />
     <Project Path="tests/TUnit.Mocks.Http.Tests/TUnit.Mocks.Http.Tests.csproj" />
+    <Project Path="tests/TUnit.Mocks.InternalsAccess.TargetLib/TUnit.Mocks.InternalsAccess.TargetLib.csproj" />
+    <Project Path="tests/TUnit.Mocks.InternalsAccess.Tests/TUnit.Mocks.InternalsAccess.Tests.csproj" />
     <Project Path="tests/TUnit.Mocks.Logging.Tests/TUnit.Mocks.Logging.Tests.csproj" />
     <Project Path="tests/TUnit.Mocks.SourceGenerator.Tests/TUnit.Mocks.SourceGenerator.Tests.csproj" />
     <Project Path="tests/TUnit.Mocks.Tests/TUnit.Mocks.Tests.csproj" />
--- a/TUnit.slnx
+++ b/TUnit.slnx
@@ -27,6 +27,7 @@
     <Project Path="src/TUnit.Mocks/TUnit.Mocks.csproj" />
     <Project Path="src/TUnit.Mocks.Assertions/TUnit.Mocks.Assertions.csproj" />
     <Project Path="src/TUnit.Mocks.Http/TUnit.Mocks.Http.csproj" />
+    <Project Path="src/TUnit.Mocks.InternalsAccess.Tasks/TUnit.Mocks.InternalsAccess.Tasks.csproj" />
     <Project Path="src/TUnit.Mocks.Logging/TUnit.Mocks.Logging.csproj" />
     <Project Path="src/TUnit.Playwright/TUnit.Playwright.csproj" />
     <Project Path="src/TUnit.Reporting.Tool/TUnit.Reporting.Tool.csproj" />
@@ -77,6 +78,8 @@
     <Project Path="tests/TUnit.Engine.Tests/TUnit.Engine.Tests.csproj" />
     <Project Path="tests/TUnit.Mocks.Analyzers.Tests/TUnit.Mocks.Analyzers.Tests.csproj" />
     <Project Path="tests/TUnit.Mocks.Http.Tests/TUnit.Mocks.Http.Tests.csproj" />
+    <Project Path="tests/TUnit.Mocks.InternalsAccess.TargetLib/TUnit.Mocks.InternalsAccess.TargetLib.csproj" />
+    <Project Path="tests/TUnit.Mocks.InternalsAccess.Tests/TUnit.Mocks.InternalsAccess.Tests.csproj" />
     <Project Path="tests/TUnit.Mocks.Logging.Tests/TUnit.Mocks.Logging.Tests.csproj" />
     <Project Path="tests/TUnit.Mocks.SourceGenerator.Tests/TUnit.Mocks.SourceGenerator.Tests.csproj" />
     <Project Path="tests/TUnit.Mocks.Tests/TUnit.Mocks.Tests.csproj" />
--- a/docs/docs/writing-tests/mocking/advanced.md
+++ b/docs/docs/writing-tests/mocking/advanced.md
@@ -287,5 +287,8 @@ used by several long-lived OSS tools, wired into the TUnit.Mocks package.
 - Not supported on .NET Framework test targets (the runtime there does not honor the attribute);
   a build warning is emitted and the pipeline stays inert.
 - Works under trimmed publishes; Native AOT is not yet verified.
+- If another package already injects an `IgnoresAccessChecksToAttribute` definition into your
+  compilation (e.g. IgnoresAccessChecksToGenerator), suppress TUnit's copy with
+  `<TUnitMocksInternalsAccessEmitAttributeDefinition>false</TUnitMocksInternalsAccessEmitAttributeDefinition>`.
 - Internal APIs are internal for a reason: they can change in any release of the target package.
   Prefer public seams when they exist.
--- a/docs/docs/writing-tests/mocking/advanced.md
+++ b/docs/docs/writing-tests/mocking/advanced.md
@@ -238,3 +238,54 @@ mock.Invocations.Count; // 0 (history cleared)
 ```
 
 The `SetupAllProperties()` flag is preserved across resets.
+
+## Internals Access (experimental)
+
+Some SDKs route behavior through types that are `internal` to their own assembly — the classic
+example is `Microsoft.Azure.Functions.Worker`, whose `IInvocationFeatures.Get<T>()` is called
+inside the SDK with `T = IFunctionBindingsFeature`, a type your test assembly cannot even name.
+Runtime-proxy libraries can auto-substitute such types (when the SDK grants `InternalsVisibleTo`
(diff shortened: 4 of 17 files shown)


thomhurst:
All review findings addressed in 5ea2559:

**Metadata drop on reference swap (Greptile P1 / Codex P2) — valid, fixed.** The replacement item now carries the original reference's metadata via `CopyMetadataTo` (then re-overrides `Private`/`CopyLocal`), so `Aliases`/`EmbedInteropTypes` survive the swap. Test asserts both round-trip.

**Timestamp-only cache (Greptile P2 / Codex P2) — valid, fixed.** Incrementality is now content-based: a `.sig` sidecar records the resolved source path + SHA256; any content change (downgrade, equal-or-older timestamp replacement) re-publicizes. Sig files added to `FileWrites`. Test manipulates content with a backdated timestamp.

**Ref assembly as compile asset with no OriginalPath (Codex P1) — valid, fixed.** The task now takes the runtime/copy-local assets (`RuntimeCopyLocalItems` + `ReferenceCopyLocalPaths`); when the chosen source carries `ReferenceAssemblyAttribute` it publicizes the implementation asset instead, and warns `TUMIA003` when none exists. Both paths tested (Cecil-forged ref assembly).

**Cecil exception handling (Claude #1) — fixed.** Resolve+publicize wrapped per assembly; failures surface as `TUMIA005` with the exception detail at low verbosity instead of an unhandled stack trace. Tested against a non-PE file.

**Duplicate attribute definition (Claude #3) — fixed.** `TUnitMocksInternalsAccessEmitAttributeDefinition=false` suppresses the generated attribute class (the `[assembly:...]` applications remain). Documented in README + docs.

**Silent first-match ambiguity (Claude #4) — fixed.** Multiple distinct paths matching one simple name now warn `TUMIA004` listing all candidates; first still wins deterministically.

**Custom task vs Krafs.Publicizer dependency (Claude #2)** — README now carries the rationale: those packages are consumer-facing MSBuild wiring, not library APIs; TUnit needs the step riding transparently inside its own package (single opt-in, buildTransitive, dual MSBuild hosts, RPWRA-only swap). Revisit if scope grows.

Suite: 17/17 including e2e targets pipeline.

thomhurst:
Round 3 addressed in 55e5049:

**Ambiguous-match removal incomplete (CS1703) — valid, fixed.** The task now outputs `SupersededReferences` — every same-simple-name match, winner included — and the targets removes all of them from `ReferencePathWithRefAssemblies` before adding the publicized copy. The ambiguous case now degrades to "first wins, cleanly". Test asserts both duplicate paths are superseded.

**Publicizer promoted private members (Codex P2) — valid, fixed.** Promotion is now scoped to assembly-visible surface only: `internal` / `protected internal` / `private protected` types and methods become public; `private` and `protected` stay untouched. A formerly-private `Describe(string)` beside a public `Describe(object)` can no longer capture `Describe(null)` in consuming code. Covered by a fixture class exercising all five visibilities plus nested types.

**IgnoresAccessChecksTo identity (minor) — fixed.** The attribute is emitted with the assembly identity read from the resolved file (`AssemblyName.GetAssemblyName`), not the file-derived requested name. Test renames the file and asserts the real identity is used.

Suite: 19/19 including e2e targets pipeline.

thomhurst:
Round 4 addressed in 05cdf1e:

**Duplicate opt-in items duplicate compiler references (Greptile P1) — valid, fixed.** Requested names are now deduped (OrdinalIgnoreCase) at the top of `Execute()`, so the same `TUnitMocksInternalsAccess` item included twice — multi-imported props included — yields exactly one publicized reference and one `IgnoresAccessChecksTo` application. Test: `Duplicate_Requests_Produce_Single_Publicized_Reference`.

**Non-winning ambiguous match loses compiler metadata (Greptile P1 / Claude edge case 1) — valid, fixed.** Since every same-simple-name match leaves `ReferencePathWithRefAssemblies`, metadata carried only by a non-selected match is now merged onto the replacement item: extern aliases are unioned (global visibility preserved when any match had no alias) and `EmbedInteropTypes` kept if any match set it. TUMIA004 text and README updated. Test: `Ambiguous_Matches_Union_Compiler_Metadata` — alias and interop flag live only on the loser, asserts `global,sdkalias` + `EmbedInteropTypes=true` on the publicized item.

**Cache signature missing publicizer version (Codex P2) — valid, fixed.** The `.sig` now leads with the task's informational version, so upgrading TUnit.Mocks re-publicizes even when the referenced assembly is byte-identical — rewrite-rule fixes are never masked by a stale `obj` copy. Test: `Task_Version_Change_Invalidates_Publicized_Copy` (tampered version line + backdated output → re-publicized).

Suite: 22/22 (task-level + e2e targets pipeline).

thomhurst:
Round 5 addressed in 0a55caf:

**Runtime-asset match by file name misses renamed compile assets (Codex P2) — valid, fixed.** When the compile asset is a ref assembly whose file name differs from its assembly identity, the implementation sits among the runtime assets under the *real* name, so the file-name-only predicate missed it and the stripped ref assembly got publicized (TUMIA003). `FindRuntimeImplementation` now reads the reference's real identity and matches runtime assets by identity name — version-exact match preferred, name-only as fallback — with the file-name match kept as the cheap fast path and the reference file itself excluded as a candidate. Unreadable/native dlls among the copy-local assets are skipped.

Test: `Renamed_Reference_Assembly_Resolves_Implementation_By_Assembly_Identity` — ref-assembly copy under a mismatched file name, implementation only findable by identity; asserts no warning and that the publicized output no longer carries `ReferenceAssemblyAttribute`.

Suite: 23/23.

thomhurst:
Round 6 addressed in 7837ea7:

**Same-name different-identity references collapsed (Greptile P1) — valid, fixed.** Matches are now partitioned by real assembly identity (name/version/public key): only the winner's exact identity is superseded and metadata-merged. A same-file-name reference with a different identity — a legal pair only when extern aliases keep the two apart — stays in the compiler's reference list untouched, so its alias never rebinds and its unique API never vanishes. Unreadable files are never assumed to be duplicates. TUMIA004 now lists retained references. Test: `Different_Identity_Same_Name_Reference_Is_Left_In_Place` (v99 Cecil-forged copy; asserts not superseded and alias not merged).

**Requested-name lookup file-name-only (Codex P2) — valid, fixed.** On a file-name miss the lookup now falls back to matching each resolved reference's real assembly identity, so the documented contract — the simple *assembly* name — works for renamed compile assets (`VendorAlias.dll` containing `Vendor.Core`) instead of failing TUMIA001. Test: `Requested_Name_Matches_Assembly_Identity_When_Compile_Asset_Is_Renamed`.

README diagnostics updated for both. Suite: 25/25.

thomhurst:
Both new findings verified and fixed in 2dc89b6:

**Retained identity stays inaccessible (Greptile P1) — valid.** A simple-name request cannot single out one of two extern-aliased same-name references, so retaining the second identity made it permanently unpublicizable. Each distinct assembly identity behind the requested name now gets its own publicized copy: same-identity duplicates still collapse onto one copy with metadata unioned, but aliases are never merged across identities — each identity's copy carries its own. One simple name still produces one IgnoresAccessChecksTo application (the runtime matches by simple name). Unreadable-identity matches are the only ones left in place, since they can't be safely classified.

**Output path collisions (Codex P2) — valid, and per-identity publicizing makes it reachable in one request.** Publicized copies now land in a per-source-path subdirectory (8-hex SHA-256 token of the full path), so same-file-name implementation assemblies from different directories can't overwrite each other. Path-derived, not content-derived, so the output location is stable across builds and the existing content-hash signature keeps doing incrementality.

Tests: identity-pair test asserts both copies exist with their own versions/aliases and distinct paths; new renamed-assets test covers two requests resolving to same-file-name sources. 26/26 including end-to-end.

thomhurst:
Both follow-up findings on 2dc89b6 verified and fixed in 6d4a0da72a:

**Filename-first runtime-asset resolution undoes per-identity publicizing (P2) — valid.** With two same-file-name metadata-only references under extern aliases, both identity groups searched the runtime assets and the filename shortcut handed each the first implementation — the second copy carried the wrong identity. Identity now decides (version-exact preferred, name-only fallback); the requested-name filename match survives only for a reference whose own identity cannot be read. Regression test builds a ref/impl pair per identity and asserts each publicized copy carries its own version.

**Path-token case folding collides on case-sensitive hosts (P2) — valid.** `ToUpperInvariant` is now applied only on Windows (`OperatingSystem.IsWindows()` on the net8.0 host; the net472 host is Windows-only), so /deps/A and /deps/a stay distinct sources on Unix while Windows keeps stable tokens across path-casing drift.

27/27 including end-to-end.

thomhurst:
Eighth round — all three findings verified and fixed in 26e2f55.

**Colliding file name hides the identity match (P2) — valid.** `FindReferenceMatches` was filename-first, so a resolved reference *named* `Vendor.Core.dll` but holding another identity blocked the identity-aware fallback from ever seeing the renamed asset that really contains `Vendor.Core`. Readable identities now decide first; file-name matching survives only for references whose identity cannot be read (they ride along as unclassifiable candidates) and as the no-identity-match compatibility fallback, which keeps the documented-name contract primary while preserving the request-by-file-name behavior the earlier renamed-asset test pinned. Identity reads are cached per invocation so the per-name scan doesn't reopen every reference. Regression test: a decoy with the requested file name plus a renamed asset holding the requested identity — the identity match wins.

**Version-only "exact" runtime match (P2) — valid.** The exact branch compared name+version but not the public key token, so a same-name-same-version candidate with a different PKT was handed back as the implementation. A conflicting PKT now excludes the candidate outright — it is a different assembly, and neither the exact branch nor the version-drift fallback will select it; the reference falls back to TUMIA003 + publicizing the stripped ref assembly, keeping its own identity. Test: signed ref assembly vs unsigned same-name-same-version runtime asset asserts the TUMIA003 path and that the publicized copy keeps the signed token.

**Unconditional case fold excluding the reference asset (P2) — valid.** The self-exclusion in runtime-candidate filtering used `OrdinalIgnoreCase` unconditionally; it now uses a platform-appropriate comparison (case folded only on Windows; the net472 host is Windows-only), same rule as the path-token hashing fixed last round.

Suite: 29/29 including end-to-end.

PR #6629, opened 2026-08-17
Fast-path unconstrained single-test runs

thomhurst:
## Summary

- fast-path one test with no dependencies or parallel constraints
- bypass cycle detection, executable copying, constraint grouping, and `Parallel.ForEachAsync`
- preserve static-property lifecycle, session hooks, and concurrent dynamic-test queue processing
- retain the normal scheduler for constrained, dependent, and multi-test runs

## Performance

31 paired alternating one-empty-test runs after 5 warmups, with Microsoft Testing Platform and .NET CLI telemetry disabled:

| Metric | Before | After | Change |
|---|---:|---:|---:|
| TUnit session median | 115.310 ms | 107.949 ms | **-7.361 ms (-6.38%)** |
| Paired session delta median | | | **-7.535 ms** |
| TUnit session mean | 119.711 ms | 107.866 ms | **-11.845 ms** |

CPU-sampling traces captured before/after with `dotnet-trace`.

## Validation

- dynamic test variant: 2 passed
- constrained fallback: 1 passed
- dependency path: 2 passed
- session/class/test hooks: 1 passed

### DIFF
--- a/src/TUnit.Engine/Scheduling/TestScheduler.cs
+++ b/src/TUnit.Engine/Scheduling/TestScheduler.cs
@@ -92,41 +92,55 @@ internal sealed class TestScheduler : ITestScheduler
         if (_logger.IsDebugEnabled)
             await _logger.LogDebugAsync($"Scheduling execution of {testList.Count} tests").ConfigureAwait(false);
 
-        var circularDependencies = _circularDependencyDetector.DetectCircularDependencies(testList);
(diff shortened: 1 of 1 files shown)


thomhurst:
Closing because this optimizes a single-test-only path, which is not representative of normal test-suite workloads. I'll replace it only if a general scheduler optimization proves faster across 10, 100, and 1,000-test suites.

PR #6677, opened 2026-08-26
fix: Avoid HTML report CLI option clashes

### LINKED ISSUE #6675: [Bug]: Clash of cli parameters with Microsoft.Testing.Extensions.HtmlReport
### Description

[Bug]: Clash of cli parameters with Microsoft.Testing.Extensions.HtmlReport:

Both packages have the same --report-html (I know deprecated now) and --report-html-filename.
Even though the option is deprecated, as both reporters try to register the same command, it hinders the process.

Your report is nicer BTW, but having the same parameter registered is "kinda" problematic. 

### Expected Behavior

Not having a conflict with an official package from Microsoft.

### Actual Behavior

Sorry in french :

L’option « --report-html » est déclarée par plusieurs fournisseurs : « TUnit', 'Générateur de rapports HTML »
L’option « --report-html-filename » est déclarée par plusieurs fournisseurs : « TUnit', 'Générateur de rapports HTML »

### Steps to Reproduce

Just reference both packages and try to get the reports.

In a random cs file : tunitexample.cs

#:package TUnit@1.65.63
#:package Microsoft.Testing.Extensions.HtmlReport@2.3.3

using TUnit;

public class TestClass
{
    [Test]
    public async Task TestMethod1()
    {
        await Assert.That(2 == 1 + 1).IsTrue();
    }

    [Test]
    public async Task TestMethod2()
    {
        await Assert.That(true).IsTrue();
    }
}

then : dotnet run tunitexample.cs -- --report-html --report-html-filename tunit.html

### TUnit Version

1.65.63

### .NET Version

.NET 10

### Operating System

Other (please specify in additional context)

### IDE / Test Runner

```shell
L’option « --report-html » est déclarée par plusieurs fournisseurs : « TUnit', 'Générateur de rapports HTML »
L’option « --report-html-filename » est déclarée par plusieurs fournisseurs : « TUnit', 'Générateur de rapports HTML »
```

### Additional Context

Whatever the system, same behavior

### IDE-Specific Issue?

mvanhorn:
## Description

Stop advertising the deprecated `report-html` option and rename TUnit’s filename option to a clearly namespaced `tunit-report-html-filename`, ensuring TUnit no longer claims either Microsoft-owned option name. Update `TestApplicationBuilderExtensions` to consume and validate the renamed constant while preserving the existing default-output behavior when no TUnit filename is supplied. Add focused coverage around `HtmlReporterCommandProvider.GetCommandLineOptions()` and its argument validation so the registered option set includes the TUnit-specific filename and excludes both conflicting legacy names.

TUnit registers `--report-html` and `--report-html-filename`, the same names registered by `Microsoft.Testing.Extensions.HtmlReport`, so Microsoft Testing Platform rejects applications that reference both reporters before tests can run. The issue includes a minimal .NET 10 reproduction using TUnit 1.65.63 and Microsoft’s HTML reporter 2.3.3. TUnit’s enable flag is already deprecated because its HTML report is generated by default, while the filename option remains the only TUnit behavior that needs an unambiguous CLI replacement. There are no prior closed attempts, claims, or competing pull requests in the supplied issue evidence.

Closes #6675

## Related Issue

Fixes #6675

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Checklist

Not applicable to this change.

### Required

- [x] **Dual-Mode Implementation**: If this change affects test discovery/execution, I have implemented it in BOTH:
    Not run: no test command resolved in this workspace, so nothing was executed to pass.
    Not run: no test command resolved in this workspace, so nothing was executed to pass.
    Not run: no test command resolved in this workspace, so nothing was executed to pass.
    Not run: no test command resolved in this workspace, so nothing was executed to pass.
    Not run: no test command resolved in this workspace, so nothing was executed to pass.
- [x] **Performance**: If this change affects hot paths (test discovery, execution, assertions):
    Not run: no test command resolved in this workspace, so nothing was executed to pass.
    Not run: no test command resolved in this workspace, so nothing was executed to pass.

### Testing

  Not run: no test command resolved in this workspace, so nothing was executed to pass.
- [x] I have tested both source-generated and reflection modes (if applicable)
- Construct TUnit's HTML command provider and verify it exposes exactly the TUnit-specific filename option, with one required argument, and does not expose `report-html` or `report-html-filename`. - Validate one filename argument for the renamed option and confirm it remains accepted by the provider. - Validate a missing or extra filename argument and confirm the provider rejects it with the existing single-output-path diagnostic.

## Additional Notes

Nothing beyond what is described above.

## Summary by CodeRabbit

* **New Features**
  * Added support for default and namespaced HTML report command-line options.
  * Use `--tunit-report-html-filename` when the HTML report extension is referenced.
  * Legacy options remain available in default mode with deprecation guidance.
* **Documentation**
  * Clarified HTML report configuration, filename behavior, disabling options, and compatibility guidance.
* **Tests**
  * Added coverage for namespaced filename configuration, option handling, and HTML report generation.


thomhurst:
Thanks for addressing the clash. This is still a breaking change: users who currently pass `--report-html` or `--report-html-filename` in CI will have their existing commands break when they update TUnit. We should defer removing or renaming existing CLI flags until v2. I have opened #6679 to track removal of deprecated CLI flags.

thomhurst:
Suggestion for implementing this without breaking existing TUnit-only users: choose the option names at build time from the registered MTP hooks, rather than using runtime reflection.

`Microsoft.Testing.Extensions.HtmlReport` contributes a well-known `@(TestingPlatformBuilderHook)` item (`A6E2BCC3-9B4D-4B6D-8AE3-2C1E12A54F4D`) through its `buildTransitive` props. TUnit already contributes its own hook and deliberately orders it last. TUnit could add an MSBuild target before `_GenerateSelfRegisteredExtensionsFileInputCache` that checks for the Microsoft HTML hook and changes TUnit's hook `TypeFullName` to an alternate public hook. That hook would call something like `AddTUnit(HtmlCliMode.Namespaced)`.

The resulting modes would be:

- Microsoft HTML hook absent: retain `--report-html` and `--report-html-filename`, preserving existing CI commands.
- Microsoft HTML hook present: do not register either conflicting name; register `--tunit-report-html-filename` for TUnit while Microsoft owns its original names.

This is AOT-safe because MTP-generated `SelfRegisteredExtensions.cs` makes a static call to the selected hook. It also handles transitive references and avoids reflection, `UnsafeAccessor`, and coupling to MTP's private `CommandLineManager._commandLineProviderFactory`.

Relevant Microsoft hook definition: https://github.com/microsoft/testfx/blob/44aa76e6a61d4908f06dfd77a51d4b7e3e7ce40f/src/Platform/Microsoft.Testing.Extensions.HtmlReport/buildMultiTargeting/Microsoft.Testing.Extensions.HtmlReport.props

Suggested coverage: TUnit-only legacy flags, both packages without duplicate-option failure, namespaced TUnit filename with both packages, and Native AOT for the combined-package case.

mvanhorn:
Thanks for spelling the design out, and for pointing at the Microsoft props file. That made it concrete. Implemented in 10a97f0.

- `TUnit.Engine.props` gains `_TUnitConfigureHtmlReportCommandLineOptions`, ordered `BeforeTargets="_GenerateSelfRegisteredExtensionsFileInputCache"`. It checks `@(TestingPlatformBuilderHook)` for `A6E2BCC3-9B4D-4B6D-8AE3-2C1E12A54F4D` and, when present, updates TUnit's hook `TypeFullName` to `TUnit.Engine.Framework.NamespacedHtmlReportTestingPlatformBuilderHook`.
- That hook calls `AddTUnit(HtmlCliMode.Namespaced)`. Both hooks are public and statically reachable, so the generated `SelfRegisteredExtensions.cs` makes a direct static call. No reflection, no `UnsafeAccessor`, no coupling to `CommandLineManager._commandLineProviderFactory`, and it stays AOT-safe.
- `HtmlReporterCommandProvider` takes the mode and registers accordingly: `Default` keeps `--report-html` and `--report-html-filename` byte-for-byte as before, `Namespaced` registers only `--tunit-report-html-filename` and claims neither Microsoft name.
- Docs updated to describe both modes rather than presenting the renamed flags as unconditional.

The important part is that existing TUnit-only CI commands are untouched. My earlier revision broke them, which was the right thing for you to reject.

On your four suggested test cases: the unit tests cover the option names in both modes. The three that actually need the MSBuild path exercised, both packages present without a duplicate-option failure, the namespaced filename with both packages, and Native AOT for the combined case, are not covered yet, because they need a test project that references `Microsoft.Testing.Extensions.HtmlReport` and a real build to prove the hook rewrite fires. I would rather add those deliberately than fake them at the unit level. Tell me where you would want that project to live and I will add it.

Verification so far is `dotnet build src/TUnit.Engine/TUnit.Engine.csproj` clean on SDK 10.0.400 to match `global.json`. I have not run the full suite.

src/TUnit.Engine/TUnit.Engine.props
```diff
@@ -44,4 +44,13 @@
         </TestingPlatformBuilderHook>
     </ItemGroup>
 
+    <Target Name="_TUnitConfigureHtmlReportCommandLineOptions"
+            BeforeTargets="_GenerateSelfRegisteredExtensionsFileInputCache">
+        <ItemGroup Condition="'@(TestingPlatformBuilderHook->WithMetadataValue('Identity', 'A6E2BCC3-9B4D-4B6D-8AE3-2C1E12A54F4D'))' != ''">
+            <TestingPlatformBuilderHook Update="6ADF853A-6945-4A06-9A4B-D99BC1DC1094">
```
thomhurst: 🔴 bug: This target changes `TypeFullName` on every `TestingPlatformBuilderHook`; combined app generates eight TUnit hook calls and crashes with `The test framework adapter factory has already been registered` before CLI parsing. Filter metadata assignment to TUnit identity while preserving other hooks.

tests/TUnit.Engine.Tests/HtmlReporterTests.cs
```diff
@@ -3,18 +3,81 @@
...
 {
+    [Test]
+    public void CommandProvider_DefaultMode_Exposes_Legacy_Options()
+    {
+        var provider = new HtmlReporterCommandProvider(new MockExtension());
+
+        var options = provider.GetCommandLineOptions();
+
+        options.Count.ShouldBe(2);
+        options.Single(x => x.Name == "report-html").Arity.ShouldBe(ArgumentArity.Zero);
```
thomhurst, line 31: 🔴 bug: `TUnit.Engine.Tests` does not compile because `ArgumentArity` comes from `Microsoft.Testing.Platform.Extensions.CommandLine`, which is not imported. Add the missing `using`.

thomhurst (commented):
Tested combined and TUnit-only CLI paths locally. Two blocking failures remain.

mvanhorn:
Both blockers are fixed in 7f904d2.

The props file conditions the `TestingPlatformBuilderHook` metadata on the TUnit identity instead of using `Update=`, so it no longer rewrites `TypeFullName` on every hook in a combined package. That is what produced the eight registrations and the adapter-factory abort before CLI parsing.

The reporter tests import `Microsoft.Testing.Platform.Extensions.CommandLine`, which is where `ArgumentArity` lives. Without it the project does not compile at all, which I confirmed: stashing just these two changes and building fails outright.

I installed SDK 10.0.400 to match global.json and the project builds and runs. There are unrelated failures on arm64 in areas like abstract classes and hooks that look environmental rather than related to this change.

One thing I did not land: coderabbit's suggestion for TUnit-only integration coverage in `HtmlReportCliTests`. I wrote it twice and could not get it green, the spawned app returns no stdout to assert on, so I left it out rather than add a red test. Happy to take a pointer on how those CLI tests are meant to invoke the app if you want that coverage.

thomhurst:
Thanks @mvanhorn !

PR #6681, opened 2026-08-26
fix: convert foreign array element-wise for trailing array parameter (#6678)

### LINKED ISSUE #6678: [Bug]: PR #6122 (allow targetting arrays in functions with args) was too broad breaking some use cases
### Description

#6122 seemed to add some nice features but I think it needs to check the source type prior to just assuming it isn't an array when going to an array arg.

In my tests I use a helpful matrix array attribute: https://github.com/thomhurst/TUnit/discussions/4596

it allows for nice syntax like:
```csharp
[Test]
[MatrixDataSource]
public async Task MyTest(

	[MatrixArray(
		[KNOWN_CONFIGS.ai, KNOWN_CONFIGS.sample_with_many_items],
		[KNOWN_CONFIGS.sample_with_many_items, KNOWN_CONFIGS.nowindow, KNOWN_CONFIGS.mcp]
	)] KNOWN_CONFIGS[] configsToload
```
and this used to work fine.    Technically I think the params []... on an ending array parameter makes the most sense but #6122 noted to match MSTest standard final args of arrays are also allowed.

The issue is the generated code no longer targets an array type for the arg:
```csharp
    return new global::System.Threading.Tasks.ValueTask(instance.MyTest((args[0] is null ? null : args[0] is global::KNOWN_CONFIGS[] arr ? arr : new global::KNOWN_CONFIGS[] { global::TUnit.Core.Helpers.CastHelper.Cast<global::KNOWN_CONFIGS>(args[0]) })));
```

The target cast type (generic arg) is not `global::KNOWN_CONFIGS[]` but instead `global::KNOWN_CONFIGS`.   

Now I can add an additional arg after the array in the test function but its a bit of a clumsy fix up.   

Even with AOT we have IsArray so before creating an array for the user, if we just check if args[0] is an array, and if so run cast on that rather than cast on the array element.

Essentially adding another test case:

```csharp
    return new global::System.Threading.Tasks.ValueTask(instance.MyTest((

   args[0] is null ? null : 
   args[0] is global::KNOWN_CONFIGS[] arr ? arr : 
   args[0].GetType().IsArray ? global::TUnit.Core.Helpers.CastHelper.Cast<global::KNOWN_CONFIGS[]>(args[0]) : 
   new global::KNOWN_CONFIGS[] { global::TUnit.Core.Helpers.CastHelper.Cast<global::KNOWN_CONFIGS>(args[0])}

)));
```

### Expected Behavior

prior behavior of an array arg being passed as an array

### Actual Behavior

the array is attempted to be double boxed but results in a cast error

### Steps to Reproduce

Likely need a matrix array helper:
```csharp
public class MatrixArrayAttribute(object[] arr1, object[] arr2 = null, object[] arr3 = null, object[] arr4 = null, object[] arr5 = null, object[] arr6 = null, object[] arr7 = null, object[] arr8 = null) : MatrixAttribute {

	protected MatrixArrayAttribute() : this(null) { }
	private object[][] AllArrays => [arr1, arr2, arr3, arr4, arr5, arr6, arr7, arr8];

	public override object?[] GetObjects(DataGeneratorMetadata dataGeneratorMetadata) {

		var lastNotNullItem = AllArrays.LastIndexOfAnyExcept((object[])null);
		return lastNotNullItem == -1 ? new object[1] : AllArrays.Take(lastNotNullItem + 1).ToArray();
	}
}
```

### TUnit Version

1.65.63

### .NET Version

.net 11

### Operating System

Windows

### IDE / Test Runner

```shell

```

### Additional Context

_No response_

### IDE-Specific Issue?

thomhurst:
Closes #6678

## Problem

#6122 made a trailing `T[]` parameter bind like `params`. For a single argument that is not already a `T[]`, both modes wrapped it as one element via `Cast<T>`:

```csharp
args[0] is null ? null : args[0] is KNOWN_CONFIGS[] arr ? arr : new KNOWN_CONFIGS[] { CastHelper.Cast<KNOWN_CONFIGS>(args[0]) }
```

Data sources frequently supply an array of a *different runtime type* — the reporter's `MatrixArray` helper (discussion #4596) yields an `object[]` per test case for a `KNOWN_CONFIGS[]` parameter. `object[]` is not `KNOWN_CONFIGS[]`, so the whole array was forced into one element and `Cast<KNOWN_CONFIGS>(object[])` threw. Reflection mode failed the same way (`Array.SetValue(object[], 0)` on an enum array). Pre-#6122 this worked because the non-params path used `Cast<T[]>`, which converts element-wise.

## Fix

- **`CastHelper.ToTrailingArray<T>(object)`** (new, AOT-safe — `new T[]` + `Cast<T>` per element):
  1. value `is T` → one-element array (C# params expansion; `params object[]` still receives an `int[]` as a single element, like C#)
  2. value is a rank-1 `Array` → convert element-wise (#6678)
  3. otherwise → `[Cast<T>(value)]` (unchanged behaviour)
- **Source-gen** (`TupleArgumentHelper`): single-argument branch now emits `… is T[] arr ? arr : CastHelper.ToTrailingArray<T>(args[i])`.
- **Reflection** (`ReflectionTestDataCollector`): same three-way rule in the single-argument branch.

Null and already-typed `T[]` passthrough are untouched; multi-argument and dynamic-count paths are untouched.

## Tests

- `ArgsAsArrayTests`: new `MatrixArray`-driven cases for a plain `ConfigKind[]` and a `params ConfigKind[]` parameter (the reporter's scenario), plus a guard that `params object[]` still receives an `int[]` as one element.
- Pass in source-gen and `--reflection` (net10.0), along with `DataDrivenTests`, `MethodDataSourceDrivenTests`, `Bugs/2112`, `Bugs/6150`.
- Source-generator snapshots: 13 `.verified.txt` updated — the only change is the single-argument expression swap plus the new test entries.
- `TUnit.PublicAPI` snapshots updated for the new public helper.

https://claude.ai/code/session_015Xx3PGYRU1KjEuEmzXMMpj

## Summary by CodeRabbit

* **New Features**
  * Improved handling of trailing array parameters, including arrays with different runtime element types.
  * Supports converting scalar values and object arrays into expected typed arrays.
  * Preserves existing arrays and treats arrays passed to `params object[]` as single elements when appropriate.

* **Bug Fixes**
  * Fixed data-driven test invocation scenarios involving matrix-provided arrays, generic arrays, jagged arrays, and enumerable values.

* **Tests**
  * Added coverage for trailing array and `params` array argument conversion across supported runtimes.

### DIFF
--- a/src/TUnit.Core.SourceGenerator/CodeGenerators/Helpers/TupleArgumentHelper.cs
+++ b/src/TUnit.Core.SourceGenerator/CodeGenerators/Helpers/TupleArgumentHelper.cs
@@ -146,8 +146,11 @@ public static class TupleArgumentHelper
                         // - null (passed as null, not as array with null element)
                         // - T[] (passed directly, not wrapped in another array)
                         // - T (wrapped in array with single element)
+                        // Anything else goes through CastHelper.ToTrailingArray, which also converts an
+                        // array of a different runtime type element-wise (e.g. an object[] supplied by a
+                        // MatrixAttribute subclass for a MyEnum[] parameter — issue #6678).
                         var singleArg = $"{argumentsArrayName}[{regularParamCount}]";
-                        var checkAndCast = $"({singleArg} is null ? null : {singleArg} is {paramsParam.Type.GloballyQualified()} arr ? arr : new {elementType.GloballyQualified()}[] {{ global::TUnit.Core.Helpers.CastHelper.Cast<{elementType.GloballyQualified()}>({singleArg}) }})";
+                        var checkAndCast = $"({singleArg} is null ? null : {singleArg} is {paramsParam.Type.GloballyQualified()} arr ? arr : global::TUnit.Core.Helpers.CastHelper.ToTrailingArray<{elementType.GloballyQualified()}>({singleArg}))";
                         argumentExpressions.Add(checkAndCast);
                     }
                     else
--- a/src/TUnit.Core/Helpers/CastHelper.cs
+++ b/src/TUnit.Core/Helpers/CastHelper.cs
@@ -52,10 +52,13 @@ public static class CastHelper
 
         if (value is Array { Rank: 1 } source)
         {
+            // Honour a non-zero lower bound (Array.CreateInstance with explicit bounds) rather than
+            // assuming the source is zero-based; the result is always a normal zero-based T[].
+            var lowerBound = source.GetLowerBound(0);
             var result = new T[source.Length];
             for (var i = 0; i < result.Length; i++)
             {
-                result[i] = Cast<T>(source.GetValue(i))!;
+                result[i] = Cast<T>(source.GetValue(lowerBound + i))!;
             }
 
             return result;
--- a/src/TUnit.Core/Helpers/CastHelper.cs
+++ b/src/TUnit.Core/Helpers/CastHelper.cs
@@ -30,6 +30,41 @@ public static class CastHelper
     }
 
     /// <summary>
+    /// Binds a single non-null argument to a trailing <c>T[]</c> parameter when the argument is not
+    /// already a <c>T[]</c>. Applies C# params expansion first (a <typeparamref name="T"/> value becomes
+    /// a one-element array), then converts any other array element-wise, and otherwise wraps the
+    /// converted scalar.
+    /// </summary>
+    /// <remarks>
+    /// The element-wise branch exists because data sources often supply arrays whose runtime type is
+    /// not the parameter's — e.g. an <c>object[]</c> from a custom <see cref="MatrixAttribute"/> for a
+    /// <c>MyEnum[]</c> parameter (issue #6678). Without it the whole array would be forced into a single
+    /// element and fail to cast. The <typeparamref name="T"/> check runs first so a <c>params object[]</c>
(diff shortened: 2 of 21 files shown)


thomhurst:
Addressed in the latest commit:

- **Non-zero lower bound** (Codex / Greptile / CodeRabbit): both `CastHelper.ToTrailingArray<T>` and the reflection mirror now iterate from `GetLowerBound(0)`. Added `TrailingArray_NonZeroLowerBoundArrayFromMatrix` (`Array.CreateInstance(typeof(object), [2], [1])`) — passes in source-gen and `--reflection`.
- **Element order** (CodeRabbit): the new tests use `IsEquivalentTo(..., CollectionOrdering.Matching)`.
- **Extract non-generic `ToTrailingArray(Type, object)`** (CodeRabbit nitpick): not doing this one. The generic overload must stay AOT-safe (`new T[]`), while a `Type`-based overload has to go through `Array.CreateInstance` (`RequiresDynamicCode`), so the two bodies can't share an implementation — extraction would only relocate the duplicate. The reflection collector already mirrors `CastHelper` inline this way for the multi-argument path, so the single-argument branch follows the same pattern.

https://claude.ai/code/session_015Xx3PGYRU1KjEuEmzXMMpj

PR #6779, opened 2026-09-11
perf: limit converter discovery to declarations

thomhurst:
`AotConverterGenerator` walks every syntax node in every test body and repeatedly formats attribute type names while deciding which declarations need converters. Restrict traversal to declarations, resolve `BaseTestAttribute` once per compilation, and compare symbols directly. Avoid allocating LINQ predicates while checking methods and classes.

Nested classes, partial declarations, constructor parameters, parameter attributes on partial-method implementations, and existing conversion generation are preserved. No runtime execution or public API changes.

## Performance evidence

Baseline: `a27cfb34bb`. BenchmarkDotNet 0.15.8, 10 measured iterations and 5 warm-up iterations, Roslyn 4.14.0, Windows 11, Intel Core i7-12700K (20 logical processors), .NET 10.0.12 x64. SDK selected by the repository's `global.json`: `11.0.100-preview.7.26381.103`.

Each operation runs the converter generator over 10,000 tests in 100 classes. Setup creates and validates the compilation, loads the actual old and new generator assemblies in separate `AssemblyLoadContext` instances, and checks identical generated sources. The benchmark prints distinct assembly MVIDs and returns each resulting driver. Both builds are measured side by side, after a successful Dry run, without concurrent builds.

| Input | Before mean | After mean | Before error | After error | Allocated before | Allocated after |
|---|---:|---:|---:|---:|---:|---:|
| Bare synchronous tests | 39.33 ms | 14.59 ms | 3.253 ms | 3.616 ms | 17.25 MB | 4.54 MB |
| 20 additional statements per body | 232.64 ms | 15.20 ms | 40.241 ms | 3.715 ms | 151.53 MB | 4.54 MB |
| One inline-data row per method | 59.20 ms | 19.02 ms | 15.100 ms | 8.205 ms | 18.09 MB | 5.69 MB |

Error is BenchmarkDotNet's 99.9% confidence-interval half-width. Timings retain substantial variance, but the allocation reductions and large timing differences are clear. This table is a fresh run after stopping a lingering Docker setup process; earlier exploratory timings are not used here. These are **converter-only measurements**, excluding parsing, project startup, other generators, and compilation of generated code.

### Whole-project rebuild experiment

A separate project references TUnit 1.65.68 and replaces only its Core source-generator assembly with the baseline or modified build. It contains 10,000 bare tests in 100 files. Restore is outside timing; each timed `dotnet build -c Release --no-restore` follows an edit to one file. One warm-up per version is discarded, then five rounds alternate the version order. Both outputs execute exactly 10,000 passing tests.

| Round | Before rebuild | After rebuild |
|---|---:|---:|
| 1 | 11,018.03 ms | 7,670.69 ms |
| 2 | 5,937.24 ms | 5,673.31 ms |
| 3 | 7,930.87 ms | 8,019.65 ms |
| 4 | 5,892.08 ms | 5,202.96 ms |
| 5 | 5,392.31 ms | 7,267.76 ms |

**These exploratory rebuild timings are inconclusive.** They preceded the background-process cleanup above. The mean improves while the median worsens; noise exceeds the isolated saving for bare tests. This PR claims reduced converter work and allocations, not a demonstrated overall rebuild speedup or elimination of the gap in [the framework comparison](https://www.meziantou.net/benchmarking-dotnet-test-frameworks-xunit-v3-nunit-mstest-and-tunit.htm).

## Validation

- Full Core source-generator test/snapshot suite: 122 passed, one pre-existing skipped test, zero failures. No existing verified snapshots changed.
- New regression test compiles generated converters and covers nested classes, partial classes, constructor data, partial-method parameter attributes, and types referenced only inside bodies.
- The new regression test also passes against the original generator assembly.
- Roslyn 4.14 variant builds with zero warnings/errors; the base generator uses the repository's Roslyn 4.7 configuration.
- Synthetic baseline and modified test applications each execute 10,000 passing tests.
- Benchmark setup verifies identical generated sources for all three input shapes.

The complete benchmark source and reproduction steps are posted in a PR comment. Temporary benchmark projects and results are kept outside the repository.

### DIFF
--- a/src/TUnit.Core.SourceGenerator/Generators/AotConverterGenerator.cs
+++ b/src/TUnit.Core.SourceGenerator/Generators/AotConverterGenerator.cs
@@ -107,6 +107,7 @@ public class AotConverterGenerator : IIncrementalGenerator
     private void ScanTestParameters(Compilation compilation, List<ConversionInfo> conversionInfos, CancellationToken cancellationToken)
     {
         var typesToScan = new HashSet<ITypeSymbol>(SymbolEqualityComparer.Default);
+        var baseTestAttribute = compilation.GetTypeByMetadataName(WellKnownFullyQualifiedClassNames.BaseTestAttribute.WithoutGlobalPrefix);
 
         foreach (var tree in compilation.SyntaxTrees)
         {
@@ -115,19 +116,26 @@ public class AotConverterGenerator : IIncrementalGenerator
             var semanticModel = compilation.GetSemanticModel(tree);
             var root = tree.GetRoot();
 
-            foreach(var nodes in root.DescendantNodes())
+            // Conversions depend on declarations and their attributes, never executable bodies.
+            foreach (var nodes in root.DescendantNodes(static node => node is not
+                         (BaseMethodDeclarationSyntax or BasePropertyDeclarationSyntax or BaseFieldDeclarationSyntax)))
             {
                 cancellationToken.ThrowIfCancellationRequested();
 
-                if(nodes is MethodDeclarationSyntax method)
+                if (nodes is MethodDeclarationSyntax method)
                 {
+                    if (method.AttributeLists.Count == 0)
+                    {
+                        continue;
+                    }
+
                     var methodSymbol = semanticModel.GetDeclaredSymbol(method);
                     if (methodSymbol == null)
                     {
                         continue;
                     }
 
-                    if (!IsTestMethod(methodSymbol))
+                    if (!IsTestMethod(methodSymbol, baseTestAttribute))
                     {
                         continue;
                     }
@@ -148,7 +156,7 @@ public class AotConverterGenerator : IIncrementalGenerator
                         continue;
                     }
 
-                    if (!IsTestClass(classSymbol))
+                    if (!IsTestClass(classSymbol, baseTestAttribute))
                     {
                         continue;
                     }
@@ -179,35 +187,42 @@ public class AotConverterGenerator : IIncrementalGenerator
         }
     }
 
-    private static bool IsTestMethod(IMethodSymbol method)
+    private static bool IsTestMethod(IMethodSymbol method, INamedTypeSymbol? baseTestAttribute)
     {
-        return method.GetAttributes().Any(attr =>
+        foreach (var attr in method.GetAttributes())
         {
             var attrClass = attr.AttributeClass;
             if (attrClass == null)
             {
-                return false;
+                continue;
             }
 
             var baseType = attrClass;
             while (baseType != null)
             {
-                if (baseType.ToDisplayString() == WellKnownFullyQualifiedClassNames.BaseTestAttribute.WithoutGlobalPrefix)
+                if (SymbolEqualityComparer.Default.Equals(baseType, baseTestAttribute))
                 {
                     return true;
                 }
                 baseType = baseType.BaseType;
             }
 
-            return false;
-        });
+        }
+
+        return false;
     }
 
-    private bool IsTestClass(INamedTypeSymbol classSymbol)
+    private static bool IsTestClass(INamedTypeSymbol classSymbol, INamedTypeSymbol? baseTestAttribute)
     {
-        return classSymbol.GetMembers()
-            .OfType<IMethodSymbol>()
-            .Any(IsTestMethod);
+        foreach (var member in classSymbol.GetMembers())
+        {
+            if (member is IMethodSymbol method && IsTestMethod(method, baseTestAttribute))
+            {
+                return true;
+            }
+        }
+
+        return false;
     }
 
     private void ScanAttributesForTypes(ImmutableArray<AttributeData> attributes, HashSet<ITypeSymbol> typesToScan)
--- a/tests/TUnit.Core.SourceGenerator.Tests/AotConverterScanTests.cs
+++ b/tests/TUnit.Core.SourceGenerator.Tests/AotConverterScanTests.cs
@@ -1,0 +1,71 @@
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using TUnit.Core.SourceGenerator.Generators;
+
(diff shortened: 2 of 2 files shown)


thomhurst:
## Reproducing the converter benchmark

Build `src/TUnit.Core.SourceGenerator` in Release at baseline `a27cfb34bb` and this PR's head. Save each `bin/Release/netstandard2.0/TUnit.Core.SourceGenerator.dll` in separate `before` and `after` directories inside a temporary console project outside the checkout. Build `src/TUnit.Core -c Release -f net10.0` and copy its `TUnit.Core.dll` into that project.

Create the project with `dotnet new console -f net10.0`, copy the repository's `global.json`, and add BenchmarkDotNet and Microsoft.CodeAnalysis.CSharp. The measured versions are **BenchmarkDotNet 0.15.8** and **Microsoft.CodeAnalysis.CSharp 4.14.0**.

Add these project items:

```xml
<ItemGroup>
  <Reference Include="TUnit.Core"><HintPath>TUnit.Core.dll</HintPath></Reference>
  <None Update="before/*.dll;after/*.dll" CopyToOutputDirectory="PreserveNewest" />
</ItemGroup>
```

The benchmark loads both builds in separate `AssemblyLoadContext` instances while sharing the host's Roslyn types. It prints their different MVIDs and checks that generated source is identical. This avoids accidentally measuring the same DLL twice when MSBuild reuses output directories across jobs.

<details>
<summary>Complete benchmark source</summary>

```csharp
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Running;
using Microsoft.CodeAnalysis;
using Microsoft.CodeAnalysis.CSharp;
using System.Text;
using System.Runtime.Loader;

BenchmarkSwitcher.FromAssembly(typeof(Program).Assembly).Run(args);

[MemoryDiagnoser]
public class ConverterBenchmarks
{
    [Params("Bare", "BodyHeavy", "InlineData")]
    public string Scenario { get; set; } = "Bare";
    private CSharpCompilation _compilation = null!;
    private GeneratorDriver _before = null!;
    private GeneratorDriver _after = null!;

    [GlobalSetup]
    public void Setup()
    {
        var trees = new List<SyntaxTree>();
        for (var c = 0; c < 100; c++)
        {
            var source = new StringBuilder("using System; using TUnit.Core; public class Tests" + c + " {");
            for (var m = 0; m < 100; m++)
            {
                source.Append("[Test]");
                source.Append(Scenario == "InlineData" ? "[Arguments(42)] public void Test" : "public void Test");
                source.Append(m).Append(Scenario == "InlineData" ? "(int value) {" : "() {");
                source.Append("int x = 42; if (x * x + 1 != 1765) throw new Exception();");
                if (Scenario == "BodyHeavy")
                    for (var i = 0; i < 20; i++) source.Append("x = Math.Abs(x) + 1;");
                source.Append('}');
            }
            source.Append('}');
            trees.Add(CSharpSyntaxTree.ParseText(source.ToString()));
        }
        var references = ((string)AppContext.GetData("TRUSTED_PLATFORM_ASSEMBLIES")!).Split(Path.PathSeparator)
            .Select(p => MetadataReference.CreateFromFile(p)).ToList();
        references.Add(MetadataReference.CreateFromFile(typeof(TUnit.Core.TestAttribute).Assembly.Location));
        _compilation = CSharpCompilation.Create("SyntheticSuite", trees, references,
            new CSharpCompilationOptions(OutputKind.DynamicallyLinkedLibrary));
        var errors = _compilation.GetDiagnostics().Where(d => d.Severity == DiagnosticSeverity.Error).ToArray();
        if (errors.Length != 0) throw new InvalidOperationException(string.Join("\n", errors.Select(e => e.ToString())));
        _before = LoadDriver("before", Path.Combine(AppContext.BaseDirectory, "before", "TUnit.Core.SourceGenerator.dll"));
        _after = LoadDriver("after", Path.Combine(AppContext.BaseDirectory, "after", "TUnit.Core.SourceGenerator.dll"));
        var beforeResult = _before.RunGenerators(_compilation).GetRunResult();
        var afterResult = _after.RunGenerators(_compilation).GetRunResult();
        if (beforeResult.Diagnostics.Concat(afterResult.Diagnostics).Any(d => d.Severity == DiagnosticSeverity.Error))
            throw new InvalidOperationException("Generator diagnostics");
        if (!beforeResult.GeneratedTrees.Select(t => t.ToString()).SequenceEqual(afterResult.GeneratedTrees.Select(t => t.ToString())))
            throw new InvalidOperationException("Generated sources differ");
    }

    private static GeneratorDriver LoadDriver(string name, string path)
    {
        var assembly = new AssemblyLoadContext(name).LoadFromAssemblyPath(path);
        Console.WriteLine($"{name} generator MVID: {assembly.ManifestModule.ModuleVersionId}");
        var generator = (IIncrementalGenerator)Activator.CreateInstance(assembly.GetType("TUnit.Core.SourceGenerator.Generators.AotConverterGenerator")!)!;
        return CSharpGeneratorDriver.Create(generator);
    }

    [Benchmark(Baseline = true)]
    public GeneratorDriver Before() => _before.RunGenerators(_compilation);

    [Benchmark]
    public GeneratorDriver After() => _after.RunGenerators(_compilation);
}
```

</details>

Run:

```powershell
dotnet run -c Release -- --filter '*ConverterBenchmarks*' --job Dry --artifacts ./dry
dotnet run -c Release --no-build -- --filter '*ConverterBenchmarks*' --iterationCount 10 --warmupCount 5 --artifacts ./measured --exporters json
```

Each operation runs the entire converter generator over 10,000 test methods across 100 classes. Syntax creation, assembly loading, and compilation diagnostics are outside measurement. The initial generator driver is reused without retaining its returned cached output, so each operation performs generation again. This measures generator work, not process startup or compilation of generated code.

The three inputs are bare synchronous tests, tests with 20 additional statements in each body, and one `[Arguments(42)]` row per method. Benchmark methods return the resulting driver; the memory column is allocation per complete generator pass.

PR #6837, opened 2026-09-18
fix(mocks): keep editors in sync with publicized project references (#6836)

### LINKED ISSUE #6836: [TUnit.Mocks] Experimental internals access: IDE reports false CS0122 because only ReferencePathWithRefAssemblies is publicized
## Summary

With the experimental compile-time internals access (`TUnitMocksExperimentalInternalsAccess` + `TUnitMocksInternalsAccess`, #6514 Tier 2 / #6520), `dotnet build` succeeds and the test passes, but every IDE / Roslyn-workspace-based tool reports a false `CS0122 'X' is inaccessible due to its protection level` on the internal type.

Cause: `TUnitMocksPublicizeReferences` swaps the publicized copy into `ReferencePathWithRefAssemblies` only. Design-time builds and `MSBuildWorkspace` resolve references from `ReferencePath`, which still points at the original, un-publicized assembly.

## Repro

`SdkLib/Api.cs`:

```csharp
namespace SdkLib;

internal interface IQuotaPolicy
{
    bool Allow(string clientId);
}
```

`Tests/Tests.csproj`:

```xml
<PropertyGroup>
  <TargetFramework>net10.0</TargetFramework>
  <TUnitMocksExperimentalInternalsAccess>true</TUnitMocksExperimentalInternalsAccess>
</PropertyGroup>
<ItemGroup>
  <PackageReference Include="TUnit" Version="1.68.0" />
  <PackageReference Include="TUnit.Mocks" Version="1.68.0" />
</ItemGroup>
<ItemGroup>
  <ProjectReference Include="..\SdkLib\SdkLib.csproj" />
  <TUnitMocksInternalsAccess Include="SdkLib" />
</ItemGroup>
```

`Tests/QuotaTests.cs`:

```csharp
using SdkLib;
using TUnit.Mocks;

public class QuotaTests
{
    [Test]
    public async Task Internal_policy_is_mockable()
    {
        var policy = IQuotaPolicy.Mock();
        policy.Allow(Any()).Returns(false);

        await Assert.That(policy.Object.Allow("acme")).IsFalse();
    }
}
```

## Actual vs expected

`dotnet build` → 0 warnings, 0 errors. `dotnet run` → 1/1 passed. Correct.

Loading the same project through Roslyn's `MSBuildWorkspace` (what OmniSharp and the C# language server use) and asking the compilation for diagnostics:

```
WORKSPACE-ERROR CS0122 10,21: 'IQuotaPolicy' is inaccessible due to its protection level
error count: 1
```

In an editor the whole test is underlined red, and a second error appears where the mock is passed on:

```
CS1503: Argument 1: cannot convert from 'SdkLib.IQuotaPolicy?' to 'SdkLib.IQuotaPolicy?'
```

Expected: the IDE agrees with the compiler, since the build is authoritative and correct.

## Root cause

```console
$ dotnet msbuild Tests/Tests.csproj -t:Compile -p:SkipCompilerExecution=true -p:ProvideCommandLineArgs=true -getItem:ReferencePath
"Identity": "/tmp/iavrepro/SdkLib/bin/Debug/net10.0/SdkLib.dll"                      # original

$ ... -getItem:ReferencePathWithRefAssemblies
"Identity": "obj/Debug/net10.0/tunit-mocks-internals/68fe7f1d/SdkLib.dll"            # publicized

$ dotnet msbuild Tests/Tests.csproj -t:ResolveAssemblyReferencesDesignTime -p:DesignTimeBuild=true -getItem:ReferencePath
"Identity": "/tmp/iavrepro/SdkLib/bin/Debug/net10.0/SdkLib.dll"                      # original
```

`TUnit.Mocks.InternalsAccess.targets` does this deliberately, and the comment explains why — `ReferencePath` has to keep the original so copy-local and `deps.json` bind to the real assembly at runtime. That reasoning holds for a real build, but it also means no design-time consumer ever sees the publicized copy.

## Suggested fix

Swap `ReferencePath` as well when `'$(DesignTimeBuild)' == 'true'`. Nothing is copied in a design-time build, so the copy-local / `deps.json` argument does not apply there, and the editor would then see what the compiler sees.

If that is not workable, it would be worth a line in the docs for the feature, since the failure mode looks exactly like a broken setup.

## Versions

- `TUnit` / `TUnit.Mocks` 1.68.0, `net10.0`, SDK 10.0.401, macOS arm64
- Same behaviour path in 1.63.0–1.68.0: `TUnit.Mocks.InternalsAccess.targets` is unchanged across those releases
- Probe used: `Microsoft.CodeAnalysis.Workspaces.MSBuild` 4.14.0 + `Microsoft.Build.Locator` 1.7.8, opening `Tests.csproj` and printing `Compilation.GetDiagnostics()` errors

thomhurst:
Fixes #6836.

## Problem

With experimental internals access (`TUnitMocksExperimentalInternalsAccess` + `@(TUnitMocksInternalsAccess)`), the build is correct but Roslyn-workspace tooling reports a false `CS0122 '<type>' is inaccessible due to its protection level` on every internal type, plus a follow-on `CS1503` where the mock is passed on. In an editor the whole test file is red.

## Root cause (differs from the issue's diagnosis)

The issue suggested `ReferencePath` is what design-time consumers read, and that only `ReferencePathWithRefAssemblies` is publicized. Probing an `MSBuildWorkspace` load of this repository's own `tests/TUnit.Mocks.InternalsAccess.Tests` project shows the publicized copy *does* reach the workspace — it arrives on the compiler command line, which is built from `ReferencePathWithRefAssemblies`:

```
METAREF ...obj\Debug\net10.0\tunit-mocks-internals\f1a7abfd\TUnit.Mocks.InternalsAccess.TargetLib.dll
PROJREF TUnit.Mocks.InternalsAccess.TargetLib      <-- also present
...
error count: 13
```

The problem is the second line. When a publicized assembly comes from a `ProjectReference`, the workspace *also* adds the referenced project as a compilation reference, whose internals were never publicized. Two references with the same assembly identity collide, the source compilation wins, and the internal types read as inaccessible. Swapping `ReferencePath` would not change this.

## Fix

In design-time builds only, set `ReferenceOutputAssembly=false` on the `ProjectReference` items backing the references this run publicized. Nothing is compiled or copied in a design-time build, so the publicized copy is left as the only reference for that assembly and the editor sees what `Csc` sees.

- Real builds are untouched: the project reference still builds, copies local, and lands in `deps.json` — the reason `ReferencePath` must keep the original.
- Only references that were actually publicized are detached, so a failed publicize leaves the project reference intact.
- Opt out with `<TUnitMocksInternalsAccessDetachDesignTimeProjectReferences>false</TUnitMocksInternalsAccessDetachDesignTimeProjectReferences>`.

Trade-off, documented: the referenced project appears in the editor as a compiled assembly, so "go to definition" lands on metadata and edits to it reach the test project after a rebuild.

## Verification

- Same `MSBuildWorkspace` probe (`Microsoft.CodeAnalysis.Workspaces.MSBuild` 4.14 + `Microsoft.Build.Locator` 1.7.8) over `tests/TUnit.Mocks.InternalsAccess.Tests`: **13 errors before, 0 after**, with the project reference gone and the publicized reference retained.
- A standalone reproduction of the issue's `SdkLib`/`Tests` shape outside the repository: same result.
- Transitive project references were already correct (the intermediate project does not re-export the assembly into the consumer's reference list) — verified, unaffected, and not touched by this change.
- `tests/TUnit.Mocks.InternalsAccess.Tests`: 32/32 pass in Debug and Release, including the runtime end-to-end tests that prove real builds still bind to the original assembly.

## Tests

New `DesignTimeProjectReferenceTests` generates a minimal SDK project pair outside the repository (so nothing races with a repository build), publicizes across the project reference, and asserts the metadata through `dotnet msbuild -getItem:ProjectReference`: detached under `DesignTimeBuild=true`, untouched in a real build, and untouched when the opt-out is set. ~3.5s for all three.

## Summary by CodeRabbit

* **New Features**
  * Editors can use publicized assemblies during design-time builds when internals access involves project references.
  * Added an option to opt out of design-time project-reference detachment.
  * Real builds continue using project references unchanged.

* **Documentation**
  * Documented design-time behavior, metadata navigation trade-offs, and configuration options.


thomhurst:
Both review points addressed in c2bcc37.

**Temporary builds leaked** — correct, fixed. `Scenario` is now `IAsyncDisposable` and deletes its generated project pair; verified a run now leaves zero directories under `%TEMP%/tunit-mocks-ia-designtime`.

**Workspace regression untested** — partly taken. Added a test asserting the design-time compiler command line still carries exactly one reference for the publicized assembly and that it is the publicized copy, so a detach that costs the reference fails the suite.

I did not add an in-process `MSBuildWorkspace` load. This test project references `Microsoft.Build.Utilities.Core` directly (the publicizer task runs in-process in its unit tests), and `MSBuildLocator` requires that no MSBuild assemblies are already resolved in the process — the two cannot share a test host. The workspace behaviour was verified out-of-process instead, on this repository's own internals-access test project and on a standalone reproduction of the issue's shape: 13 errors before, 0 after, with the duplicate project reference gone and the publicized metadata reference retained. What the in-repo tests guard is the MSBuild side that TUnit controls: that `ReferenceOutputAssembly=false` is applied at design time, only at design time, and never at the cost of the reference itself.

thomhurst:
CodeRabbit's two points, both valid, fixed in c925213:

- **XML escaping** — the generated probe project now escapes the checkout paths it interpolates (`SecurityElement.Escape`), so a path containing markup characters no longer produces an invalid project file.
- **Cleanup on failed setup** — the scenario root is deleted when its build fails as well as on dispose (dispose itself landed in c2bcc37).

Chasing the first of these surfaced a real defect in the test that neither bot flagged: the spawned builds were inheriting the MSBuild server, which outlives the process, keeps the publicizer task assembly loaded, and locks this test project's own build output against the next build in the repository (`MSB3027`). `-nr:false` does not cover it. The spawned processes now run with `MSBUILDUSESERVER=0` and node reuse disabled; verified by running the tests and immediately rebuilding the project, which previously failed. 33/33 pass.

